PR #18, opened 2020-10-17
Make packet type header always 16 bit

mir3x:
It will be less code later :dancing_women:


lmoureaux:
I'm not sure… apart from a bit less code, what do we gain from breaking compat with mainline clients?

mir3x:
:dancers:  good point

PR #21, opened 2020-10-18
Windows compilation with MinGW32

mir3x:
Not sure if moving DIRENT_H out of if(UNIX) wont break msys build ?


lmoureaux:
> Not sure if moving DIRENT_H out of if(UNIX) wont break msys build ?

Nope, I think I had to move it out also for msys. And msys doesn't work with the current code anyway.

PR #25, opened 2020-10-20
Use QUdpSocket in client 

mir3x:
in lan scans endless loop was removed fot signal/slot so it could be moved out of threads in future.

### DIFF
--- a/client/servers.cpp
+++ b/client/servers.cpp
@@ -15,42 +15,6 @@
 #include <fc_config.h>
 #endif
 
-#include "fc_prehdrs.h"
-
-#include <errno.h>
-#include <stdlib.h>
-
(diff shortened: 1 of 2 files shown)


mir3x:
+217 −418, not bad :1234:

mir3x:
I found new bug, create new PR later

lmoureaux:
> in lan scans endless loop was removed fot signal/slot so it could be moved out of threads in future.

I think this can live in the main thread, signals and slots make it async ("collaborative parallel processing").

> +217 −418, not bad 1234

And you've not counted all the `fc_socket` stuff that this will help get rid of!

> I found new bug, create new PR later

You can just change the title to `[WIP] something` and it will prevent merging :wink:

Let's also make sure that `clang-format` is happy :smiling_imp:

PR #26, opened 2020-10-20
Udp scan in client

mir3x:
Removed main loop in scan to signal/slot so thread could be removed there.

### DIFF
--- a/client/servers.cpp
+++ b/client/servers.cpp
@@ -15,42 +15,6 @@
 #include <fc_config.h>
 #endif
 
-#include "fc_prehdrs.h"
-
-#include <errno.h>
-#include <stdlib.h>
-
-#ifdef FREECIV_HAVE_SYS_TYPES_H
-#include <sys/types.h>
-#endif
-#ifdef HAVE_SYS_SOCKET_H
-#include <sys/socket.h>
-#endif
-#ifdef HAVE_NETINET_IN_H
-#include <netinet/in.h>
-#endif
-#ifdef HAVE_ARPA_INET_H
-#include <arpa/inet.h>
-#endif
-#ifdef HAVE_NETDB_H
-#include <netdb.h>
-#endif
-#ifdef HAVE_PWD_H
-#include <pwd.h>
-#endif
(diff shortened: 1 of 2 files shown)


client/servers.cpp
```diff
@@ -873,27 +667,14 @@ void server_scan_finish(struct server_scan *scan)
...
     fc_destroy_mutex(&scan->meta.mutex);
 
-    /* This mainly duplicates code from below "else" block.
-     * That's intentional, since they will be completely different in future
-     * versions. We are better prepared for that by having them separately
-     * already. */
-    if (scan->sock >= 0) {
-      fc_closesocket(scan->sock);
-      scan->sock = -1;
-    }
```
lmoureaux, line 883: Missing `fcUdpScan::i()->drop();`?

client/servers.cpp
```diff
@@ -15,70 +15,29 @@
...
-#ifdef HAVE_SYS_UIO_H
-#include <sys/uio.h>
-#endif
-#ifdef HAVE_UNISTD_H
-#include <unistd.h>
-#endif
-
 // Qt
 #include <QByteArray>
+#include <QDebug>
```
lmoureaux, line 20: Unused? You got rid of my QtDebug (oops) but it came back :D

client/servers.cpp
mir3x: Its there in server scan finish line 677 ( no idea how to link code :D )
`  } else {
    fcUdpScan::i()->drop();`

client/servers.cpp
```diff
@@ -873,27 +667,14 @@ void server_scan_finish(struct server_scan *scan)
...
       scan->srvrs.servers = NULL;
       fc_release_mutex(&scan->srvrs.mutex);
     }
   } else {
-    if (scan->sock >= 0) {
-      fc_closesocket(scan->sock);
-      scan->sock = -1;
-    }
-
+    fcUdpScan::i()->drop();
```
mir3x, line 677: Its here ( if  not global scan)

lmoureaux: Oh I see, it's only used for local scans

PR #27, opened 2020-10-20
Server lan UDP response moved to QUdpSocket

mir3x:


mir3x:
forgot to make clang-format happy .

PR #29, opened 2020-10-20
Fixed city.cpp::566]: assertion 'NULL == city_map_index' failed.

mir3x:
Origianl alloc was: city_map_index = fc_malloc(city_count_tiles * sizeof(*city_map_index));


mir3x:
:-1:  Im blind

PR #36, opened 2020-10-24
Added QProcess to start local server instead fork

mir3x:


client/connectdlg_common.cpp
```diff
@@ -16,8 +16,9 @@
 
 #include "fc_prehdrs.h"
 
+#include <QDebug>
```
lmoureaux, line 19: Useless

client/connectdlg_common.cpp
```diff
@@ -66,25 +56,53 @@
...
+  if (!m_instance) {
+    m_instance = new serverProcess;
+  }
+  return m_instance;
+}
+/* Removes server process */
+void serverProcess::drop()
+{
+  if (m_instance) {
+    delete m_instance;
```
lmoureaux, line 101: Direct `delete` on a `QObject` is dangerous (there may be pending events still addressed at this object), use `deleteLater()`.

client/connectdlg_common.cpp
```diff
@@ -274,7 +222,7 @@ bool client_start_server(void)
     return FALSE;
   }
 
-  storage = freeciv_storage_dir();
+  storage = QString(freeciv_storage_dir());
```
lmoureaux, line 225: `QString::fromUtf8` for those users with an accent in their name (all Freeciv strings are UTF-8)

client/connectdlg_common.cpp
```diff
@@ -285,299 +233,67 @@ bool client_start_server(void)
...
-    log_verbose("Tried with commandline: '%s'", cmdline3);
-    log_verbose("Tried with commandline: '%s'", cmdline4);
-    output_window_append(ftc_client, _("Couldn't start the server."));
-    output_window_append(ftc_client,
-                         _("You'll have to start one manually. Sorry..."));
-    return FALSE;
+  for (auto trueServer : qAsConst(program)) {
+    serverProcess::i()->start(trueServer, arguments);
+    trueFcser = trueServer;
+    if (serverProcess::i()->waitForStarted(3000) == true) {
```
lmoureaux, line 269: (Question not comment) Does this block the event loop or is it in another thread?

lmoureaux (commented):
Nice stats +86 -375 :D :D

client/connectdlg_common.cpp
mir3x: No, its used by qInfo()

client/connectdlg_common.cpp
mir3x: it blocks

PR #40, opened 2020-10-27
Fixed governor crash and assert fail

mir3x:
It fixes governor crash
and assert fail - it removess assert but it was introduced with 6f5d4f4 when adding "utype_build_shield_cost" which has only those arguments(const struct city *pcity, const struct unit_type *punittype) and homeless units have no way to find pplayer by pcity.


lmoureaux (approved):
It's ok, `get_target_bonus_effects` can use `NULL` everywhere.

PR #58, opened 2020-11-03
C++ify cmdhelp.cpp

mir3x:
Shaved cmdhelp by 80 lines.

### DIFF
--- a/common/fc_cmdhelp.cpp
+++ b/common/fc_cmdhelp.cpp
@@ -1,31 +1,23 @@
-/*****************************************************************************
- Freeciv - Copyright (C) 1996 - A Kjeldberg, L Gregersen, P Unold
-   This program is free software; you can redistribute it and/or modify
-   it under the terms of the GNU General Public License as published by
-   the Free Software Foundation; either version 2, or (at your option)
-   any later version.
-
-   This program is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


common/fc_cmdhelp.cpp
```diff
@@ -113,9 +93,8 @@ void cmdhelp_display(struct cmdhelp *pcmdhelp, bool sort, bool gui_options,
   fc_fprintf(stderr, _("Usage: %s [option ...]\nValid option are:\n"),
              pcmdhelp->cmdname);
 
-  cmdarg_list_sort(pcmdhelp->cmdarglist, cmdarg_compare);
-  cmdarg_list_iterate(pcmdhelp->cmdarglist, pcmdarg)
-  {
+  std::sort(pcmdhelp->cmdarglist->begin(), pcmdhelp->cmdarglist->end());
```
lmoureaux, line 96: This `sort` will sort by the value of the pointer, while entries were sorted in alphabetical order before.

lmoureaux (changes requested):
In the server code I replaced `cmdhelp` with `QCommandLineParser`, still not done in `client` and `tools` but it's on the roadmap. I don't think this code is used for server command line parsing apart from the tokenizer function, so let's not invest too much time in it. (If you want me to speed up on `QCommandLineParser` migration, we can drop this code by the end of the week.)

common/fc_cmdhelp.cpp
mir3x: Hehe, I was checking it on server and results were the same, but it seems only client uses it

PR #63, opened 2020-11-06
Dont use threads in network scan

mir3x:
Removed threads from client network scans


lmoureaux:
Wait, it was *already* not using threads? Hilarious.

PR #72, opened 2020-11-09
Fixed some qt Issues

### LINKED ISSUE #48: On-map worked tile chooser doesn't highlight the correct tiles
It can highlight too little, or too much. In the picture below tiles on the left are missing to create a complete hexagon, but too many tiles are highlighted (the working radius is 1 and the vision radius is 2, alio ruleset and tileset):
image

mir3x:
It should be close #48 not 49 in b45f0bc

### DIFF
--- a/client/tilespec.cpp
+++ b/client/tilespec.cpp
@@ -1470,7 +1470,7 @@ static struct sprite *load_gfx_file(const char *gfx_filename)
   }
 
   log_error("Could not load gfx file \"%s\".", gfx_filename);
-  color *c = color_alloc(255, 0 , 0);
+  color *c = color_alloc(255, 0, 0);
   return create_sprite(20, 20, c);
 }
 
@@ -4734,7 +4734,7 @@ static int fill_city_overlays_sprite_array(const struct tileset *t,
                                            const struct tile *ptile,
                                            const struct city *citymode)
 {
-   const struct city *pcity;
+  const struct city *pcity;
   const struct city *pwork;
   struct unit *psettler;
   struct drawn_sprite *saved_sprs = sprs;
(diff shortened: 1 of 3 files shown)


client/tilespec.cpp
```diff
@@ -5224,13 +5224,14 @@ static int fill_grid_sprite_array(
...
       }
-        if (tile) {
-          { // Draw city grid for main citymap
-            xcity = find_city_near_tile(tile);
-            if (xcity && xcity->client.city_opened) {
-              ADD_SPRITE_SIMPLE(t->sprites.grid.selected[pedge->type]);
-            }
+      if (tile) {
+        { // Draw city grid for main citymap
+          xcity = find_city_near_tile(tile);
```
lmoureaux, line 5229: I think it might still fail because `find_city_near_tile` always returns the closest city?

mir3x: comment says that:
   Find the "best" city/settlers to associate with the selected tile.
     a.  If a visible city is working the tile, return that city.
     b.  If another player's city is working the tile, return NULL.
     c.  If any selected cities are within range, return the closest one.
     d.  If any cities are within range, return the closest one.
     e.  If any active (with color) settler could work it if they founded a
         city, choose the closest one (only if punit != NULL).
     f.  If any settler could work it if they founded a city, choose the
         closest one (only if punit != NULL).
     g.  If nobody can work it, return NULL.

mir3x: I tried with some cities near and was working as expected but maybe I was lucky

lmoureaux: ok let's merge and reopen the issue if it's still there

mir3x: or maybe I jsut iterate to be sure

PR #73, opened 2020-11-10
Stop useless minimap updates and thread spam

mir3x:
Looks like working well


client/gui-qt/minimap.cpp
```diff
@@ -272,7 +272,8 @@ void minimap_view::update_image()
   if (isHidden()) {
     return;
   }
-  thread.render(scale_factor, width(), height());
+  QTimer::singleShot(
+      10, [this] { thread.render(scale_factor, width(), height()); });
```
lmoureaux, line 276: What if `this` is deleted before the timer fires?

mir3x: No one will ever know :  it doesnt exist anymore

lmoureaux: `thread`, `width` or `height` may crash since they're not pointing anywhere…

mir3x: what u mean? There is no lambda anymore.

lmoureaux: The lambda stays there (in the Qt event loop mechanism) and its captured `this` becomes a dangling pointer if the object it points to gets deleted. [Apparently](https://stackoverflow.com/a/19721153) one needs to use `connect` and set the "context" of the connection, or wrap it into a `QPointer` and check it's still alive in the lambda. Bit annoying but necessary.

mir3x: :-1: 
Read what I wrote, there is no lambda. Someone deleted lambda! Poor lambda.

lmoureaux: lol sorry :P poor lambda indeed, rip

PR #81, opened 2020-11-13
asanfixes

mir3x:
Commit fast so I won't need to do merge conflicts  :+1: 
Introduced new mem leaks with genhash, but Im going to remove genhash next


lmoureaux:
So we can almost delete `fc_malloc` and `fc_free` now?

PR #93, opened 2020-11-19
New option to increase/decrease all fonts at once

mir3x:
Option is in fonts category.
It has all changes from "useless" PR bc it changes similar files.


lmoureaux:
Rebasing should fix Travis

PR #110, opened 2020-11-22
Some minor bugs

mir3x:


lmoureaux:
I'm still fine-tuning Codacy warnings. I disabled a few MISRA rules that either aren't essential or lead to false positives.

Issue #112, opened 2020-11-22
Hide stack traces by default?


lmoureaux:
They're useless unless for devs. Maybe better require devs to set an env var than every user to do the same.

ec429:
If they weren't so long it'd be arguable, but 100+ lines of stack trace per error quickly pushes the interesting bits out of xterm's scrollback buffer.
At least make them off-by-default if built without debugging symbols; a backtrace full of hex addresses is not at all informative.

mir3x:
I was very happy when those logs caused my tmp to be full so I couldnt even compile again without clearing it.

lmoureaux:
Done in #120.

PR #120, opened 2020-11-23
Print stack traces with qDebug (was qInfo)

lmoureaux:
Traces are useless for most users and should therefore be disabled by default. They can still be selectively enabled using `QT_LOGGING_RULES`.


lmoureaux:
Closes #112

PR #124, opened 2020-11-24
Elimated over 200 coverity warnings

mir3x:
Defect density 0.48 from 1.08
Im the best :D


server/gamehand.cpp
```diff
@@ -197,12 +197,8 @@ static struct tile *place_starting_unit(struct tile *starttile,
   /* Expose visible area. */
   map_show_circle(pplayer, ptile, game.server.init_vis_radius_sq);
 
-  if (utype != NULL) {
-    (void) create_unit(pplayer, ptile, utype, FALSE, 0, 0);
-    return ptile;
-  }
-
-  return NULL;
+  (void) create_unit(pplayer, ptile, utype, FALSE, 0, 0);
```
lmoureaux, line 200: False positive, `utype` can be `NULL` if `ptype` is `NULL`, because `crole_to_unit_type` can return `NULL`.

server/ruleset.cpp
```diff
@@ -3527,11 +3527,6 @@ static bool load_ruleset_terrain(struct section_file *file,
         ok = FALSE;
         break;
       }
-
-      if (!ok) {
-        break;
-      }
-
```
lmoureaux, line 3535: False positive, `ok` can be set to `false` two lines above.

lmoureaux (changes requested):
Two of the issues you fixed were false positives, the rest looks fine :)

server/ruleset.cpp
mir3x: 2 lines above is         
 ok = FALSE;
        break;
and it break iteration loop and never reaches that if

server/gamehand.cpp
mir3x: Coverity doesnt have false positives in such cases
struct tile *ptile = NULL;
 utype is set to something 

  if (utype != NULL) {
       ** some iteartion combo and ptile is set here
  }

  if (ptile == NULL) {
    ** if utype omited ealier iteration combo(beacause was NULL) ptile would be NULL also and would return here
    return NULL;
  }

lmoureaux:
LOL at Codacy issues.

PR #126, opened 2020-11-25
Fix assertion fail for global observer

mir3x:
research.cpp:555) - Assertion state == TECH_UNKNOWN failed


mir3x:
it also in original  freeciv failing

PR #127, opened 2020-11-25
FIxed many clazy warnings - mostly unneeded temporaries while iterati…

mir3x:
…ng qhash


lmoureaux:
When should we use `QStringLiteral`, and when should we use `QLatin1String`? I see changes from `""` to `QLatin1String("")` and from `""` to `QStringLiteral("")`.

mir3x:
> When should we use `QStringLiteral`, and when should we use `QLatin1String`? I see changes from `""` to `QLatin1String("")` and from `""` to `QStringLiteral("")`.

https://github.com/KDE/clazy/blob/master/docs/checks/README-qstring-allocations.md
Overall when using empty string "" -> then QLatin1String(""); "QString is optimized for this case"
When comparing both are correct, just clazy picked QLatin1
s == QLatin1String("foo") // No allocation
s == QStringLiteral("foo") // No allocation

When doing "append", "compare", "endsWith", "startsWith", "indexOf", "insert", "lastIndexOf", "prepend", "replace", "contains" an QString methods clazy picked QLatin1, but maybe QStringLiteral is correct too?

So overall I think QLatin1String for empty strings, QStringLiterals for others, or just dont use anything and run clazy once per year :D

lmoureaux:
Thanks for the explanation, I'll try to stick to it :)

PR #128, opened 2020-11-27
Misc bugs

mir3x:


mir3x:
After last patch AI will ignore units 30 tiles away or 15 tiles away on another continet. It has a lot of other issues to fix.
This speeds up turn change in big save more than 300%.

client/clinet.cpp
```diff
@@ -359,7 +359,7 @@ double try_to_autoconnect(void)
     return FC_INFINITY;
   } else {
     // All errors are fatal
-    qFatal(_("Error contacting server \"%s\" at port %d "
+    qInfo(_("Error contacting server \"%s\" at port %d "
```
lmoureaux, line 362: Shouldn't this be at least `qWarning`? I put `qFatal` because we fail to behave as mandated by cmd line arguments, but could lower it to `qCritical` or `qWarning` if we want to continue.

mir3x: OK I used qCritical and it doesnt crash, no idea why its crashing with qFatal, at first I wanted to use printf :D

lmoureaux: By default, `qFatal` calls `abort()`. I could change it to call `exit(EXIT_FAILURE)` instead.

PR #130, opened 2020-11-29
Reduced packet diplstate

mir3x:
fixed compilation of packet statistics - it must be set in source file ( it wasnt updated for ages, replaced log_debugs with printfs there)
enable is-game-info for packet diplstate

### DIFF
--- a/common/networking/packets.cpp
+++ b/common/networking/packets.cpp
@@ -274,14 +274,14 @@ int send_packet_data(struct connection *pc, unsigned char *data, int len,
     packets_stats[packet_type].size += size;
 
     packet_counter++;
-    if (packet_type == PACKET_START_TURN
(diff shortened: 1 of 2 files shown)


lmoureaux:
If you need to change it in the code then it's intentional so `printf` is ok. A logging category and `qCInfo` would rock :man_dancing:

Is there a good reason to keep these stats disabled by default? Incrementing few counters every time we send a packet doesn't look like a huge overhead for quite valuable information.

mir3x:
They spam stats every 100 packets and write all packets stats - every packet type and how much times was sent/received, how much data it took, probably few kilo of lines at big turn change

lmoureaux:
Ok so if we do this every TC it's not an issue, but every 100 packets sounds like a lot.

PR #133, opened 2020-11-30
Iconv

mir3x:
Not sure if everythign works bc conversions are from UTF8 -> UTF8, maybe we could add language switcher in client

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -166,7 +166,7 @@ static bool client_quitting = FALSE;
  **************************************************************************/
 static char *put_conv(const char *src, size_t *length)
 {
-  char *out = internal_to_data_string_malloc(src);
+  char *out = qstrdup(src);
 
   if (out) {
     *length = strlen(out);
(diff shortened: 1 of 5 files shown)


lmoureaux:
Worth mentioning that `QTextCodec` [will be deprecated in Qt6](https://bugreports.qt.io/browse/QTBUG-75665), so it's not like we're 100% done with text encoding issues :cry: .

We could remove transliteration support as well, it's always `false` for the Qt client.

mir3x:
Seems they changed their mind:
https://doc-snapshots.qt.io/qt6-dev/internationalization.html
find_package(Qt6 COMPONENTS Core5Compat REQUIRED)
target_link_libraries(mytarget PUBLIC Qt6::Core5Compat)

lmoureaux:
> Seems they changed their mind:
> https://doc-snapshots.qt.io/qt6-dev/internationalization.html
> find_package(Qt6 COMPONENTS Core5Compat REQUIRED)
> target_link_libraries(mytarget PUBLIC Qt6::Core5Compat)
> 
> That new QStringEncoder/Decoder looks eazy anyway.

The `xxCompat` is only provided for the first releases and will disappear after that. Was just pointing out that sadly another rewrite will be needed.

mir3x:
Hmm, some bugs in japanese in russian are not related to that patch or patches from few days back

PR #143, opened 2020-12-04
New theme + theme changes

mir3x:
It might be uglier than it was but I tired of themes for now.


lmoureaux:
Missing tree icon? 
image

lmoureaux:
Contrast in the tech tree is too low.

PR #155, opened 2020-12-07
Citybar

mir3x:


lmoureaux:
I get rendering artifacts for AIs:
image
Also, there is some unused space on the left where the growth bar should be.

PR #160, opened 2020-12-09
I fixed everything !!!

mir3x:


ai/aitraits.h
```diff
@@ -10,19 +10,13 @@
     /o)  (o/\ \_                General Public License along with Freeciv21.
     \_____/ /                     If not, see https://www.gnu.org/licenses/.
       \____/        ********************************************************/
-#ifndef FC__AITRAITS_H
-#define FC__AITRAITS_H
+#pragma once
 
 /* common */
 #include "traits.h"
```
lmoureaux, line 18: Still two consecutive empty lines? (And it repeats in several locations)

mir3x: I removed only 3 or more empty lines, u can use my utility to remove 2 
https://github.com/mir3x/header2pragma
there is remEmptyLines

PR #170, opened 2020-12-12
Tileset qstrings - removed *(s)printf*, replaced many chars with QStrings

mir3x:


lmoureaux:
I bet you'll get new clazy warnings because `QString` is too big to be passed by value (or because it has a non-trivial copy constructor because of the reference counting).

mir3x:
> I bet you'll get new clazy warnings because `QString` is too big to be passed by value (or because it has a non-trivial copy constructor because of the reference counting).

I already run clazy and did his suggestions :1st_place_medal:

PR #181, opened 2020-12-13
Added city unit list icons style for themes

mir3x:
[skip travis][skip travis][skip travis][skip travis][skip travis][skip travis][skip travis][skip travis][skip travis][skip travis][skip travis]


lmoureaux:
I think that `[skip travis]` should be a single line in the commit message

PR #183, opened 2020-12-13
Codefactory more fixes

mir3x:


mir3x:
29 more fixed, now u fix 100 :+1:

PR #185, opened 2020-12-14
Seperators

mir3x:
Replaces strvec directors to QStringLists, 
Now mingw freeciv client starts on windows


utility/shared.cpp
```diff
@@ -795,9 +795,7 @@ static char *expand_dir(char *tok_in, bool ok_to_free)
 static QStringList *base_get_dirs(const char *dir_list)
 {
   QStringList *dirs = new QStringList;
-
-  *dirs = QString(dir_list).split(QDir::listSeparator(),
-                                  QString::SkipEmptyParts);
+  *dirs = QString(dir_list).split(":", QString::SkipEmptyParts);
```
lmoureaux, line 798: How does it work with paths like `C:\Program Files\Freeciv`?

PR #186, opened 2020-12-15
Python codeFactored and autopeped8 with deafult values

mir3x:
We deserve A in CodeFactor  :1st_place_medal:


lmoureaux:
Bless autopep8

PR #188, opened 2020-12-15
Fix reconnecting to local server

mir3x:
It didnt break anything, 
I have no idea what I disconnected with client.conn.sock->disconnect(); but it's crucial to fix.


mir3x:
Codacy failed reconnecting !!!

PR #204, opened 2020-12-21
Start position scenario fix

mir3x:


common/map.cpp
```diff
@@ -1657,7 +1656,7 @@ struct startpos *map_startpos_get(const struct tile *ptile)
 bool map_startpos_remove(struct tile *ptile)
 {
   bool ret;
-  struct tile *prtile = static_cast<tile *>(tile_hash_key(ptile));
+  struct tile *prtile = static_cast<tile *>(ptile);
```
lmoureaux, line 1659: No need to cast

PR #219, opened 2020-12-23
Coverity again

mir3x:
We can get under 100 soon

### DIFF
--- a/common/featured_text.cpp
+++ b/common/featured_text.cpp
@@ -258,6 +258,7 @@ static bool text_tag_init_from_sequence(struct text_tag *ptag,
       ptag->color.background[0] = '\0';
     }
     return true;
+    break;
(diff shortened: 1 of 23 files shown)


common/featured_text.cpp
```diff
@@ -258,6 +258,7 @@ static bool text_tag_init_from_sequence(struct text_tag *ptag,
       ptag->color.background[0] = '\0';
     }
     return true;
+    break;
```
lmoureaux, line 261: break after return? unreachable statement will kick in

mir3x: here is the same and never kicked in:
bool is_player_phase(const struct player *pplayer, int phase)
{
  switch (game.info.phase_mode) {
  case PMT_CONCURRENT:
    return true;
    break;

PR #234, opened 2021-01-01
Add extra time information to timeout sidebar

mir3x:
If timeout > 3600 show time (hours + mins) change
If day differs then show also day

### DIFF
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -1717,8 +1717,8 @@ const QString format_duration(int duration)
   if (duration > 3600) {
     QDateTime time = QDateTime::currentDateTime();
     QDateTime tc_time = time.addSecs(duration);
-    QString day_now = time.toString("ddd ");
-    QString day_tc = tc_time.toString("ddd ");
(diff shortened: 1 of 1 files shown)


client/text.cpp
```diff
@@ -1712,7 +1713,16 @@ const QString format_duration(int duration)
...
   }
+  // Show time if there is more than 1hour left
+  if (duration > 3600) {
+    QDateTime time = QDateTime::currentDateTime();
+    QDateTime tc_time = time.addSecs(duration);
+    QString day_now = time.toString("ddd ");
+    QString day_tc = tc_time.toString("ddd ");
 
+    str += QStringLiteral("\n") + ((day_now != day_tc) ? day_tc : "")
+           + tc_time.toString("hh:mm");
```
lmoureaux, line 1724: From the docs:

> Note: If localized month and day names are desired, please switch to using QLocale::system().toString() as QDateTime methods shall change to use English (C locale) names at Qt 6.

Do it now to avoid surprises later?

PR #259, opened 2021-02-07
Fix 3.13 cmake on debian buster

wackyjacky101:


lmoureaux:
Does it work with [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)?

PR #262, opened 2021-02-11
Create save/scenario dir if doesnt exists

mir3x:


lmoureaux:
Actually there's no need to check anything, `mkpath()` simply won't do anything if the directory exists. Could display a message if `mkpath` fails (returns `false`)

Issue #265, opened 2021-02-14
KArchieve SEGV


mir3x:
Just opening network page on ubuntu 20.04

Thread 1 "freeciv-qt" received signal SIGSEGV, Segmentation fault.
0x00007ffff7c23ff8 in KCompressionDevice::error() const () from /lib/x86_64-linux-gnu/libKF5Archive.so.5
(gdb) bt
#0  0x00007ffff7c23ff8 in KCompressionDevice::error() const () at /lib/x86_64-linux-gnu/libKF5Archive.so.5
#1  0x00005555559c7629 in inf_close_partial(inputfile*) (inf=0x7fffac008cc0) at /home/mirex/git/freeciv21/utility/inputfile.cpp:261
#2  0x00005555559c7a22 in inf_close(inputfile*) (inf=0x7fffac008cc0) at /home/mirex/git/freeciv21/utility/inputfile.cpp:291
#3  0x00005555559acf37 in secfile_from_input_file(inputfile*, QString const&, QString const&, bool)
    (inf=0x7fffac008cc0, filename=..., section=..., allow_duplicates=true) at /home/mirex/git/freeciv21/utility/registry_ini.cpp:512
#4  0x00005555559ad81b in secfile_from_stream(QIODevice*, bool) (stream=0x7fffac008d40, allow_duplicates=true)
    at /home/mirex/git/freeciv21/utility/registry_ini.cpp:583
#5  0x00005555557df181 in parse_metaserver_data(QIODevice*) (f=0x7fffac008d40) at /home/mirex/git/freeciv21/client/servers.cpp:255
#6  0x00005555557df9cc in meta_read_response(server_scan*) (scan=0x555556664a30) at /home/mirex/git/freeciv21/client/servers.cpp:371
#7  0x00005555557dfb54 in metaserver_scan(void*) (arg=0x555556664a30) at /home/mirex/git/freeciv21/client/servers.cpp:400
#8  0x00005555557e0505 in server_scan_begin(server_scan_type, void (*)(server_scan*, char const*))
    (type=SERVER_SCAN_GLOBAL, error_func=0x5555557767b2 <server_scan_error(server_scan*, char const*)>)

mir3x:
Restoring qobject cast fixes 
-  if (reinterpret_cast<KFilterDev *>(inf->fp)) {
-    error = reinterpret_cast<KFilterDev *>(inf->fp)->error() != 0;
+  if (qobject_cast<KFilterDev *>(inf->fp)) {
[...]

PR #270, opened 2021-02-15
Reinterpret cast ->dynamic cast to avoid crash due to debian compile …

mir3x:
…fail

Works on debian and doesnt crash on other distros.


mir3x:
Fixes #265

Issue #297, opened 2021-02-21
Build errors following INSTALL instructions


jwrober:
If you follow the directions in the current INSTALL file, it doesn't work.

lmoureaux:
How does it break?

jwrober:
In Debian section the `sudo apt install build-dep-freeciv` is invalid for starters. It should be `sudo apt build-dep freeciv` and then `sudo apt install <packages>`. And then also the next section of deps to be installed has a number of duplicates.

Also testing build on a fresh install of Debian produces some other issues, which might need to be captured elsewhere.

lmoureaux:
This works for Ubuntu 20.04: [windows/Dockerfile](https://github.com/longturn/freeciv21/blob/master/windows/Dockerfile)

lmoureaux:
@jwrober did #302 fix everything reported here?

jwrober:
Yes I believe so. Other issues are potential build related per environment so probably handled in other issues.

PR #302, opened 2021-02-21
INSTALL Fixes

jwrober:
For #297

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -146,6 +146,7 @@ Start with ensuring your have a source repository (deb-src) turned on in apt sou
      qt5-default \
      libkf5archive-dev \
      libmagickwand-dev \
+     libsdl2-mixer-dev \
      libunwind-dev \
      libdw-dev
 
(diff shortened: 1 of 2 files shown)


INSTALL
```diff
@@ -112,40 +113,41 @@ Once the build files have been written, then compile with this command:
...
+
+   sudo apt install git \
      cmake \
      ninja-build \
      python3 \
-     gettext \
      qt5-default \
      libkf5archive-dev \
-     libsqlite3-dev \
-     libsdl2-mixer-dev \
```
lmoureaux, line 146: `sdl2-mixer` is needed (but optional) for sound support.

jwrober: That is why I add it to the list of deps to install after the base `build-dep freeciv` step.

lmoureaux: Doesn't your patch remove it?

jwrober: Good catch. Fixed

PR #315, opened 2021-03-05
Add 50% penalty for changing production when you buy but didnt build

### LINKED ISSUE #310: Forbid buying a settler at size 1 then using the shields
> you can buy the setller with 30/40 shields, which is cheapier then buy, let say a caravan for 30/50 shields

This can be fixed by forcing a loss of half/all the built shields when changing production.

mir3x:
Fix #310, heavily untested :dango:


mir3x:
hmm, player still could buy 2 turns after :D

PR #317, opened 2021-03-05
❤️ Penalize buying without building

### LINKED ISSUE #310: Forbid buying a settler at size 1 then using the shields

mir3x:
Fix #310 
Set 50% penalty when buying units/improvements and changing to somethign else later.

### DIFF
--- a/client/packhand.cpp
+++ b/client/packhand.cpp
@@ -776,6 +776,7 @@ void handle_city_info(const struct packet_city_info *packet)
 
   pcity->airlift = packet->airlift;
   pcity->did_buy = packet->did_buy;
+  pcity->did_buy_production =  packet->did_buy_production;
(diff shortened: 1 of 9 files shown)


lmoureaux:
Why did you remove the penalty for buildings?

lmoureaux:
In addition, this is changing the network protocol so a netcap bump is needed.

mir3x:
> Why did you remove the penalty for buildings?

Because some ppl might buy wonders and they would be very angry if they switch to another wonder and lose production
and I was to lazy to disticint buildings from wonders

lmoureaux:
I see, thanks for the explanation!

PR #343, opened 2021-03-17
NSIS Package installs i18n files

### LINKED ISSUE #335: Windows/Msys2 - i18n files not installed
This is I think due to the way we do translations on Windows/Msys2. There is no `install()` in the `CMakeLists.txt files`. They should end up in the `$Install_Dir/share/locale` directory at package creation time and populate the `translations` component.

This issue should cover two tasks

- [ ] Get the i18n files to install for Windows/Msys2
- [ ] The files should be copied and compiled as part of the translations component for the NSIS installer

jwrober:
Small update to get the NSIS package to contain the locale files. Fixes #335


cmake/GettextTranslate.cmake
```diff
@@ -288,6 +288,7 @@ macro(GettextTranslate)
     install(FILES ${GMO_FILE_NAME} DESTINATION
       ${LOCALEDIR}/${lang}/LC_MESSAGES
       RENAME ${MAKEVAR_DOMAIN}.mo
+	  COMPONENT translations
```
lmoureaux, line 291: Tab :stuck_out_tongue_closed_eyes:

lmoureaux (changes requested):
Looking good! I hate tabs because Github doesn't agree with your editor on how long they should be :smile:

cmake/GettextTranslate.cmake
jwrober: Fixed. Agree I gotta figure out my editor on Windows.

lmoureaux (approved):
Can merge when the pipeline succeeds

Issue #331, opened 2021-03-08
Generate .desktop and AppStream files


lmoureaux:
These are required to show menu entries in Linux and BSD desktop environments. One can start from the Freeciv ones still found in the `bootstrap` folder (I find this folder name meaningless so bonus point if you take them out).

The CMake function [`file(GENERATE OUTPUT ...)`](https://cmake.org/cmake/help/latest/command/file.html#generate) or [`file(CONFIGURE OUTPUT ...)`](https://cmake.org/cmake/help/latest/command/file.html#configure) can be used to fill in paths with absolute paths from [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) paths. This should only be done if the target is neither Windows nor Mac (nor Android).

lmoureaux:
Here's the [AppStream spec](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps). Freeciv recently renamed their files to something like `org.freeciv.freeciv.desktop` to avoid being kicked out from distros, we should do the same with `net.longturn`.

jwrober:
I am using the following IDs. I actually found that Spec from Debian's website on my own yesterday ;)
```
net.longturn.freeciv_modpack_qt
net.longturn.freeciv_qt
net.longturn.freeciv_ruledit
net.longturn.freeciv_server
```
I don't think we need AddStream files for ruleup and the CLI Modpack Installer. Those are more special purpose for experts that would know they are there to be used.

jwrober:
Closed

PR #344, opened 2021-03-17
Rename bootstrap/ to dist/. Update files and add content for #331

jwrober:
Renamed the bootstrap directory to dist. Updated files that we may or
may not need in the future that referred to bootstap directory. Added
content to the desktop and appdata files for issue #331.


jwrober:
I'm not sure why this PR pulled in the last two commits. My fork was same as master when I did the last commit. Hopefully won't cause an issue.

lmoureaux:
I think the git issues are due to my use of "squash and merge" for the last PR. This rewrites history by merging the commits, which made sense for your merge requests since there was no clear separation between commits.

I think you did `git pull` locally. It saw one commit on Github and two in your local branch, so it created a merge commit to unite the two trees. You can prevent it by passing `--ff-only` and there's also an option to make it the default. After pulling, you committed your change on top, creating the four commits visible here (2 old+merge+1).

I avoid this by never using my local `master` branch. Instead, I create a new branch for every pull request and delete it afterwards. I start by updating the remote without pulling:
```
git fetch upstream
```
The command to create the branch is then:
```sh
git checkout -b feature/something upstream/master
```
This ensures that you always start from a clean upstream `master`.

To delete it afterwards, do the following after switching to a new branch:
```sh
git branch -d feature/something
```

I'm going to use "squash and merge" again to wipe the merge out of history. The result will be a single commit on `master`.

Issue #350, opened 2021-03-21
Create a modpack repository


lmoureaux:
Once #308 is fixed and we have a first version of the modpack format, we should think about creating a repository for modpacks compatible with Freeciv21. The general idea I'd like to go for is summarized below:

* Create a Github repository under the Longturn organization with the modpack list and control files for individual modpacks.
* Change the default modpack URL in the installer to download the files directly from Github. (Note that this might require a privacy policy… this is probably a more general question.)
* In the README of the new repo, document how to add new modpacks and update existing ones. One key point is that URLs should point to a specific commit for safety reasons.
* Add existing modpacks (Aviation already has a fc21 variant).
* Accept PRs for new content!

PR #351, opened 2021-03-21
JSON modpack definitions

### LINKED ISSUE #308: Use JSON for the modpack installer
The modpack installer is backed by Freeciv's old INI-like file format. JSON is ubiquitous on the Internet and Qt has built-in support for it. It is a safer (no custom parser) and better supported format.

Since there isn't a lot of FCMP content in the wild and the files are usually small, changing the format won't impose too much of a burden on content makers. Hopefully it will also help setting up a Longturn-backed FCMP server.

lmoureaux:
This changes the format of modpack definition files in a backward incompatible way. The versioning of the format itself is decoupled from releases because it will likely be rather stable.

The documentation has also been updated.

Closes #308

### DIFF
--- a/doc/README.modpack_installer
+++ b/doc/README.modpack_installer
@@ -16,34 +16,33 @@ A "modpack" can consist of one or more of:
  - 'tilesets' for game graphics
  - 'soundsets' or 'musicsets' for sound effects and in-game music
 
-Depending which client you installed, you will probably have one modpack
-installer tool called something like 'freeciv-mp-gtk3' or
-'freeciv-mp-qt' -- these different versions have the same features.
-There may also be a command-line-only tool called 'freeciv-mp-cli',
+They can be installed using the modpack installer tool
+'freeciv-modpack-qt' that comes with the standard installation.
+There is also a command-line-only tool called 'freeciv-modpack',
 which is mainly useful for headless game servers.
 
-In standard Freeciv builds, when you start the graphical modpack
-installer, it will show a list of modpacks from modpack.freeciv.org,
-curated by the Freeciv developers. You can select one and press
-"Install modpack"; the tool will download the files for the selected
-modpack, and any other modpacks it depends on, and install them ready
-for the main Freeciv programs to use.
+In standard Freeciv21 builds, when you start the graphical modpack
+installer, it will show a list of modpacks curated by the Freeciv21
+developers. You can select one and press "Install modpack"; the tool
+will download the files for the selected modpack, and any other
+modpacks it depends on, and install them ready for the main Freeciv21
+programs to use.
 
 You can also point the installer at modpacks or lists on other servers:
 
  * If someone has given you an individual modpack URL (ending in
-   '.mpdl'), you can paste it into the "Modpack URL" box, and when
+   '.json'), you can paste it into the "Modpack URL" box, and when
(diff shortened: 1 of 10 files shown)


lmoureaux:
CodeFactor: Looks like deeper refactoring is in order.

doc/README.modpack_installer
```diff
@@ -279,33 +254,29 @@ The [info] section has overall control information:
...
-     May be relative to the .mpdl file (starting with "./"), or
+ - "base_url":
+     URL to prepend to the "src" filenames in the "files" list.
+     May be relative to the .json file (starting with "./"), or
      absolute (in which case the files can be on some web server
-     different to where the .mpdl file lives).
+     different to where the .json file lives).
 
-The [files] section lists the individual files comprising your modpack.
+The "files" list defineds the individual files comprising your modpack.
```
jwrober, line 263: Defines? Looks like a spelling error.

doc/README.modpack_installer
```diff
@@ -340,22 +311,31 @@ Some advice on the structure of files in modpacks:
    of "rulesetdir <name>", as this is needed for the server to enumerate
    the available rulesets.
 
-The [dependencies] section is optional, and not normally needed
-(although "Group" type modpacks should have a [dependencies] section
-and no [files] section). If your modpack requires another one that can
-also be installed separately (for instance, a scenario that needs a
-specific ruleset), you can specify it here. After the header, each
-line consists of:
+In some cases, a modpacks may depend on other modpacks, for instance if
```
jwrober, line 314: a modpack may depend - use of plural

doc/README.modpack_installer
```diff
@@ -209,64 +188,60 @@ here, and followed by one line for each modpack:
...
-{ "modpack", "URL", "type", "version"
-  "Ancients-tiles", "./ancients-tiles.mpdl", "Tileset", "3.1-0"
+.json file.)
+
+Again, this is a file in standard JSON format. Its filename must end
+in ".json".
+
+Here is an example of a modpack control file:
+{
+  "info": {
```
jwrober, line 217: Should modpacks be fc21 version specific? They are "ish" with classic fc. There is a control code in the header that is read and parsed.

jwrober (changes requested):
Couple spelling/grammar issues + a question.

doc/README.modpack_installer
lmoureaux: Good question. I would like to keep backward compatibility as much as possible, but sometimes it's just too much work. I also kind of have a plan to get rid of spec files altogether (maybe using Lua, YAML or some other outsourced parser), which will probably induce some compatibility breach… can keep both versions available in the transition period.

I think the modpack index will be version specific using one branch per major version (see #350). It will require some maintenance but it's not so bad that the maintainers are aware of how much work it takes to port a modpack.

jwrober (approved):

jwrober:
I've approved what I know I can contribute to. Don't know enough C++ to weigh in on the code factor issues. looking forward to updating my SongOfDoom music set to this new JSON file format.

lmoureaux:
I'm going to give up on CodeFactor's complexity warnings. The code was awful, it's no better with JSON but that's not because of this PR.

PR #360, opened 2021-03-22
Updated flags of Yugoslavia

Corbeau-1:
"New" flags are actually the flags of Yugoslav state during the longest period of its existence AND its greatest prominence in international relations.


lmoureaux:
Could [yugoslavia.svg](https://github.com/longturn/freeciv21/blob/master/data/flags/yugoslavia.svg) also be updated to match? It will be the starting point if we make flags at a different size or render them for high density displays, for instance.

lmoureaux (approved):
Old flag: 1918 to ca. 1940 and 1992 to 2003, total between 33 and 38 years.
New flag: from ca. 1945 to 1992 which is 47-52 years and is the best known period of unified Yugoslavia.

So the change looks reasonable.

I'm going to merge the 4 commits to make history cleaner and avoid the short appearance of a file at the root of the tree.

PR #385, opened 2021-03-28
Added unit waittime to client

mir3x:
Quite a mess now, currently shows if unit waits under middle click


mir3x:
client always shows game.server.unitwaittime  as 0
I had to add also action_turn
Time will probably work as single int32, will test later.

mir3x:
Added time when unit can move in network - server sends that time in utc time
Added civstatus widget - small on top with gold /sci income and number waiting units if any.
Added table in units view - F2 - with list of all wating untis
Changed focusing units - not to focus waiting at all.

lmoureaux (commented):
The new status widget can be hard to read on light backgrounds such as desert, suggest adding a dark background or shadow (offset the text by (1,1) and draw it below in black).
image

The F2 dialog covers the status widget by default on my screen:
image

I don't understand if "waiting" is "waiting on uwt" or "waiting orders". I think "waiting on uwt" would be more useful in LT… or simply list all units without orders/not tired and use different colors.

Workers on auto (I know, it's bad) appear in the "waiting" list. I don't think it's updated on `/set unitwaittime X` (minor).

The "waiting" list doesn't always activate the correct unit when clicked.

Would it be hard to show UWT for the current unit be shown in the same panel as unit actions? A tileset icon would also be useful to see them on map (but that can for sure be left for later).

PR #406, opened 2021-04-04
Don't log unknown source locations in non-debug builds

### LINKED ISSUE #277: Disable source location in logs when not available
The Qt logging format set by Freeciv includes the source location. This is useless in builds without debug info, and produces ugly results:
```
[warning] freeciv-server (unknown:0) - Here's a warning!
```
The format string in non-debug build should be changed not to include the source location.

lmoureaux:
Closes #277.


lmoureaux:
CodeFactor once again failed to take notice.

PR #411, opened 2021-04-08
A port of WYSIWYG Patch

pranavtifr:
I have tried to keep the same commit pattern as ilkkachu inspite of certain commits being redundant since this has been extensively tested now.


pranavtifr:
I still have to port the wonders take effect the following turn patch, and fix the use-after-free bug. I just created this already so that others can playtest it if untill I get to the other 2 commits :)

lmoureaux:
Well it appears I can't find time to test this. Let's merge and handle any bugs later.

PR #412, opened 2021-04-10
unit shields set to normal size

Canik05:
the unit shield graphic was set to shields-large.spec which sometimes partially covered up units and distracted from the unit graphic itself. 
now set to the normal sized shields.spec the shield will be more 'out of the way'

### DIFF
--- a/data/amplio2.tilespec
+++ b/data/amplio2.tilespec
@@ -166,7 +166,7 @@ files =
   "misc/cursors.spec",
   "misc/overlays.spec",
   "misc/citybar.spec",
-  "misc/shields-large.spec",
+  "misc/shields.spec",
   "misc/editor.spec",
   "misc/icons.spec"
 


lmoureaux:
Not convinced for `amplio2` tbh. I've used it for years and the shield size never bugged me. Maybe I've just grown accustomed to it… It's for sure an option for custom tilesets.

Also, why `amplio2` and not `amplio` or `cimpletoon`?

Canik05:
oh, we were looking at the github, you mentioned mass scaling them down. then you said change it to shield.spec so I thought you mention here on github. (you are Louis94, right? xD)

I can do it in my custom tileset, sure. that'll work. I would still recommend the change though, a normal person won't know about custom ruleset. expect a few native users to complain, but in time I think they'll prefer the standard.

lmoureaux:
Let me get more voices in

jwrober (commented):
I like the proposed change. Having the shield a bit to the left does look better.

cgalik:
I like the bigger shield, as I can't even count the number of times I have gotten flags mixed up between players... larger shield = more easily to see the difference. 

That being said, I don't feel strongly about it and would be fine with the change, as it's not a _big_ deal...  get it not a "big" deal... hehe

Canik05:
There is an option to remove shield and then there is a large flag behind the unit instead. If you're getting nations mixed up.

lmoureaux:
Given how mixed feelings this change triggered, I'm going to close this PR and keep the flags as they are.

Issue #409, opened 2021-04-05
Bug with display of long goto paths


lmoureaux:
soundnfury on Discord:
> if you [g]o to tile and the path is more than 10 turns long, the two-digit numbers display with both digits in the same place

I guess this was with the Aviation tileset.

PR #415, opened 2021-04-11
Fixed numbers in goto 10-99

mir3x:
Its hack :D.
Doesnt work for numbers > 99
It uses the same icons as in city and everythign looks the same, but doesnt behave the same.
Just replaced hundreds numbers with tens.


mir3x:
closes #409

Issue #423, opened 2021-04-12
Weekly Travis build failed on OS X


lmoureaux:
CMake could not find KF5Archive. Doesn't look like there was an update to the Homebrew bottle recently.

https://travis-ci.org/github/longturn/freeciv21/jobs/766779115

lmoureaux:
Different log here, CMake cannot find Ninja: https://travis-ci.org/github/longturn/freeciv21/jobs/766814044

lmoureaux:
```
Upgrading cmake formula. It is installed but not up-to-date.
==> Upgrading 1 outdated package:
cmake 3.18.4 -> 3.20.1
==> Upgrading cmake 3.18.4 -> 3.20.1 
==> Downloading https://homebrew.bintray.com/cmake-3.20.1.catalina.bottle.tar.gz
curl: (22) The requested URL returned error: 404 Not Found
Trying a mirror...
==> Downloading https://ghcr.io/v2/homebrew/core/cmake-3.20.1.catalina.bottle.tar.gz
==> Downloading from https://github.com/-/v2/packages/container/package/homebrew%2Fcore%2Fcmake-3.20.1.catalina.bottle.tar.gz
curl: (22) The requested URL returned error: 404 
Error: Failed to download resource "cmake"
Download failed: https://ghcr.io/v2/homebrew/core/cmake-3.20.1.catalina.bottle.tar.gz
Installing cmake has failed!
```

Installing packages from Homebrew fails.

lmoureaux:
[Well, let's use a more recent OS X.](https://travis-ci.community/t/the-default-osx-image-xcode9-4-is-now-unusable-because-macos-10-13-is-eol/10628)

PR #424, opened 2021-04-12
Allow an arbitrary number of tile output sprites

### LINKED ISSUE #345: The client clips the output values for tileset at 9
We need to fix this because in LTT we go beyond 9 in some cases.

lmoureaux:
Some rulesets (eg LTT) go above the hard-coded maximum of 10. Let the artist
specify as many sprites as he wants.

Closes #345.


lmoureaux:
Bypassing the build failure on Mac (#423)

PR #425, opened 2021-04-12
Avoid text wrapping in the city bar

### LINKED ISSUE #400: Long city names can wrap
The bar should be long enough to hold them:
image

lmoureaux:
Add one extra pixel to give some more room.

Closes #400.


lmoureaux:
Going to bypass Travis again because of #423

Issue #321, opened 2021-03-07
Travis Mac OS build is failing


lmoureaux:
From [the logs](https://travis-ci.org/github/longturn/freeciv21/jobs/761698937):
```
In file included from ../client/attribute.cpp:17:
In file included from /usr/local/opt/qt5/lib/QtCore.framework/Headers/QHash:1:
In file included from /usr/local/opt/qt5/lib/QtCore.framework/Headers/qhash.h:44:
In file included from /usr/local/include/QtCore/qchar.h:43:
/usr/local/include/QtCore/qglobal.h:667:43: error: no template named 'is_arithmetic_v' in namespace 'std'; did you mean 'is_arithmetic'?
         typename = std::enable_if_t<std::is_arithmetic_v<T> && std::is_arithmetic_v<U> &&
                                     ~~~~~^
/Applications/Xcode-12.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/type_traits:994:50: note: 'is_arithmetic' declared here
template <class _Tp> struct _LIBCPP_TEMPLATE_VIS is_arithmetic
                                                 ^
```
And many more.

[`std::is_arithmetic_v`](https://en.cppreference.com/w/cpp/types/is_arithmetic) requires C++ 17 while we set C++ 14. If C++ 17 is supported on every platform we build for, the simplest fix is probably to bump the C++ version.

PR #426, opened 2021-04-12
Don't list tilesets, sound sets and music sets twice

### LINKED ISSUE #399: Tileset/soundset are listed twice when they appear twice in the path
This is the case for instance when they are in `./data` and the final install location. Only the first appearance in the path should be taken into account.

lmoureaux:
There was a bug in `fileinfoname` and an oversight in the "Load another tileset" box.

Closes #399.


lmoureaux:
Bypassing Travis again (#321)

Issue #269, opened 2021-02-15
The city dialog should display how many citizens can be made content with existing buildings


lmoureaux:
Currently, the citizens made content by buildings are shown, but one can't see how much the city can grow before the first unhappy citizen appears. This makes celebration harder.

Initial proposal is to add it to the relevant tooltip in the Happiness tab. In the long run some more graphical depiction could be added.

pranavtifr:
Lol. I just saw this. Anyways, relevant thing I did with the city-report tab

https://github.com/pranavtifr/freeciv/commit/ea17ba086b7602472d8b364dd92bf9ad0fc9b308

PR #428, opened 2021-04-13
happyCityTooltip

mir3x:
Added tooltip to city dialog -> buildings in citizens about possible grow without requiring more luxuries.
Might not work in some cases


mir3x:
close #269

mir3x:
Somehting wrong

Issue #337, opened 2021-03-12
Windows/Msys2 - Install() duplicates the .dll files into / and /bin directories


jwrober:
This is not a huge issue per se, but from an efficiency standpoint we don't need the `.dll` files in the root install directory and then a copy of the same ones in a `/bin` directory too.  The `exe`'s call/reference the ones in the root install directory.

PR #433, opened 2021-04-16
Discover dependencies dynamically

### LINKED ISSUE #396: Windows installer: discover dependencies dynamically
The current Windows installer uses a hard-coded list of dependencies which will need maintenance to deal with new versions. It is also not dynamic when configuration options are changed (eg sound is not enabled).

The list should be based on what the installed programs actually need, as determined with `file(GET_RUNTIME_DEPENDENCIES)` or something similar.

jwrober:
Fixes #396, #337

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,51 +34,7 @@ if(WIN32 OR MSYS OR MINGW)
   set(CMAKE_INSTALL_BINDIR ".")
   set(PROJECT_NAME "data")
   set(CMAKE_INSTALL_DOCDIR "${CMAKE_INSTALL_DATAROOTDIR}/doc/")
-
   get_filename_component(MINGW_PATH ${CMAKE_CXX_COMPILER} PATH)
-  set(CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS
-    ${MINGW_PATH}/libbrotlicommon.dll
-    ${MINGW_PATH}/libbrotlidec.dll
-    ${MINGW_PATH}/libbz2-1.dll
-    ${MINGW_PATH}/libdouble-conversion.dll
-    ${MINGW_PATH}/libFLAC-8.dll
-    ${MINGW_PATH}/libfreetype-6.dll
-    ${MINGW_PATH}/libgcc_s_seh-1.dll
-    ${MINGW_PATH}/libglib-2.0-0.dll
-    ${MINGW_PATH}/libgraphite2.dll
-    ${MINGW_PATH}/libharfbuzz-0.dll
-    ${MINGW_PATH}/libiconv-2.dll
-    ${MINGW_PATH}/libicudt68.dll
-    ${MINGW_PATH}/libicuin68.dll
-    ${MINGW_PATH}/libicuuc68.dll
-    ${MINGW_PATH}/libintl-8.dll
-    ${MINGW_PATH}/libKF5Archive.dll
-    ${MINGW_PATH}/liblzma-5.dll
-    ${MINGW_PATH}/libmpg123-0.dll
-    ${MINGW_PATH}/libogg-0.dll
-    ${MINGW_PATH}/libopus-0.dll
-    ${MINGW_PATH}/libopusfile-0.dll
-    ${MINGW_PATH}/libpcre-1.dll
-    ${MINGW_PATH}/libpcre2-16-0.dll
-    ${MINGW_PATH}/libpng16-16.dll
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -147,12 +103,33 @@ install(
...
+      POST_EXCLUDE_REGEXES "C:[\\\\/][Ww][Ii][Nn][Dd][Oo][Ww][Ss][\\\\/].*"
+      EXECUTABLES
+        "${CMAKE_INSTALL_PREFIX}/freeciv21-*.exe"
+      )
+      message(STATUS "Installing library dependencies for freeciv21 executables...")
+      foreach(_file ${r_deps})
+        file(INSTALL
+          DESTINATION "${CMAKE_INSTALL_PREFIX}"
+          TYPE SHARED_LIBRARY
+          FILES "${_file}"
```
lmoureaux, line 129: Can't we avoid the loop and pass `${r_deps}` directly here? `FILES` makes me think that this should be possible.

CMakeLists.txt
```diff
@@ -34,51 +34,7 @@ if(WIN32 OR MSYS OR MINGW)
   set(CMAKE_INSTALL_BINDIR ".")
   set(PROJECT_NAME "data")
   set(CMAKE_INSTALL_DOCDIR "${CMAKE_INSTALL_DATAROOTDIR}/doc/")
-
   get_filename_component(MINGW_PATH ${CMAKE_CXX_COMPILER} PATH)
```
lmoureaux, line 37: I think this line is no longer needed.

lmoureaux (changes requested):
Looks very neat!! With this we won't need to keep track of library versions anymore :v: 

I have two small comments on the code.

CMakeLists.txt
jwrober: I got error messages when doing this so went with the loop to correct

CMakeLists.txt
jwrober: Still need the 3 set() and the get_filename_component()

lmoureaux: Oh you mean for the Qt plugins? Right...

CMakeLists.txt
lmoureaux: Fair enough

jwrober: second commit forthcoming...

CMakeLists.txt
```diff
@@ -147,12 +103,33 @@ install(
...
+  install(CODE [[
+    message(STATUS "Collecting dependencies for freeciv21 executables...")
+
+    # Take a variable that is available at "install" time and repurpose
+    string(REGEX REPLACE "objdump.exe" "" MINGW_PATH ${CMAKE_OBJDUMP})
+
+    # Function to analyze the third party dll files linked to the exe's
+    #   Uses the repurposed variable from above to tell the function where
+    #   the dll files are located. Ignores dll's that come with Windows.
+    file(GET_RUNTIME_DEPENDENCIES
```
lmoureaux, line 115: One last comment: this command is new in 3.16 so we need to require it on line 1 (but please keep 3.12 support for Linux, as it's what the Longturn server has).

jwrober: Handled in second commit

CMakeLists.txt
jwrober: yes

lmoureaux (approved):

lmoureaux:
Ubuntu builds failed:
```
$ cmake . -B build -G Ninja -DCMAKE_INSTALL_PREFIX=$PWD/build/install
CMake Error at CMakeLists.txt:37 (project):
  VERSION not allowed unless CMP0048 is set to NEW
-- Configuring incomplete, errors occurred!
```

CMakeLists.txt
```diff
@@ -152,7 +152,8 @@ if(WIN32 OR MSYS OR MINGW)
     DIRECTORY ${MINGW_PATH}/../share/qt5/plugins/platforms
     DESTINATION ${CMAKE_INSTALL_BINDIR}/plugins
     COMPONENT freeciv21
-    FILES_MATCHING PATTERN *.dll)
+    FILES_MATCHING PATTERN *.dll
+    COMPONENT freeciv21)
```
lmoureaux, line 156: Isn't `COMPONENT` passed twice now?

jwrober: oh crud it is

jwrober:
Ok I'm going to let this go. I think its done.

Issue #443, opened 2021-04-18
Modpack installer for Windows TLS initialization failed


jwrober:
The Linux compiled version works as expected. Top picture is Windows, bottom is Linux

![image](https://user-images.githubusercontent.com/23037938/115166266-e0eb4700-a077-11eb-8e65-b4ffadfaf5aa.png)

jwrober:
This might be part of the problem

```[warning] freeciv21-modpack-qt - QSslSocket::connectToHostEncrypted: TLS initialization failed```

lmoureaux:
Need to install the CA certificates.

jwrober:
With the NSIS or is this a Windows thing?

jwrober:
Classic FC installs `/ssl/certs/ca-bundle.crt`. I see a similar file in MSys2 at `/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt`. Need to figure out how to inlcude in CPack.

lmoureaux:
It seems that we'll also need some supporting code to load the certificates into the [`QSslConfiguration`](https://doc.qt.io/qt-5/qsslconfiguration.html).

Issue #339, opened 2021-03-12
NSIS Installer - Shortcuts on Start Menu point to .cmd.exe files, should be .cmd


jwrober:
This is probably related to the way that CPack handles the `CPACK_PACKAGE_EXECUTABLES` variable that is used to create the shortcuts. Any text you put in there it will tack the extra `.exe` on the and assuming its an actual `.exe`.  Current code is 

`set(CPACK_PACKAGE_EXECUTABLES "freeciv-qt.cmd;Freeciv21 Qt Client;freeciv-modpack-qt.cmd;Freeciv21 Qt Modpack Installer")`

Creates broken shortcut off Start Menu. Related to #338 

![image](https://user-images.githubusercontent.com/23037938/110969198-04ef9600-831e-11eb-8940-5beb47d5359d.png)

Issue #336, opened 2021-03-12
NSIS Installer - GUI Screen 5 - "custom" component is listed twice


jwrober:
Not a huge issue, but is a cosmetic annoyance.

![image](https://user-images.githubusercontent.com/23037938/110967242-c8bb3600-831b-11eb-8fa3-b6e8857dd08a.png)

Issue #334, opened 2021-03-12
NSIS Installer - GUI Screen 5 - freeciv21 component not on top, tools is


jwrober:
I think the main required installed component (freeciv21) should be first in the list and then optional components like the tools and eventually translations should be below.

![image](https://user-images.githubusercontent.com/23037938/110966257-b12f7d80-831a-11eb-92f0-e2c2dac9ded8.png)

PR #451, opened 2021-04-27
Windows NSIS Installer Updates

jwrober:
Collection of changes closing the following issues: #334, #336, #339, #443


lmoureaux:
I don't understand where exactly the file `cmake/NSIS.template.in` is used. It doesn't seem to appear in `CMakeLists.txt` or `CPackConfig.cmake`. It it implicitly used by the CPack NSIS backend?

jwrober:
Yes CPack calls the `makensis` program and inside copies the template to the appropriate directory and then populates the variables.

PR #459, opened 2021-05-06
Fix popup when disembarking 

### LINKED ISSUE #439: Unit disembark off ship not able to with mouse, only keyboard
I have a unit that is loaded on a ship. I activate the unit, press 'g' and then click the tile I want the unit to go to. In below screen shot I am clicking on the tile to the right of the Explorer in Trireme (shield) and nothing happens.

image

If I use the right arrow on the keypad, I get the menu of what to do, which is what I would expect with the mouse click.

image

mir3x:
Seems out of place, might introduce some tranport bugs, didnt observed any so far

### DIFF
--- a/client/goto.cpp
+++ b/client/goto.cpp
@@ -1921,8 +1921,11 @@ void send_goto_route()
     if (gui_options.popup_last_move_to_allied
         && goto_last_order == ORDER_LAST
         && ((is_allied_city_tile(tgt_tile, client_player())
-             || is_allied_unit_tile(tgt_tile, client_player()))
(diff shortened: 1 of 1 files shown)


lmoureaux:
I want to test this before merging. Or maybe @jwrober ?

Closes #439

jwrober:
I would like the opportunity to test this. Let me merge the diff to my local repo and test a new compile.

jwrober:
Ok I was able to place some units out of a transport exactly where I wanted them I think this fixes the issue.

lmoureaux:
Thanks for the check!

PR #463, opened 2021-06-06
Combat_Rounds-related enhancements

ec429:
With the ability to limit Combat_Rounds, it now makes sense to have full combat between units even when the defender's defence strength or FirePower is zero, as it might still survive if the attacker runs out of rounds.
Against a defender with D=0, you simply need to rack up enough rounds to take out its HP, because every shot hits.
Against a defender with D>0 but FP=0, combat becomes a bit like bombardment: you might not hit every round (you have to roll A/(A+D)), but you don't take damage when you miss.  This represents a unit that's unarmed but difficult to hit (e.g. high-flying aeroplane).

I have tested this patch with the following cases:
* A1 FP1 CR3 vs D0 FP1 HP4.  Defender took 3 HP of damage.  Attacker took none.
* A1 FP1 CR3 vs D1 FP0 HP4.  Defender sometimes took 1 HP of damage, sometimes 2.  (Presumably 0 and 3 are also possible, but didn't happen in my small sample.)  Attacker took no damage.

I haven't (yet) tested to verify the right thing happens if Combat_Rounds are not limited, or if attacker FP is 0.

### DIFF
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -298,12 +298,14 @@ void unit_versus_unit(struct unit *attacker, struct unit *defender,
   player_update_last_war_action(plr1);
   player_update_last_war_action(plr2);
 
-  if (attackpower == 0) {
-    *att_hp = 0;
-  } else if (defensepower == 0) {
-    *def_hp = 0;
-  }
   max_rounds = get_unit_bonus(attacker, EFT_COMBAT_ROUNDS);
+  if (max_rounds <= 0) {
+    if (attackpower == 0) {
+      *att_hp = 0;
+    } else if (defensepower == 0) {
+      *def_hp = 0;
+    }
+  }
   for (rounds = 0; *att_hp > 0 && *def_hp > 0
                    && (max_rounds <= 0 || max_rounds > rounds);
        rounds++) {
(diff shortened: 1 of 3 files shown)


server/unittools.cpp
```diff
@@ -298,12 +298,14 @@ void unit_versus_unit(struct unit *attacker, struct unit *defender,
...
-    *def_hp = 0;
-  }
   max_rounds = get_unit_bonus(attacker, EFT_COMBAT_ROUNDS);
+  if (max_rounds <= 0) {
+    if (attackpower == 0 || attack_firepower == 0) {
+      *att_hp = 0;
+    } else if (defensepower == 0 || defense_firepower == 0) {
+      *def_hp = 0;
+    }
+  }
```
lmoureaux, line 308: Is it possible to reach this function if `attack_firepower = defense_firepower = 0` ? I think this would trigger an infinite loop below since `def_hp` and `att_hp` would never be reduced.

ec429: Surely we'd match the first `if` clause, and set `att_hp = 0`.

lmoureaux: Oh, right, an infinite loop can only happen if `max_rounds <= 0`.

Issue #455, opened 2021-05-01
RNG improperly seeded on some platforms with buggy std::random_device


ec429:
This might just be a problem with my system (it's an AMD cpu so RDRAND doesn't work, but you'd hope the stdlib is smart enough to realise that and not use it), but the RNG seems very not random.  I did a combat with debug logging enabled, and herewith the results:
```
[debug] freeciv-server (/home/edward/devel/freeciv21/server/unithand.cpp:3876) - T133: Start attack: English Attacker against Japanese Troops.
[debug] freeciv-server (/home/edward/devel/freeciv21/server/unittools.cpp:292) - T133: attack:72, defense:60, attack firepower:1, defense firepower:1
...
[debug] freeciv-server (/home/edward/devel/freeciv21/utility/rand.cpp:53) - T133: fc_rand(132) = 29 at /home/edward/devel/freeciv21/server/unittools.cpp:308
[debug] freeciv-server (/home/edward/devel/freeciv21/utility/rand.cpp:53) - T133: fc_rand(132) = 29 at /home/edward/devel/freeciv21/server/unittools.cpp:308
[debug] freeciv-server (/home/edward/devel/freeciv21/utility/rand.cpp:53) - T133: fc_rand(132) = 29 at /home/edward/devel/freeciv21/server/unittools.cpp:308
[debug] freeciv-server (/home/edward/devel/freeciv21/utility/rand.cpp:53) - T133: fc_rand(132) = 29 at /home/edward/devel/freeciv21/server/unittools.cpp:308
[debug] freeciv-server (/home/edward/devel/freeciv21/utility/rand.cpp:53) - T133: fc_rand(132) = 29 at /home/edward/devel/freeciv21/server/unittools.cpp:308
[debug] freeciv-server (/home/edward/devel/freeciv21/utility/rand.cpp:53) - T133: fc_rand(132) = 29 at /home/edward/devel/freeciv21/server/unittools.cpp:308
```

Yes, we rolled a 132-sided die six times and got the same answer every time ;_;

ec429:
Further investigation (as discussed on Discord) determined the following:

- `generator()` is returning the same value every time
- calling `generator.discard(1)` usually, but not always, gets us a different value on the next `generator()` call
- the problem does not occur in a simple test program built with `g++ -x c++ -`:

```
#include <iostream>
#include <chrono>
#include <random>

int main(void)
{
  static std::mt19937 generator = std::mt19937();
  generator.seed(1234);
[...]

PR #475, opened 2021-06-14
Workaround for lack of random_device randomness

lmoureaux:
The RDRAND instruction is broken on certain AMD processors and always returns
-1. libstdc++ trusts it for the default behaviour of its std::random_device,
which obviously results in a very bad seed.

This patch switches to QRandomGenerator for seeding. QRandomGenerator
implements a workaround since Qt 5.12.6, disabling hardware random generation
if it looks broken.

See GCC bug #100444 at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100444
See QTBUG-69423 at https://bugreports.qt.io/browse/QTBUG-69423
See issue #455

### DIFF
--- a/utility/rand.cpp
+++ b/utility/rand.cpp
@@ -25,6 +25,9 @@
 
 #include "rand.h"
 
+// Qt
+#include <QRandomGenerator>
+
 #define log_rand log_debug
 
 /* A global random state:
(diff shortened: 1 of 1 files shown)


ec429:
I've tested this and newly-created games have random-looking RNG state on my system, where previously they had lots of repeated values.

lmoureaux:
Thanks for checking.

Issue #490, opened 2021-06-23
Add to Modpacks Manual - Tilesets


jwrober:
Sub issue/task of #488 Create a collection of sections on how to edit/create tilesets for Freeciv21

jwrober:
Might be some material from legacy docs. See #485

jwrober:
PR #682

PR #502, opened 2021-06-24
Port existing ruleset docs to Sphinx

lmoureaux:
Only `README.effects` so far.

See #490.


jwrober (approved):
This looks good. I like the sub-directory for `rulesets`, gives a good categorization.

PR #517, opened 2021-07-19
Use system Lua (>= 5.3) instead of a built-in version

### LINKED ISSUE #504: Make it possible to use system Lua for FCDB
PR #417 made it possible to use `require` to load external Lua packages from the FCDB script. There are still two problems that prevent its use:

* The loading of binary modules (required for database connections) is disabled
* Lua 5.4 is always used, but it's not always compatible with system libraries

This leads to the following error message on the Longturn server (Debian Buster):
```
[critical] freeciv21-server - lua error:        error loading module 'luasql.postgres' from file '/usr/lib/x86_64-linux-gnu/lua/5.3/luasql/postgres.so':
        dynamic libraries not enabled; check your Lua installation
stack traceback:
        [C]: in ?
        [C]: in function 'require'
        /home/dude/.config/freeciv21/database.lua:183: in function 'database_init'
```
Notice how I tried to load a library for Lua 5.3 while the built-in version is 5.3.

This is probably an opportunity to switch to a more general (and C++!) Lua abstraction layer instead of `tolua`, maybe [sol2](https://sol2.readthedocs.io/en/latest/index.html).

Solving this issue is required to start LTEx24.

lmoureaux:
The built-in Lua was not compatible with Lua modules installed using the platform's package manager. Using the system-provided Lua should improve compatibility, allowing to close #504.

**document-me**: Lua is no longer provided and Lua >=5.3 is now a hard dependency. This should improve interoperability with other Lua packages on your system.

Closes #504.


lmoureaux:
Tested, seems to work.

PR #544, opened 2021-07-22
Modify AUTHORS

pranavtifr:
Include @pranavtifr and @DobbyK


lmoureaux (changes requested):
Let's keep it order alphabetically.

lmoureaux (approved):

PR #575, opened 2021-07-29
Cmake sphinx

jwrober:
Two commits. First is to wire in sphinx-build into cmake as a custom target and the second is an update to install.rst showing new target w/ some cleanup to better support RTD theme.

### DIFF
--- a/cmake/FindSphinx.cmake
+++ b/cmake/FindSphinx.cmake
@@ -4,9 +4,9 @@ include(FindPackageHandleStandardArgs)
 if(PythonInterp_FOUND)
   get_filename_component(_PYTHON_DIR "${PYTHON_EXECUTABLE}" DIRECTORY)
     set(_PYTHON_PATHS
-        "${_PYTHON_DIR}"
-        "${_PYTHON_DIR}/bin"
-        "${_PYTHON_DIR}/Scripts")
+      "${_PYTHON_DIR}"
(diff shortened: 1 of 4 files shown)


docs/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# CMakeLists for Freeciv21 Docs
+
+find_package(Sphinx REQUIRED)
```
lmoureaux, line 3: This makes Sphinx a hard dependency, which I don't think we want (and if we do, it should be documented). Remove the `REQUIRED` and wrap the rest into `if (Sphinx_FOUND)`?

cmake/FindSphinx.cmake
```diff
@@ -0,0 +1,17 @@
+# Support for sphinx-build
+include(FindPackageHandleStandardArgs)
+
+if(PythonInterp_FOUND)
+  get_filename_component(_PYTHON_DIR "${PYTHON_EXECUTABLE}" DIRECTORY)
+    set(_PYTHON_PATHS
+        "${_PYTHON_DIR}"
+        "${_PYTHON_DIR}/bin"
+        "${_PYTHON_DIR}/Scripts")
```
lmoureaux, line 9: Inconsistent indentation.

lmoureaux (changes requested):
I don't think Sphinx should be required to build Freeciv21.

PR #585, opened 2021-08-10
Fix install from deleted old doc files

jwrober:


lmoureaux (approved):
Looks like someone should fix the CI.

PR #589, opened 2021-08-17
Feature: Make nukes destroy city improvements

pranavtifr:
longturn/freeciv@0c1b537 implemented as an effect instead of server setting.

### DIFF
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -2799,8 +2799,25 @@ static void do_nuke_tile(struct player *pplayer, struct tile *ptile)
                     _("You nuked %s."), city_link(pcity));
     }
 
+    city_built_iterate(pcity, pimprove)
+    {
+      /* only destroy regular improvements, not wonders, and not those
+       * that are especially protected from sabotage (e.g. Walls) */
+      if (is_improvement(pimprove) && pimprove->sabotage == 100
+          && fc_rand(100) < (float(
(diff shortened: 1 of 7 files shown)


server/unittools.cpp
```diff
@@ -2800,8 +2800,24 @@ static void do_nuke_tile(struct player *pplayer, struct tile *ptile)
                     _("You nuked %s."), city_link(pcity));
     }
 
+    /* destroy improvements, 50 % chance each */
```
lmoureaux, line 2803: I think the chance depends on the effect

server/unittools.cpp
```diff
@@ -2800,8 +2800,24 @@ static void do_nuke_tile(struct player *pplayer, struct tile *ptile)
...
+    city_built_iterate(pcity, pimprove) {
+      /* only destroy regular improvements, not wonders, and not those
+       * that are especially protected from sabotage (e.g. Walls) */
+      if (   is_improvement(pimprove)
```
lmoureaux, line 2807: Did `clang-format` leave the whitespace?

lmoureaux (changes requested):
Let's keep the documentation up-to-date: update `docs/Modding/Rulesets/effects.rst`

lmoureaux:
Any update on this? The suggested changes should be quite easy.

lmoureaux:
I'm thinking of this only now, but the list of effects are sent to the client so it should be checked that old-ish clients don't crash immediately...

lmoureaux (approved):

PR #608, opened 2021-08-31
Refactor: Replace `QString::SkipEmptyParts` with `Qt::SkipEmptyParts`

pranavtifr:
`QString::SkipEmptyParts` has been deprecated.
https://doc.qt.io/qt-5/qstring-obsolete.html


lmoureaux:
Unfortunately we need to support Qt 5.10 which doesn't have `Qt::SkipEmptyParts`.

Issue #617, opened 2021-09-02
JSON file of the Library to License(s)


jwrober:
**Is your feature request related to a problem? Please describe.**
As part of #434 we will get all the licenses defined for all of the third-party libraries and tools the project relies on.

**Describe the solution you'd like**
Include a JSON file that can be picked up at compile/build time to add the details to the in-game help.

**Describe alternatives you've considered**
Doing nothing comes to mind :-)

**Additional context**
None

Issue #577, opened 2021-07-30
Output Docs as Plain Text Files


jwrober:
**Is your feature request related to a problem? Please describe.**
As part of generating release or a windows installer package, include the reStructuredText in `docs` as plain text files packaged in `doc` directory. 

**Describe the solution you'd like**
As we replace all the documentation in the current source `doc` directory we are increasingly going to have double documentation that is very hard to maintain. We would maintain it once in `docs` and then package it in `doc` at release time. 

Integrate all of this in using CMake and CPack configuration

**Describe alternatives you've considered**
As mentioned above, the current process is to maintain two different versions. This is not sustainable.

**Additional context**
It has been expressed that it is good practice to allow "offline" documentation available in our packages. This feature would enable that and also allow us to only maintain one copy of the documentation.

jwrober:
From an implementation standpoint I am thinking of moving all existing documentation that currently resides in `doc` and move them to `doc/legacy`. This would free up the root of `doc` to hold the new generated files. We would also want to figure out how to NOT create the index files and also what to do with files that live in the root such as `INSTALL`

jwrober:
After much discussion Discord, we don't think this is needed. Instead made a change in PR #801 to setup RTD to instead produce a PDF of latest.

PR #628, opened 2021-09-08
Licenses

### LINKED ISSUE #434: Include licenses of the libraries and third-party tools we rely on
Generally considered good practice to include the licenses of the libraries we use in the doc directory.

jwrober:
Closes #434

Looking for feedback before commiting.


lmoureaux:
I understand correctly, this PR adds a new folder under `doc` (why not `docs` or `dist`?) with copies of the licenses and an index file. You appear to have done a great job collecting all of them.

The main point that lead to #434 is to comply to conditions like the following (from the MIT license):

> The above copyright notice and this permission notice shall be included in all copies or substantial portions
of the Software.

These clauses only apply to software that we do include. No free software license requires that we do anything if we only use a program. This means that the licenses and copyright notices for CMake, `ninja`, `xgettext` and other build tools do not need to be included.

This also applies only to platforms where we distribute dependencies ourselves, i.e. the Windows installer. When the dependencies are provided by a package manager, it's up to the packager to provide the license information. Debian, for instance, keeps very detailed structured information in `/usr/share/doc/*/copyright`. MSYS2 distributes the licenses in `/mingw64/share/licenses/`; copyright notices appear to be lost. Another system that distributes licenses is Android, reachable from the "About System" menu in the settings.

For full compliance we should keep the copyright notices. However, given that MSYS2 doesn't do it I don't feel compelled to do it... we could copy their information if they provided it. Having the licenses is already a major step forward. (We could keep the notices when part of the license files, for instance with a structure like this:
```
doc/licenses
doc/licenses/qt
doc/licenses/qt/GPLv3.txt
doc/licenses/lua
doc/licenses/lua/whatever-it-is.txt
...
doc/licenses/README.txt
```
Then each lib can have its own license file. This is mostly a copy of what MSYS2 provides and it could be copied directly from there.)

Another point is that we distribute the whole thing as GPLv3, so idk if it makes sense to include incompatible licenses like the GPLv2.

jwrober:
I used `doc` versus `docs` because I do expect us to have documents in `doc` after I complete #577. So having the licenses there seemed logical to me. I started going directory based route you provided in your comment until I noticed that there is so much duplication and moved to a single index type file to gather more details about the component. I can still break it out if we want to go that route. The single INDEX type file also sets us up for the JSON output you wanted for #617

Issue #632, opened 2021-09-10
Add a window for tileset debugging


lmoureaux:
**Is your feature request related to a problem? Please describe.**

Understanding how the map is drawn is hard, partly because there is no tool to break down the sprites used to draw an arbitrary tile. Adding a tool to dissect existing tileset would help raise awareness of the map-drawing techniques employed by the various tilesets and be helpful for modders.

**Describe the solution you'd like**

A window with a button to can pick a tile to debug. Once a tile is selected, the list of sprites used to draw it is displayed.

**Describe alternatives you've considered**

None so far.

**Additional context**

Everyone agrees that making new tilesets is hard, especially terrain.

jwrober:
Some thoughts on this - the window should be disconnected from the client so you can move it to a second monitor from the main client. The spec files and associated line data should also be displayed for quick troubleshooting. Option to be in "development mode" and edit the spec file(s) in the window would also be pretty cool if the client would be able to refresh to see the change on the fly.

lmoureaux:
Very crude implementation that might already be useful:
![image](https://user-images.githubusercontent.com/22327575/132932569-53d70767-6bbf-4fd7-ab84-d950a7f4f72f.png)

jwrober:
Per a conversation on Discord, there is still some work to do on this one in a another pull request before it is complete.

PR #450, opened 2021-04-26
Refactor tilespec.cpp

lmoureaux:
This is a major refactoring of the tileset drawing code to split it and simplify what can be. Hopefully at the end we get something much easier to work with.

The following layers have been ported:
* `LAYER_BACKGROUND`
* `LAYER_SPECIAL*`
* `LAYER_BASE_FLAGS`
* `LAYER_DARKNESS`
* `LAYER_TERRAIN*`

The following layers have not yet been ported:
* `LAYER_WATER`
* `LAYER_ROADS`
* `LAYER_FOG`
* `LAYER_UNIT`
* `LAYER_CITY*`
* `LAYER_GRID*`
* `LAYER_BASE_FLAGS` (which is new)
* `LAYER_OVERLAYS`
* `LAYER_TILELABEL`
* `LAYER_CITYBAR`
* `LAYER_FOCUS_UNIT`
* `LAYER_GOTO`
* `LAYER_WORKERTASK`
* `LAYER_EDITOR`
* `LAYER_INFRAWORK`

Features dropped:
* Hiding the terrain: #449
* Support for `is_reversed` in `[tile_*]` sections (unused in shipped tilesets, no longer needed since all layers are equivalent)

Closes #430.

PR #633, opened 2021-09-11
Initial tileset debugger

lmoureaux:
Implements #632. It's not quite perfect yet; in particular, only tile-based drawing is supported (as opposed to corner- and edge-based). Darkness is corner-based; grid and borders are edge-based.

It looks like this:

image

Built on top of #450 and helped understand a bug spotted there.


docs/Modding/Tilesets/debugger.rst
```diff
@@ -0,0 +1,48 @@
...
+================
+
+.. versionadded:: 3.0-alpha6
+
+The Tileset Debugger, accessible from the :guilabel:`Game` menu, lets you
+inspect how the map is drawn. This is very helpful when developing a tileset, to
+understand why something is rendering incorrectly or to understand how other
+tilesets work.
+
+.. note::
```
jwrober, line 11: Do we specifically need this note? Its pretty obvious we like suggestions.

lmoureaux: I wanted to emphasize this because the direction it takes in its early days will directly affect the quality of the end result (unless we rewrite it three times)

PR #635, opened 2021-09-12
Freeciv21 or bust

jwrober:
Collection of commits to attempt to clear out all vestiges of legacy freeciv in any dialogs, game messages/strings. etc.


jwrober:
I think I will commit this and then fix the clang formatting in another PR. My local branch copy is jacked up right now.

PR #670, opened 2021-09-27
Bump version to Alpha8

jwrober:
Set for Friday 10/08/2021


jwrober:
@lmoureaux - This version is set for tomorrow 10/8/21. With the doc changes and the other code fixes, seems like a good time to do another alpha release. thoughts?

PR #689, opened 2021-10-15
Prevent clicking on HUD widgets from acting on the map

### LINKED ISSUE #687: Clicking on Unit in "Unit Bar" at bottom of screen opens City window when overlapping
**Describe the bug**
If the Unit action bar at the bottom of the screen overlaps a city on the map and you click on unit to center it on the map, the city window will pop up instead.

**To Reproduce**
1. Notice a unit is active and shown in the unit action bar at the bottom of the screen (see screen shot) and also notice that the unit bar overlaps a bit of a city.
2. Click on the in the unit action bar to attempt to center the unit on the screen. 
3. The screen will attempt to center on the unit and then will instead open the city window.

**Expected behavior**
Because the city is "behind" the unit action bar, I would expect to have the unit centered on the screen and ignore the city.

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.0-alpha8+
 - Ruleset/Longturn game (if applicable): Classic

**Additional context**
None

lmoureaux:
Clicking on a widget displayed on top of the map was sometimes triggering
actions as if the map had been clicked underneath. Prevent this by disabling
the propagation of mouse events to the parent widget.

Closes #687.

### DIFF
--- a/client/gui-qt/citydlg.cpp
+++ b/client/gui-qt/citydlg.cpp
@@ -1321,6 +1321,11 @@ city_dialog::city_dialog(QWidget *parent) : QWidget(parent)
   small_font = fcFont::instance()->getFont(fonts::notify_label);
   ui.setupUi(this);
 
+  // Prevent mouse events from going through the panels to the main map
+  for (auto child : findChildren<QWidget *>()) {
+    child->setAttribute(Qt::WA_NoMousePropagation);
+  }
+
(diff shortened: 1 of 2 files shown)


lmoureaux:
Ok right, it fixes what Kevin originally reported for 566 but not what I wrote there.

Issue #686, opened 2021-10-14
Don't scale city-bar flags with Map Zoom


HansLemurson:
**Is your feature request related to a problem? Please describe.**
When using the + and - keys to shrink or enlarge the tiles of the map (very useful), the city-bar grows and shrinks with the rest of the map.  This makes the city-bars take up too much screen area and become a nuisance.
Fortunately, there's an option in the "View" menu for "scale fonts", that can be un-checked to prevent the text-size from growing.
_Unfortunately_, the city-bar enlarges anyways, because the flag graphics within it shrinks and grows with the map.
![image](https://user-images.githubusercontent.com/10428933/137282285-8b846578-a405-4893-91ce-72c2bc80265c.png)

**Describe the solution you'd like**
I would like the city-bar to remain a constant size while the map is shrunk or enlarged behind it.
Maybe have that be controlled by the "scale fonts" checkbox.  Just rename it to

**Describe alternatives you've considered**
Suffering silently.

**Additional context**
The problem seems to be related with _all_ tileset graphics being shrunk or enlarged, regardless of whether the graphics are for the UI or the Map.  Is there a way to deal with graphics used in the city-bar UI separately?  
The graphics in the city-view itself seem to be immune to the changes, so that's good.

lmoureaux:
Is this what you suggest?
![image](https://user-images.githubusercontent.com/22327575/137562750-1ffdabb6-5c83-422c-8f3b-ae504c07b04e.png)

HansLemurson:
> Is this what you suggest? ![image](https://user-images.githubusercontent.com/22327575/137562750-1ffdabb6-5c83-422c-8f3b-ae504c07b04e.png)

Beautiful!  Now my only concern is that the gigantic whale will devour us all.

PR #690, opened 2021-10-15
Make "scale fonts" play nicer with the city bar

lmoureaux:
Implements Hans' request in #686


jwrober:
I can confirm that this handles the zooming in/out issue that Hans reported.

PR #692, opened 2021-10-16
Open the help dialog with at least 1280*800px

### LINKED ISSUE #563: The Help dialog is too small by default
**Describe the bug**

The Help dialog opens by default at a very small size. It could be adaptative and occupy maybe 2/3 of the available by default.

**To Reproduce**
1. Remove `~/.config/freeciv21-client.ini`
2. Open the client
3. Start a game
4. Open Help > Terrain
5. Navigate to Hills
6. Watch the content as reproduced above

**Expected behavior**

The help dialog has a sensible default size.

**Screenshots**
image

 - OS: Kde Neon user
 - Freeciv21 version: `master`
 - Ruleset/Longturn game (if applicable): most, `classic` or `LTEx24` will do

**Additional context**

The default size was chosen when 1024*768 resolution was still a thing and the Freeciv guidelines even recommended testing on 768*480.

lmoureaux:
This will enhances its layout on very busy pages (e.g. terrain and units).

Closes #563.

### DIFF
--- a/client/gui-qt/helpdlg.cpp
+++ b/client/gui-qt/helpdlg.cpp
@@ -193,7 +193,8 @@ void help_dialog::showEvent(QShowEvent *event)
     QList<QScreen *> screens = QGuiApplication::screens();
     QRect rect = screens[0]->availableGeometry();
 
-    resize((rect.width() * 3) / 5, (rect.height() * 3) / 6);
+    resize(qMax((rect.width() * 3) / 4, 1280),
+           qMax((rect.height() * 3) / 4, 800));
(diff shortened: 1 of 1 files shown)


jwrober:
Should we override a display that is smaller than the default?

lmoureaux:
Do you mean screens smaller than 1280*800?

jwrober:
No, I was thinking more along the lines of my testing. If the client finds that the help dialog is set to less that 1200x800 and the overall screen geometry allows for it, that the client will override the setting and go back to 1200x800 as a default minimum size. If I set it to something bigger than that, the client ignores the change and does not do anything else.

Issue #660, opened 2021-09-23
A 3.0 folder is created inside the 3.0 folder


lmoureaux:
**Describe the bug**
```
$ ls ~/.local/share/freeciv21/3.0/
3.0  aviation  aviation.serv  aviation.tilespec  scenarios
```
The 3.0 folder should not be there. It's possibly created by the modpack installer?

**To Reproduce**
1. See above

**Expected behavior**

See above

**Platform and version (please complete the following information):**
 - OS: Windows (10?), Linux
 - Freeciv21 version: 3.0-alpha6
 - Ruleset/Longturn game (if applicable): None

PR #694, opened 2021-10-16
Don't create a double 3.0 folder

lmoureaux:
See #660.

As a side-effect of this PR, the modpack installer will "forget" which modpack versions are installed. Installed modpack will remain usable; installing them again merely overwrites the files.


jwrober:
I like the documentation idea as part of the release notes. I also agree about writing migration code if we were in a more stable release cycle. If a person decides to ignore the instructions, will the modpack installer create a new `.control` file when an item is downloaded?

PR #696, opened 2021-10-22
Switch to a LT-controlled metaserver

lmoureaux:
This will implement the move to `https://longturn.net/meta/` as the default metaserver, using a brand new JSON-based protocol.

For now only the documentation of the protocol is implemented. The server part is also ready (for the `www` repo where we don't use PRs). I'd like to have comments on the API (and its docs) already now.


lmoureaux:
Not going to finish this anytime soon, closing so as not to clutter the list of open PRs.

Issue #388, opened 2021-03-29
New Coverity warnings


lmoureaux:
Please find the latest report on new defect(s) introduced to longturn/freeciv21 found with Coverity Scan.

2 new defect(s) introduced to longturn/freeciv21 found with Coverity Scan.
2 defect(s), reported by Coverity Scan earlier, were marked fixed in the recent build analyzed by Coverity Scan.

New defect(s) Reported-by: Coverity Scan
Showing 2 of 2 defect(s) 
**Both fixed :heavy_check_mark:**

```
*** CID 1451498:  Integer handling issues  (SIGN_EXTENSION)
/server/citytools.cpp: 2026 in unit_conquer_city(unit *, city *)()
2020           (void) civil_war(cplayer);
2021         }
2022         return true;
2023       }
2024     
2025       coins = cplayer->economic.gold;
>>> CID 1451498:  Integer handling issues  (SIGN_EXTENSION)
>>> Suspicious implicit sign extension: "city_size_get(pcity)" with type "citizens" (8 bits, unsigned) is promoted in "coins * city_size_get(pcity) / 200" to type "int" (32 bits, signed), then sign-extended to type "unsigned long" (64 bits, unsigned).  If "coins * city_size_get(pcity) / 200" is greater than 0x7FFFFFFF, the upper bits of the result will all be 1.
2026       coins = MIN(coins, fc_rand((coins / 20) + 1)
2027                              + (coins * (city_size_get(pcity))) / 200);
2028       pplayer->economic.gold += coins;
2029       cplayer->economic.gold -= coins;
2030       send_player_info_c(pplayer, pplayer->connections);
2031       send_player_info_c(cplayer, cplayer->connections);
```
________________________________________________________________________________________________________
```
*** CID 1436040:  Incorrect expression  (COPY_PASTE_ERROR)
/server/savegame/savemain.cpp: 253 in save_game(const char *, const char *, bool)()
247           if (!srvarg.saves_pathname.isEmpty()) {
248             QDir().mkpath(srvarg.saves_pathname);
249           }
250           tmpname = srvarg.saves_pathname;
251         } else {
252           // Make sure scenario directory exist
>>> CID 1436040:  Incorrect expression  (COPY_PASTE_ERROR)
>>> "saves_pathname" in "srvarg.saves_pathname" looks like a copy-paste error.
253           if (!srvarg.saves_pathname.isEmpty()) {
254             QDir().mkpath(srvarg.scenarios_pathname);
255           }
[...]

PR #710, opened 2021-11-13
Fix a memory leak when destroying layer objects

lmoureaux:
The absence of a virtual destructor meant that data from subclasses wasn't freed properly.

Found by Coverity.
See #388.


lmoureaux:
@pranavtifr approved on Discord

PR #726, opened 2021-11-18
Disallow drag-and-drop in the city dialog lists

### LINKED ISSUE #708: One can multiply units with drag-and-drop in the city dialog
**Describe the bug**

The city dialog present unit list has dnd enabled and this allows multiplying units…

**To Reproduce**
1. Found a city
2. Build a warrior
3. Drag-and-drop the warrior to free space (try a bit, it's not easy…)
4. You have two warriors now (ok not for real)

**Expected behavior**

You can't multiply units unless you are Jesus.

**Screenshots**

Present units: 2 -- displayed units: 6
image

 - OS: Arch Linux
 - Freeciv21 version: master

lmoureaux:
It was possible to drag-and-drop unit icons within the lists in the city dialog. This allowed one to seemingly duplicate units.

The reason for this was that QListView defaults to allowing drag and drop when in icon mode.

Closes #708.

### DIFF
--- a/client/gui-qt/citydlg.cpp
+++ b/client/gui-qt/citydlg.cpp
@@ -65,12 +65,12 @@ extern QString cut_helptext(const QString &text);
 /**
    Constructor
  */
-icon_list::icon_list(QWidget *parent) : QListWidget(parent)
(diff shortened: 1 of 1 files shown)


jwrober (approved):
I can't drag any longer with this patch.

PR #730, opened 2021-11-19
Resize sidebar buttons more sensibly

### LINKED ISSUE #556: Sidebar needs to resize more sensibly.
**Describe the bug**
If you reduce the width of the sidebar the formatting is weird. 

**To Reproduce**
1) Option > Interface
2) Reduce side bar size

**Current Behavior**
Icons gets squished, fonts remain the same size and wrap badly. 

**Expected behavior**
The icons should maintain a 1:1 aspect ratio and the fonts should be automatically reduced. The user should not be expected to tweak all the other elements of the interface to get the width right. 

**Screenshots**
Current Sidebar with reduced width. 
image

 - OS: Linux (Arch)

lmoureaux:
Always keep aspect ratios. Seems to work well on my small monitor.

Closes #556

I'd like two reviews on this one.

### DIFF
--- a/client/gui-qt/page_game.cpp
+++ b/client/gui-qt/page_game.cpp
@@ -71,33 +71,41 @@ pageGame::pageGame(QWidget *parent)
   sw_map = new sidebarWidget(
       fcIcons::instance()->getPixmap(QStringLiteral("view")),
       Q_("?noun:View"), QStringLiteral("MAP"), sidebarShowMap);
+  sw_map->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
   sw_tax = new sidebarWidget(nullptr, nullptr, QLatin1String(""),
-                             sidebarRatesWdg, SW_TAX);
-  sw_indicators = new sidebarWidget(nullptr, nullptr, QLatin1String(""),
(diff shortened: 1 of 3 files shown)


lmoureaux:
This is what I get for size 2:

image

pranavtifr (commented):
Should the font also be scaled correctly ? We dont expect the user to tweak around with sizes until it looks right,.

lmoureaux:
Is it possible to make it look good with the text at such a small size?

pranavtifr:
I think some simple constraint like.. Text shouldnt go beyond borders, and wrap it to no more than 2 lines should work (also needs to wrap sensibly along spaces).

lmoureaux:
Then when researching "Werkzeugmaschinen" you'll get a font size smaller than one pixel…

lmoureaux:
So I suggest that we merge this as the fundamental that makes things look better (the aspect ratio was never correct before) and handle the rest together with a much-needed iteration on the sidebar.

pranavtifr:
Yup. Makes sense.. Hard to write warping logic when tech names can be ruleset specific.

lmoureaux:
@jwrober's turn then?

lmoureaux:
As per discussions on Discord, no need for another review.

PR #734, opened 2021-11-21
Drop cma_apply_result()

### LINKED ISSUE #719: cma_apply_result() is never used
**Describe the bug**

Said function is never used anywhere in the code.

**To Reproduce**
1. Search the code for cma_apply_result()

**Expected behavior**

Either it's used or it's removed.

 - OS: irrelevant

**Additional context**

Many memory leaks `governors.cpp`, trying to clean up.

lmoureaux:
It was used only in the SDL client, which we dropped.

Closes #719.

### DIFF
--- a/client/governor.cpp
+++ b/client/governor.cpp
@@ -73,7 +73,6 @@ class cma_yoloswag {
 public:
   cma_yoloswag();
   ~cma_yoloswag();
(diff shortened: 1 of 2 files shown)


lmoureaux:
Reviewing this should be very simple since it only removes code: the CI already checks that things still build.

jwrober (approved):

PR #742, opened 2021-11-21
Add support for Zstandard compression

### LINKED ISSUE #713: Support zst compression when karchive supports it

Latest KArchive support ZStandard compression, which is as fast as gzip and as good as xz. Add support for this format when upstream supports it.

Detect ZStandard availability and enable it in fc21 if it's there.

* Hide it behind a configure flag: why would you disable it?
* Do nothing: and lack this great zst format?

**Additional context**

* https://bugs.kde.org/show_bug.cgi?id=434577
* https://api.kde.org/frameworks/karchive/html/classKCompressionDevice.html

lmoureaux:
Zstandard is as fast as gzip and as good as xz. This makes it a good candidate for savegame compression. Support it when KArchive is recent enough.

Leaving the default to xz for now, but we might want to revisit this in the future.

Closes #713.

Test plan (you need KArchive 5.82 or later built with zst support):

1. `set compresstype zstd`
2. Save game
3. Load game just saved

### DIFF
--- a/client/gui-qt/menu.cpp
+++ b/client/gui-qt/menu.cpp
@@ -3015,7 +3015,7 @@ void mr_menu::save_game_as()
   }
 
   str = QString(_("Save Games"))
-        + QStringLiteral(" (*.sav *.sav.bz2 *.sav.gz *.sav.xz)");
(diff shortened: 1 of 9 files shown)


jwrober (approved):
works for me

PR #748, opened 2021-11-21
Fix loading WWI scenario

### LINKED ISSUE #746: Large Scenario (Europe 1900) won't open
**Describe the bug**
I was trying to load up the Europe 1900 scenario, but every time I do so, the server starts with a "random game" instead.

**To Reproduce**
Start Scenario Game
Choose "europe_1900_ww1"
Load scenario
Server will load a random default game
Lots of server messages appear, but it's not obvious when/why the scenario failed to load.

**Expected behavior**
_The chosen scenario should actually load_ 
Or at least give a useful message about its failure.

**Screenshots**
Hard to take a picture of something that didn't happen.  It would just be a screenshot of a server ready to start a random game with a handful of easy AIs.

 - OS: Windows 10
 - Freeciv21 version: 3.0-alpha 9
 - Ruleset:  Standard?  Scenario won't load at all.

**Additional context**
Other scenarios can load, but not this one.  No idea whether it's specific to this scenario, or has a more general cause of trying to load a massive map with 30 players.

lmoureaux:
Collection of patches related to Freeciv(.org)'s new AI types "threaded" and "tex" that are not supported in fc21. Basically we always fall back to the classic AI, and only mention "classic" in our scenarios.

Closes #746.

Test plan:
* Save old Europe 1900 scenario somewhere else
* Checkout and build new code
* Load old Europe 1900
* Load new Europe 1900

### DIFF
--- a/common/ai.cpp
+++ b/common/ai.cpp
@@ -121,6 +121,7 @@ const char *ai_type_name_or_fallback(const char *orig_name)
 
   auto fb = ai_type_by_name("classic");
   if (fb != nullptr) {
+    qWarning(_("Unknown AI type %s, using \"classic\" instead"), orig_name);
     // Get pointer to persistent name of the ai_type
     return ai_name(fb);
   }
--- a/common/ai.cpp
+++ b/common/ai.cpp
@@ -119,15 +119,10 @@ const char *ai_type_name_or_fallback(const char *orig_name)
     return orig_name;
   }
 
-  if (!strcmp("threaded", orig_name)) {
-    struct ai_type *fb;
-
-    fb = ai_type_by_name("classic");
-
-    if (fb != NULL) {
-      // Get pointer to persistent name of the ai_type
-      return ai_name(fb);
-    }
+  auto fb = ai_type_by_name("classic");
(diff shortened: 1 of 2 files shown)


jwrober:
Is this problem in any of the other built-in scenarios that should be fixed now as well?

lmoureaux:
> Is this problem in any of the other built-in scenarios that should be fixed now as well?
```
$ grep -r '"tex"' data/scenarios/
$
```
No (and they'd load anyway).

common/ai.cpp
```diff
@@ -119,15 +119,10 @@ const char *ai_type_name_or_fallback(const char *orig_name)
...
-  if (!strcmp("threaded", orig_name)) {
-    struct ai_type *fb;
-
-    fb = ai_type_by_name("classic");
-
-    if (fb != NULL) {
-      // Get pointer to persistent name of the ai_type
-      return ai_name(fb);
-    }
+  auto fb = ai_type_by_name("classic");
```
pranavtifr, line 122: log a warning when doing this.

pranavtifr (changes requested):

PR #753, opened 2021-11-28
Make sprite getters return const QPixmap*

### LINKED ISSUE #745: City flags change size when opening F3
**Describe the bug**

image
->
image

**To Reproduce**
1. Use city style "simple"
1. Found a city
2. Do F3
3. Make sure the map at your city gets redrawn (for instance by scrolling away)

**Expected behavior**
City flags remain fixed in size

**Screenshots**
See above

 - OS: Arch
 - Freeciv21 version: 3.0-alpha9 (today's master)

**Additional context**
The pix is likely resized in-place when drawing the F3 list

lmoureaux:
Functions that retrieve sprites from the tileset used to return a non-const
pointer, which allowed other code to change them. This resulted in #745.
This commit changes all sprite getters to return const QPixmap* instead of
QPixmap*, solving #745 and preventing it from happening in the future.

Maybe this should've been done in smaller commits but it's too late for that.

Closes #745.

Test plan (from #745 instructions):

*    Use city style "simple"
*    Found a city
*    Do F3
*    Make sure the map at your city gets redrawn (for instance by scrolling away)
*    See error

### DIFF
--- a/client/citybar.cpp
+++ b/client/citybar.cpp
@@ -801,7 +801,7 @@ QRect polished_citybar_painter::paint(QPainter &painter,
     line.add_icon(production_progress.get());
 
     // Icon
-    QPixmap *xsprite = nullptr;
+    const QPixmap *xsprite = nullptr;
     const auto &target = pcity->production;
     if (can_see_inside && (VUT_UTYPE == target.kind)) {
       xsprite =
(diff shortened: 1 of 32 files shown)


jwrober (approved):
Works for me. The nation tile didn't change as expected.

PR #754, opened 2021-11-28
Make fcFonts more const-correct

### LINKED ISSUE #712: Changing fonts while the map is zoomed in has unexpected results
**Describe the bug**

Setting font sizes and so on while the map has a non-default zoom setting leads to unexpected jumps in size.

**To Reproduce**
1. Start a game, found a city
2. Zoom in
2. Select citybar style "Simple"
3. In the settings, change the fonts for city names
4. Repeat 3 until the font for city productions changes wildly

This is transient:

5. Zoom out
6. Watch the fonts returning to normal

**Expected behavior**

Only the font that you change actually changes.

**Screenshots**

image

 - OS: Linux
 - Freeciv21 version: 3.0-alpha8
 - Ruleset/Longturn game (if applicable): Sim30Test, confirmed with `classic`

**Additional context**

Reported by @Corbeau-1 on Discord.

lmoureaux:
In a manner similar to #745, fonts in fcFonts were being modified from the
outside. Expose them only by value so this can't happen.

Closes #712.

Test plan: combination of zoom in, zoom out, enable and disable View > Scale fonts, and change fonts.

### DIFF
--- a/client/citybar.cpp
+++ b/client/citybar.cpp
@@ -364,7 +364,7 @@ QRect simple_citybar_painter::paint(QPainter &painter,
   // First line
   if (gui_options.draw_city_names) {
     // City name
-    format.setFont(*get_font(FONT_CITY_NAME));
+    format.setFont(get_font(FONT_CITY_NAME));
     format.setForeground(*get_color(tileset, COLOR_MAPVIEW_CITYTEXT));
     first.add_text(name, format);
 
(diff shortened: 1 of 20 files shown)


jwrober (approved):
Works for me.

lmoureaux:
Rebased to fix conflicts and build failures. I guess we'd need another test: I don't think any compiler would have built the buggy code.

jwrober (approved):
I am no longer able to recreate the issue with this patch.

PR #756, opened 2021-12-03
Fix colored menus on System Theme

### LINKED ISSUE #567: Windows System Theme black menu text hard to read
**Describe the bug**
On Windows 10, the "System" theme shows menu entries as black text so you can't see anything

**To Reproduce**
1. Go to Game -> Options -> Set Local Options
2. Click on Graphics Tab
3. Select "System" from the Theme drop down box
4. Click OK
5. Click on Game menu again

**Expected behavior**
A font color you can see

**Screenshots**

2021-07-26 10_21_32-freeciv21-client

 - OS: Windows 10

### LINKED ISSUE #576: Change the way restrictinfra works
Shifting borders sometimes cause attacks to be incredibly swift where people lose entire kingdoms in a single turn. 

One prevailing idea is to have restrictinfra as an extension of ZOC, where each unit has a strength for imposing restrictinfra around it. Similarly own borders has a strength and also, cities. Final restrictinfra is an accumulation of all these strengths. 

Other ideas for the problem were changing borders to have something called noman's land. Changing borders only at tc, or  restrictinfra around newly conquered city irrespective of borders.

jwrober:
Closes #567


jwrober:
When evaluating on Linux, test with a "dark mode" system theme such as Breeze Dark on KUbuntu.

lmoureaux:
This looks like NightStalker, not native Windows…
image

lmoureaux:
Same issue when forcing the `Plastic` theme instead of the native Windows theme (`Plastic` is supposedly Qt's best maintained theme).

lmoureaux:
Clearing everything from the stylesheet makes the menu look native (but obviously creates other issues).

lmoureaux (changes requested):
image
The black color for `QLabel` breaks dark themes badly (here Breeze Dark on Linux).

jwrober:
I still ended up with the NightStalker menu look on all platforms, but it at least works now.

pranavtifr:
Maybe we should not use classic bg for the system theme. 
image

GTK does something like this, and I think it is better
image

PR #758, opened 2021-12-04
Add what Linux Distros are supported

### LINKED ISSUE #714: Document the minimum version of dependencies

Following #698, we should document which version of the major distros we support. Based on [Repology](https://repology.org/project/qt/versions), requiring Qt 5.15 would imply Debian Bullseye or later, Ubuntu 21.04 or later, Fedora 33 or later, current MinGW and Homebrew.

I think we should keep Qt 5.12 support until 22.04 LTS is released, because that's what 20.04 is stuck with.

**Do you have suggestions?**

Document minimum version requirements at https://longturn.readthedocs.io/en/latest/General/install.html. Test that none of the listed distros is affected by #698 (rather easy to do with Docker, I can handle that).

jwrober:
Closes #714

### DIFF
--- a/docs/General/install.rst
+++ b/docs/General/install.rst
@@ -9,7 +9,7 @@ configuration process is smart enough to work out whether your system is suitabl
 
 An operating system that support Qt
     Any modern operating system that supports Qt 5.12+ is required. As of this writing this is Linux, Microsoft
-    Windows\ |reg| and Apple Mac OS X\ |reg|. On Windows MSYS2 (MingW) is supported.
+    Windows\ |reg| and Apple Mac OS X\ |reg|. On Windows MSYS2 (MingW) is required.
 
     Linux Distributions:
 
@@ -21,9 +21,17 @@ An operating system that support Qt
     * KDE Neon
     * Manjaro
     * Mint 20+ or Mint Debian Edition
-    * openSUSE Leap 15.2+
+    * openSUSE 15.2+
     * Slackware
     * Ubuntu 20.04 LTS+
+
+
+.. note::
+  The above list of Linux distributions is, of course, not exhaustive. The Freeciv21 Community has simply
+  listed the mainline, well supported, distributions here. The code repository has Continuous Integration
+  enabled and all code commits pass through Ubuntu, Mac OS and Windows for testing. It is assummed that the
+  user is keeping his/her computer OS up to date. Support by the community for these distributions will be
+  better than for some of the others out there, so keep that in mind if you are not an experienced Linux user.
 
 
 A C and C++ compiler
--- a/docs/General/install.rst
+++ b/docs/General/install.rst
@@ -13,16 +13,16 @@ An operating system that support Qt
 
     Linux Distributions:
 
-    * Arch Current
+    * Arch
     * CentOS 8+
     * Debian 11+ (Bullseye)
     * Fedora 28+
     * Gentoo
-    * KDE neon LTS
-    * Manjaro Stable
-    * Mint 20+ or Mint Debian Edition Current
+    * KDE Neon
+    * Manjaro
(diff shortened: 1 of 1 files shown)


docs/General/install.rst
```diff
@@ -8,8 +8,23 @@ Freeciv21 has a number of prerequisites.  Note, that apart from the first prereq
...
 
 An operating system that support Qt
-    Any modern operating system that support Qt is required. As of this writing this is Linux, Microsoft
-    Windows\ |reg| and Apple Mac OS X\ |reg|. On Windows Msys2 is supported.
+    Any modern operating system that supports Qt 5.12+ is required. As of this writing this is Linux, Microsoft
+    Windows\ |reg| and Apple Mac OS X\ |reg|. On Windows MSYS2 (MingW) is supported.
+
+    Linux Distributions:
+
+    * Arch Current
```
pranavtifr, line 16: What does arch current mean ? Is that a term ? I use arch and afaik, people just arch to denote it. Because it is rolling release.

docs/General/install.rst
```diff
@@ -8,8 +8,23 @@ Freeciv21 has a number of prerequisites.  Note, that apart from the first prereq
...
+    * CentOS 8+
+    * Debian 11+ (Bullseye)
+    * Fedora 28+
+    * Gentoo
+    * KDE neon LTS
+    * Manjaro Stable
```
pranavtifr, line 22: Same with manjaro, I dont think the stable postfix is needed. But better than current I guess, since we wont be worrying too much about supporting the testing branch. (Though it isnt really needed, cuz manjaro's testing is older than arch's stable anyways)

docs/General/install.rst
```diff
@@ -8,8 +8,23 @@ Freeciv21 has a number of prerequisites.  Note, that apart from the first prereq
...
+    Windows\ |reg| and Apple Mac OS X\ |reg|. On Windows MSYS2 (MingW) is supported.
+
+    Linux Distributions:
+
+    * Arch Current
+    * CentOS 8+
+    * Debian 11+ (Bullseye)
+    * Fedora 28+
+    * Gentoo
+    * KDE neon LTS
```
lmoureaux, line 21: Neon (capital N): the available editions are User, Testing, Unstable, and Developer. We support User but I guess the unstable ones will also work. Let's just say "KDE Neon"?

docs/General/install.rst
jwrober: I think it means that you are keeping it up to date aka current.

docs/General/install.rst
jwrober: I can remove current from Arch and stable from Manjaro if you think that is more accurate.

docs/General/install.rst
pranavtifr: Well, isnt that assumed? All these are supported only after they keep it current. We cant and the users should not expect us to support outdated distro even if it is on the list. Maybe add a line that everything should be current and skip the current here ?

docs/General/install.rst
jwrober: ok will be fixed in next commit. I also got rid of "current" and "stable"

lmoureaux:
Maybe also mention that others distros may work, we just didn't check?

jwrober:
We are listing the mainline/mainstream distros. I think the odds that someone goes to some esoteric distro is low and if they do they should know what they are doing.

pranavtifr:
I guess adding that, it will work just not officially supported is good. Ofc, people can try in older distros if they know what they are doing. and, we can help if it is simple enough.. Like we dont want to users whom we most likely wont have problem supporting left out. I dont know how exactly to phrase this, but something like.. It should work in other linux distributions too..feel free to try in other distros and let us know if it works .. etc

lmoureaux (approved):

PR #759, opened 2021-12-05
Fix a possible null pointer dereference in sidebar.cpp

lmoureaux:
See Coverity CID 1466477
See Issue #388

Test plan: No-op change, only checks that a pointer is not null before dereferencing it. A null pointer would have lead to a crash, which I didn't see.


jwrober:
Not sure what to evaluate on this one. The code compiles and I am able to open the client and a savegame.

pranavtifr (approved):
I am not sure why Coverity even gave this error, we are reassigning `scaled_pixmap` not referring to it in any manner. Anyways, I approve as I consider this a "noise" warning from coverity.

Issue #445, opened 2021-04-21
Freeciv Server QProcess::CrashExit


jwrober:
I'm playing AI based game on pretty current code. I am getting the above error at TC. Attached is save game and debug output.

[freeciv-T0214-Y01565-auto.sav.zip](https://github.com/longturn/freeciv21/files/6354189/freeciv-T0214-Y01565-auto.sav.zip)
[freeciv21-debug.log.zip](https://github.com/longturn/freeciv21/files/6354192/freeciv21-debug.log.zip)

jwrober:
Update, if you don't do any attacking of the Pirate city or move any of the other units you can go to the next turn.

lmoureaux:
Probably unrelated, but this AI thinks Cruisers will defend his cities well:
```
T214: Iroquois Skohare(37,100) (s2) {d800 u0 g0}  m_a_c_d urgency=0 danger=100 num_def=0 our_def=0
T214: Kaintwakon::Guerilla Warfare (want 78.000000, dist 10)  + 25 for Partisan to defend Skohare
T214: Kaintwakon::Tactics (want 338.000000, dist 5)  + 88 for Alpine Troops to defend Skohare
T214: Kaintwakon::Conscription (want 228.000000, dist 2)  + 73 for Riflemen to defend Skohare
T214: Kaintwakon::Amphibious Warfare (want 793.000000, dist 6)  + 12 for Marines to defend Skohare
T214: Kaintwakon::Combined Arms (want 3.000000, dist 24)  + 1 for Paratroopers to defend Skohare
T214: Kaintwakon::Labor Union (want 197.000000, dist 21)  + 63 for Mech. Inf. to defend Skohare
T214: Kaintwakon::Steel (want 56.000000, dist 6)  + 18 for Cruiser to defend Skohare
T214: Kaintwakon::Rocketry (want 2414.750000, dist 23)  + 136 for AEGIS Cruiser to defend Skohare
T214: Iroquois Skohare(37,100) (s2) {d800 u0 g0}  m_a_c_d wants first defender with 200.000000
```

lmoureaux:
No crash for me on Linux with the latest `master`. Can you be more specific about what units I need to move (or maybe save just before TC if you can reproduce from there).

jwrober:
I compiled 80d2c15. I followed these steps and then clicked check to end turn and got a crash.

- Attack Lodz w/ Cavalry at 23,60 twice
[...]

PR #770, opened 2021-12-11
Fix integer overflow leading to division by zero

lmoureaux:
Taking the square of danger creates too large numbers in some cases. The
previous workaround - casting to long - is not guaranteed to work by the
standard and indeed it doesn't work on Windows.

This reverts commit c67fc6a2058028582ecdaca48d847bfdb4db1112. The succint
commit message there says that making danger quadratic is in line with defense.
But danger is already quadratic in assess_danger, so I guess that commit was a
mistake. In the worst case this restores the 2.6 formula, and AI didn't behave
that bad there.


lmoureaux:
Closes #445 (again)

jwrober (approved):
I am no longer able to recreate the crash on my msys2 environment with this patch.

PR #775, opened 2021-12-17
Add Emscripten Webassembly build

zekoz:


client/gui-qt/CMakeLists.txt
```diff
@@ -62,3 +62,9 @@ target_include_directories(gui-qt PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 
 target_link_libraries(gui-qt PRIVATE client)
 target_link_libraries(gui-qt PRIVATE Qt5::Widgets)
+if(APPLE)
+  target_link_libraries(gui-qt PRIVATE "-framework CoreAudio" "-framework ForceFeedback" "-framework AudioToolbox")
+  qt5_import_plugins(gui-qt 
+    INCLUDE Qt5::QCocoaIntegrationPlugin
+  )
+endif()
```
lmoureaux, line 70: Can you move this to #776 ?

patches/lua.CMakeLists.txt
```diff
@@ -0,0 +1,146 @@
+# Lua can be compiled as either C or C++.
```
lmoureaux, line 1: What's the license of this file? Any copyright notice we need to keep?

lmoureaux (changes requested):
Code looks reasonable, two simple comments

patches/lua.CMakeLists.txt
zekoz: MIT license https://github.com/microsoft/vcpkg/blob/master/ports/lua/CMakeLists.txt

lmoureaux: Then we need to comply with:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

zekoz: Done

lmoureaux: Thanks!

PR #778, opened 2021-12-21
Unhardcode sprite counts

### LINKED ISSUE #777: Allow for tileset-defined number of unit select frames and frame rates
I am attempting to create a smoother animated unit selection ring. However, the client is hard-coded to receive 4 frames (hardcoded [here](https://github.com/longturn/freeciv21/blob/9ee5b788557d7bef798e2191a705a7938dc8d26b/client/tilespec.cpp#L120)) that last 100ms each (hardcoded [here](https://github.com/longturn/freeciv21/blob/9ee5b788557d7bef798e2191a705a7938dc8d26b/client/tilespec.cpp#L5206)). 

I would like for the client to be able to interpret two new variables provided by the tileset: a variable that determines the number of frames used in the unit selection animation, and a variable that determines the number of ms each frame stays on.

This was discussed in the discord channel for project-vectron. The presented solution seems to be the most inclusive.

**Additional context**
none

lmoureaux:
This closes #777.

I unhardcoded a few extra elements that happened to be defined nearby. Feature flags were added.

### Test plan

Edit `amplio2/select.spec` as follows:
```py
tiles = { "row", "column", "tag"
  0, 0, "unit.select0"
  0, 1, "unit.select1"
  0, 2, "unit.select2"
  0, 3, "unit.select3"
  0, 3, "unit.select4"
  0, 2, "unit.select5"
  0, 1, "unit.select6"
  0, 0, "unit.select7"
}
```

Uncomment the `select_step_ms` line in `amplio2.tilespec` and set it to something unreasonable (tested 1 and 1000).

Read the new documentation.

Ideally one would also test the other sprites.

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -1136,9 +1136,7 @@ double real_timer_callback()
     }
   }
 
-  /* Make sure we wait at least 50 ms, otherwise we may not give any other
-   * code time to run. */
-  return MAX(time_until_next_call, 0.05);
+  return std::max(time_until_next_call, 0.0);
 }
 
 /**
--- a/client/tilespec.cpp
+++ b/client/tilespec.cpp
@@ -88,7 +88,9 @@
 
 #include "tilespec.h"
 
-#define TILESPEC_CAPSTR "+Freeciv-tilespec-Devel-2019-Jul-03 duplicates_ok"
+#define TILESPEC_CAPSTR                                                     \
+  "+Freeciv-tilespec-Devel-2019-Jul-03 duplicates_ok precise-hp-bars "      \
+  "unlimited-unit-select-frames unlimited-upkeep-sprites"
 /*
  * Tilespec capabilities acceptable to this program:
  *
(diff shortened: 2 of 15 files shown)


daavko:
I think you might need to modify some things in `client_main.cpp` as well for the `select_step_ms`. Right now it's limited to 50 ms per frame (20 fps) by this code:

```cpp
/**
   This function should be called at least once per second.  It does various
   updates (idle animations and timeout updates).  It returns the number of
   seconds until it should be called again.
 */
double real_timer_callback()
{

// ---------- some skipped code ----------

  if (get_num_units_in_focus() > 0) {
    double blink_time = blink_active_unit();

    time_until_next_call = MIN(time_until_next_call, blink_time) / 1000;
  }

// ---------- some skipped code ----------

  /* Make sure we wait at least 50 ms, otherwise we may not give any other
   * code time to run. */
  return MAX(time_until_next_call, 0.05);
}
```

(the `blink_active_unit` is confusing, it either blinks the unit when the select ring is missing or just advances to the next frame of the select ring animation)

Or just set 50 ms as the lower bound and leave it as it is 😁.

pranavtifr:
I was just checking if there is big performace issues for a higher framerate. It seems to work well, because of the failsafe daavko pointed out. We either need to document it, or remove the failsafe and check its impact on performance before approving this.

lmoureaux:
Removed the limitation. CPU goes 100% because the code is inefficient but Qt will limit the refresh rate to 60fps (or whatever the system wants). The comment about other code not running was obsolete (written in 2005 when the main client was using XAW).

Issue #342, opened 2021-03-13
Change city view to show Present Units all across the bottom


jwrober:
Right now you see a single unit in a vertical scroll. Seems that expanding the windows horizontally to show more units present in the site would make it easier to manage units in city, especially in late game when a city can have tons of units inside.

![image](https://user-images.githubusercontent.com/23037938/111015694-caaee480-836f-11eb-8909-fef465a787b3.png)

lmoureaux:
Actually this list should already be expanding horizontally. Maybe it's due to the theme?

jwrober:
I would think the window would size out horizontally and only get a second row when the width gets to be the same as the city dialog. Some examples of what is occurring now.

4 Units
![image](https://user-images.githubusercontent.com/23037938/111916272-159fba80-8a48-11eb-826a-f13e1463c09f.png)

3 Units
![image](https://user-images.githubusercontent.com/23037938/111916287-2819f400-8a48-11eb-96b9-18a40a8d1740.png)

2 Units
![image](https://user-images.githubusercontent.com/23037938/111916300-3bc55a80-8a48-11eb-9306-9a287d176c45.png)

lmoureaux:
Hexamplio: 2 units
![image](https://user-images.githubusercontent.com/22327575/111916602-020e4700-8a7c-11eb-9240-cf512d92523b.png)

3 units:
![image](https://user-images.githubusercontent.com/22327575/111916631-20744280-8a7c-11eb-8720-977ec0457fd9.png)

4 units:
![image](https://user-images.githubusercontent.com/22327575/111916651-31bd4f00-8a7c-11eb-8f02-b625ed29be75.png)

lmoureaux:
Maybe this is related?

![image](https://user-images.githubusercontent.com/22327575/114315295-a2e1a680-9afe-11eb-85ec-954e7fa3a4b3.png)

jwrober:
Master as of today

2 Units - not good
![image](https://user-images.githubusercontent.com/23037938/134981878-149eeb05-2ca9-4567-adcf-1b21cfe3b4f1.png)

3 units - good
![image](https://user-images.githubusercontent.com/23037938/134981963-6f34af9e-e2f5-4c00-83c4-17cbc6578b27.png)

jwrober:
I'm using Aviation Tileset and using NightStalker Theme. Here are my font settings.
[...]

PR #791, opened 2021-12-29
Solve layout issues in the city screen right panel

### LINKED ISSUE #665: City Window right-panel text gets squished
When a city has a large number of supported units, the "Supported Units" subpanel expands and seems to squish the rest of the contents of its panel.

**To Reproduce**
Home a large number of units to the same city
Open the city window
Look at squished text in the production numbers window.

**Expected behavior**
The text of the numbers for Food, production, science, etc. should maintain their normal legible spacing.

**Screenshots**
image

 - _OS:_ Windows 10
 - _Freeciv21_ version: 3.0-alpha7
 - _Ruleset/Longturn game:_ classic

**Additional context**
Maybe the scroll-bar should go left/right like it does for the improvements?

lmoureaux:
The core of this PR is a fix for #665 in the third commit:

heightForWidth() was reporting inconsistent results. Turns out it's not needed:
drop it.

Closes #665.
The symptoms described there (overlapping QLabels) sill happen when the window
is resized to a size too small for the contents, but this will be solved later
when we get rid of absolute positioning.  (This time, I hope, for good.)

Other commits are small fixes here and there that aren't worth commenting IMHO.

Test plan:
* Make a save with many units owned by the same city
* Reproduce #665
* Apply patch
* Observe that #665 is harder to reproduce

### DIFF
--- a/client/gui-qt/citydlg.cpp
+++ b/client/gui-qt/citydlg.cpp
@@ -1161,7 +1161,7 @@ void city_info::update_labels(struct city *pcity, cityIconInfoLabel *ciil)
                          + "</pre>");
     }
   }
-  // handjob
+
   ciil->updateTooltip(0, buf[FOOD + 1]);
   ciil->updateTooltip(2, buf[SHIELD + 1]);
   ciil->updateTooltip(4, buf[GOLD + 1]);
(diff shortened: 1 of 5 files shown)


lmoureaux:
This might also solve #342 (not tested)

jwrober (approved):
Resolves #665 does not #342

PR #795, opened 2021-12-31
Port Wiki Pages

### LINKED ISSUE #709: Port Github wiki pages

There's some rather fresh [documentation](https://github.com/longturn/freeciv21/wiki) in the Github wiki that I'd like to see ported to Sphinx so everything is in one place.

jwrober:
Closes #709


docs/Coding/macros.rst
```diff
@@ -0,0 +1,121 @@
+Feature Macros
```
lmoureaux, line 1: I don't think this file should be in the documentation (sorry). It was more of a WIP document while we were cleaning up the feature macros. The list is now much smaller:

```
#elif defined(ENABLE_NLS) && defined(HAVE_STRCASECOLL)
#elif defined(ENABLE_NLS) && defined(HAVE_STRICOLL)
#ifdef FREECIV_HAVE_BZ2
#ifdef FREECIV_HAVE_LIBINTL_H
#ifdef FREECIV_HAVE_LZMA
#ifdef FREECIV_HAVE_ZSTD
#ifdef HAVE_AT_QUICK_EXIT
#ifdef HAVE_CONFIG_H
#ifdef HAVE_DIRECT_H
#ifdef HAVE_MAPIMG_MAGICKWAND
#ifdef HAVE_STRING_H
#ifdef HAVE_STRINGS_H
#ifdef HAVE_SUPPRESS_APPEND
#if defined(ENABLE_NLS) && defined(HAVE__STRICOLL)
```

Some of them could still be dropped but it's nothing like what we had when we started...

It would make sense to keep the command that lists all feature macros (was quite horrible to write it), but even this isn't critical.

docs/Modding/Tilesets/nations.rst
```diff
@@ -0,0 +1,41 @@
+Adding a Nation
```
lmoureaux, line 1: This is already contained in https://longturn--795.org.readthedocs.build/en/795/Modding/Rulesets/nations.html, although I fully support having something in the tilesets section too. Maybe the modding docs should be reorganized as "How to" (with tutorials and quick guides) and "Reference" (in-depth, dry reference) but certainly not now.

lmoureaux (changes requested):
I'm afraid you did too much work on this one :(

docs/Modding/Tilesets/nations.rst
jwrober: I totally forgot that I wrote that one. I will remove one of them and then we can figure out a better way of organizing the modding section.

docs/Coding/macros.rst
jwrober: I had a feeling that that document was something of a scratch pad. I'll revert.

lmoureaux (approved):

PR #797, opened 2022-01-01
set GET_RUNTIME_DEPENDENCIES to use objdump

### LINKED ISSUE #796: CMake incorrectly uses dumpbin instead of objdump with GET_RUNTIME_DEPENDENCIES
**Describe the bug**
CMake incorrectly uses dumpbin instead of objdump with GET_RUNTIME_DEPENDENCIES. We want to use objdump, not dumpbin.

**To Reproduce**
This one is really hard to duplicate. It's related to this CMake bug - https://gitlab.kitware.com/cmake/cmake/-/issues/20854 - and was discovered by a Discord community member who wanted to contribute to project-vectron.

**Expected behavior**
A Windows package is built successfully with objdump to capture the dependencies.

**Screenshots**
N/A

 - OS: Windows - MSys2
 - Freeciv21 version: Any

**Additional context**
None

jwrober:
Closes #796

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +144,7 @@ if(WIN32 OR MSYS OR MINGW)
   #   dynamically instead of doing it manually.
   install(CODE [[
     message(STATUS "Collecting dependencies for freeciv21 executables...")
+    set(CMAKE_GET_RUNTIME_DEPENDENCIES_TOOL objdump)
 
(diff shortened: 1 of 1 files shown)


lmoureaux (approved):
This will most likely fail for environments without `objdump`, for instance when using MSVC. I don't see this as a problem: while we should aim for building and running on as many platforms as possible, building the installer is a more specialized step that can reflect our usage only.

PR #800, opened 2022-01-02
Fix bugs in the UWT panel (F2)

### LINKED ISSUE #792: Client, UWT window adjustments
Suggestions:
1. write single-digit numbers (e.g. "6 minutes") with a "0" ("06 minutes") for sorting purposes. As it is now, sorting doesn't take into account that 50 is more than 6. See image.

OR write times in time format: 01:04:23, then sorting would probably work automatically.

2. create a slide bar 
3. make it possible to move the window

Due to 2. and 3. not implemented, it's impossible to see *all* units.

image

lmoureaux:
Closes #792.

#### Test plan:

* `/set unitwaittime 9999`
* Move a bunch of units (the more the better)
* F2
* Check that all items in the list can be seen by scrolling
* Check that sorting by clicking on the headers works as expected

### DIFF
--- a/client/gui-qt/unitreport.cpp
+++ b/client/gui-qt/unitreport.cpp
@@ -50,6 +50,7 @@ units_waiting::units_waiting(QWidget *parent)
   waiting_units->setEditTriggers(QAbstractItemView::NoEditTriggers);
   waiting_units->verticalHeader()->setVisible(false);
   waiting_units->horizontalHeader()->setVisible(true);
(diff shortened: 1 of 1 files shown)


jwrober:
I can do all but the last item. What is to be clicked as a "header"?

lmoureaux:
> I can do all but the last item. What is to be clicked as a "header"?

The column names (for once, this uses a standard behavior).

jwrober (approved):
Works for me.

PR #802, opened 2022-01-03
Build a basic .deb file

jwrober:


lmoureaux:
I installed the resulting package and started the client from the menu. It started an old client installed in my home directory. The `.desktop` points to:
```
Exec=/home/louis/freeciv/freeciv21/build/install/bin/freeciv21-client
Icon=/home/louis/freeciv/freeciv21/build/install/share/freeciv21/misc/civicon.png
```
However, the `.deb` installs in `/usr`. There are more inconsistencies: for instance, translations are loaded from my home folder, which also appears in the data path.

Is this an issue on my end or a package issue?

lmoureaux:
Also, the installed size is above 400MB for my, is it because it's a debug build?

jwrober:
That is strange. Let me look more closely at the output. It should not be that big. It is a Release build and the .desktop files included point to /usr... I tested on a blank VM with a fresh OS build.

jwrober:
I wonder if `.desktop` files in home directory take precedence over the ones in `/usr`? The deb file in created with a simple build `cmake . -B build -G Ninja` and then `cmake --build build --target package`. In my local `build` directory, the meta and desktop files going to /usr/local and so does the package then uncompress ~80MB of data to `/usr/local/`.

jwrober:
I see what the issue is. The executable files are being placed in `/usr/bin`, but the desktop files are pointing to `/usr/local/bin`.

jwrober:
I found the fix - configure with this `cmake . -B build -G Ninja -DCMAKE_INSTALL_PREFIX=/usr` and then do the package with `cmake --build build --target package`. This will send the correct command to `configure_file()` and setup the package to install all components to correct spot on the file system and also configure the `.desktop` files to point to where the binaries are actually installed. I think there is an upstream defect here, but can't exactly pin it down.

lmoureaux:
OK we can do this in CI. Does the documentation need an update?

jwrober:
> OK we can do this in CI. Does the documentation need an update?

Yes, good point. I'll push a doc update to capture the steps that we can then add to CI in another PR after this one.

PR #812, opened 2022-01-09
Fix sol2 code and interoperability with tolua

### LINKED ISSUE #810: Calling sol2 functions taking tolua parameters fails silently
**Describe the bug**

For instance, the following `database.lua` never prints user names and rejects all connections:
```lua
function database_init()
end
function database_free()
end
function user_log()
end
function user_save()
end
function user_exists(conn)
    print('user_exists')
    return true
end
function user_verify(conn, password)
    print('user_verify', conn, password)
    print('username:', auth.get_username(conn))
    return true
end
function user_take(conn, conn2, player, observer)
    print("user_take", conn, conn2, player, player.name, observer)
    return false
end
```

**To Reproduce**
1. Put above code into `~/.config/freeciv21/database.lua`
2. Create the following `db.conf`:
    ```
    [fcdb]
    ```
3. Start a server with `-D db.conf -a`
3. Try to connect
4. Be rejected by the server
4. Watch terminal output

**Expected behavior**
The username is printed to the terminal and you're let in

 - OS: Linux
 - Freeciv21 version: master

**Additional context**

Setting up auth at https://freecivweb.com.

There are actually two issues:
1. Lua errors are not reported
2. sol2 is missing a `sol_lua_check` hook

Fixing this will require a somewhat big PR but everything is connected so I'll do it in one go.

lmoureaux:
Closes #810.

**Test plan:**
1. Try to reproduce #810
2. Don't reproduce #810 with the patch
3. Generate a lua error by doing something illegal, for instance `variable_that_doesnt_exist[0]`
4. Check that the error is printed to stderr

### DIFF
--- a/common/scriptcore/CMakeLists.txt
+++ b/common/scriptcore/CMakeLists.txt
@@ -47,5 +47,5 @@ target_include_directories(scriptcore PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
 
 target_link_libraries(scriptcore PUBLIC common)
 target_link_libraries(scriptcore PUBLIC tolua)
-target_link_libraries(scriptcore PRIVATE sol2)
+target_link_libraries(scriptcore PUBLIC sol2)
 target_link_libraries(scriptcore PRIVATE Qt5::Core)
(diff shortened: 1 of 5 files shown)


jwrober:
I think it best if @zekoz evaluates this one.

PR #813, opened 2022-01-09
Fix Help > Effects when the first req is negated

### LINKED ISSUE #811: Missing negations in Help>Effects output
**Describe the bug**
When a reqs vector begins with a `present=FALSE` row, the Help>Effects output doesn't correctly negate the item corresponding to that row.

**To Reproduce**
With ruleset built from https://github.com/ec429/fc-aviation/commit/6d5f6fe3ff0171a8bf2fc0c71540fda5d4b6cd56
which has the following in `effects.ruleset`:
```
[effect_trade_routes_airstrip]
type    = "Max_Trade_Routes"
value   = 1
reqs	=
    { "type", "name", "range", "present"
      "Building", "Mooring Mast", "City", FALSE
      "Building", "Commercial Airstrip", "City", TRUE
    }
```
Go to Help>Effects>Max_Trade_Routes.  The generated help will include the row
`* 1 route with Mooring Mast+Commercial Airstrip`

**Expected behavior**
Generated help includes the row
`* 1 route with not Mooring Mast+Commercial Airstrip`

 - OS: Linux (Debian)
 - Freeciv21 version: 3.0.20211112.9-alpha, gui-qt client (Qt 5.15.1), built from `a909454cc6`
 - Ruleset/Longturn game (if applicable): Aviation

**Additional context**
Swapping the order of the reqs in the vector produces the output
`* 1 route with Commercial Airstrip+not Mooring Mast`
which is also correct, showing that the problem is specifically with negations in the first row of the reqs.

lmoureaux:
The "not X" would only show for negated effects after the first.

Got confirmation from soundnfury that this worked.

Closes #811.

### DIFF
--- a/common/effects.cpp
+++ b/common/effects.cpp
@@ -1022,11 +1022,10 @@ void get_effect_req_text(const struct effect *peffect, char *buf,
   requirement_vector_iterate(&peffect->reqs, preq)
   {
     if (buf[0] != '\0') {
-      if (preq->present) {
-        fc_strlcat(buf, Q_("?req-list-separator:+"), buf_len);
-      } else {
-        fc_strlcat(buf, Q_("?req-list-separator:+not "), buf_len);
-      }
+      fc_strlcat(buf, Q_("?req-list-separator:+"), buf_len);
+    }
+    if (!preq->present) {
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Works for me

PR #814, opened 2022-01-09
Add buildings ruleset file for Royale and double wonder buy cost

zekoz:


jwrober:
Looks good to me. I like the change to `.gitignore`.

PR #815, opened 2022-01-10
Improve tileset error handling

### LINKED ISSUE #806: The "tile is not in the image" warning is generated billions of times
**Describe the bug**
When a tileset is badly written and sprite coordinates are go outside of the source image, the client will emit a warning (and fail to load the tileset). The warning comes as a modal popup window. When several sprites are out of the image, several windows are opened. This can go as far as freezing X, effectively making tilesets a DDOS attack vector.

Note: the actual message is `Sprite '%s' in file \"%s\" isn't within the image!`.

**To Reproduce**
1. Modify a `.spec` file to include coordinates like 156, 468689
3. Try to load a tileset using that `.spec` file
4. See one warning for each set of dumb coordinates

**Expected behavior**
Quick&dirty fix: show only the first error
Better fix: emit a warning, load anyway and include this info in the tileset debugger

**Screenshots**
I can't take a screenshot of a frozen X session :D

 - OS: Linux
 - Freeciv21 version: any

**Additional context**
`project-vectron` development.

lmoureaux:
This PR makes architectural changes to the way tileset errors are handled to avoid situations like #806. Errors and warnings are now saved in memory and displayed in the tileset debugger. When fixable errors are detected, a single dialog box is opened instead of one per error previously. The only recoverable issue is missing sprites for the terrain layers.

Here is a screenshot in an error situation (notice black tiles underneath the dialog):
image

The tileset dialog has gained a new tab with the message log:
image

Icons are from the theme. They will only appear on Linux and I'm not even sure every theme will have them.

I still need to update its documentation, but maybe I can get help on this one?

Closes #806.

**Testing plan:**

You will first want to reproduce #806.
:warning: This may freeze your X server! If this happens, go to a new VT with Ctrl+Alt+F2 or similar, log in and do `killall -9 freeciv21-client`. Make sure that you know how to do it before you try! :warning: 

1. Open `data/amplio2/terrain2.spec` and comment out all sprite definitions.
2. Start a game using `amplio2`
3. If there's no freeze, you'll see 64 message boxes

Now you can try the same with the patch. Relevant error messages should appear in the debugger. You can also try to change `gfx = "amplio2/terrain2"` to a non-existing name; since the file contains only terrain sprites, this case is also supported as a non-fatal tileset error.

Finally, there's a whole class of edge cases with zoom, since this basically reloads the tileset. So you should also try that.


jwrober (approved):
Works for me, we will need an update to the docs.

lmoureaux:
Docs added

Issue #593, opened 2021-08-25
Remove hardcoded automatic unit healing


Corbeau-1:
Two reasons: it is unnecessary and it is limiting.

Unnecessary: behaviour can be mimicked exclusively using ruleset effects.

Limiting: You can't adjust the healing rate in a sensible way.

(A "quick fix" was suggested to create effect with negative healing; however, this would cause a unit in a city with, for example, 9 (out of 10) HP to begin the next turn with ***less** HP than it started, and maybe even reduce health of 100% healthy units to 66%)

**_removed the other thing_**

lmoureaux:
What is automatic unit healing?

Corbeau-1:
> 
> 
> What is automatic unit healing?

In a city, a unit that doesn't move for the whole turn  unconditionally regains 1/3 of its HP. In the field, it regains 1/10.

Corbeau-1:
**Low-priority non-crucial addition**: Add another effect. `HP_Regen` heals a **percentage** of unit's HP. It would be neat if there was another effect that adds a **number** of HP.

lmoureaux:
> **Low-priority non-crucial addition**: Add another effect type. `HP_Regen` heals a **percentage** of unit's HP. It would be neat if there was another effect that adds a **number** of HP.

This effect exists, it is called `HP_Regen`.

With `HP_Regen` and `Unit_Recover`, there is enough flexibility to move the hard-coded rules for HP recovery to the ruleset.

lmoureaux:
The current code for automatic healing works as follows:

* `EFT_HP_REGEN` is evaluated, with a minimum of 30% for units in cities (rounded down)
* Units without `hp_loss_pct` get an additional 10%, rounded up
* Fortified units get an additional 10%, rounded up

So every unit always gets at least 10% back each turn, which translates into:
```ini
[effect_hp_regen_base]
type = "HP_Regen"
value = 10
reqs = {}

[effect_unit_recover_base]
type = "Unit_Recover"
value = 1
reqs = {
  ; Unit type HP not a multiple of 10
  ; This is important for units with less than 10 HP, otherwise they'd not get anything
}
```

This should be duplicated for fortified units.

For units in cities, one has to look at every `HP_Regen` effect and implement the `max()` manually. This makes an automatic migration of older rulesets hard since one would need to implement the `max()` in ruleset semantics. Imagine if there are 10 `HP_Regen` effects with partially overlapping requirements...

Corbeau-1:
[...]

PR #816, opened 2022-01-10
Unhardcode unit healing

ec429:
Replace the hardcoded extra healing in `hp_gain_coord()` with a new bonus `EFT_HP_REGEN_MIN`, which supplies the minimum (as a percentage of unittype's HP) to clamp to.
This PR currently includes updates to the Classic ruleset to replicate the old hardcoded behaviour, but does not (yet) update the other rulesets, nor touch ruleup.
Testing: build only so far.

Closes longturn#593.

### DIFF
--- a/ai/default/daieffects.cpp
+++ b/ai/default/daieffects.cpp
@@ -425,6 +425,7 @@ adv_want dai_effect_value(struct player *pplayer, struct government *gov,
     v += unit_list_size(pcity->tile->units) * 2;
     break;
   case EFT_HP_REGEN:
+  case EFT_HP_REGEN_MIN:
     num = num_affected_units(peffect, adv);
     v += (5 * c + num);
     break;
--- a/common/effects.cpp
+++ b/common/effects.cpp
@@ -1079,6 +1079,7 @@ QString effect_type_unit_text(effect_type type, int value)
   case EFT_SPY_RESISTANT:
   case EFT_VETERAN_COMBAT:
   case EFT_HP_REGEN:
+  case EFT_HP_REGEN_MIN:
   case EFT_TRADEROUTE_PCT:
   case EFT_DEFEND_BONUS:
   case EFT_CIVIL_WAR_CHANCE:
--- a/common/effects.h
+++ b/common/effects.h
@@ -301,6 +301,8 @@
 #define SPECENUM_VALUE125NAME "Illegal_Action_HP_Cost"
 #define SPECENUM_VALUE128 EFT_NUKE_IMPROVEMENT_PCT
 #define SPECENUM_VALUE128NAME "Nuke_Improvement_Pct"
+#define SPECENUM_VALUE129 EFT_HP_REGEN_MIN
+#define SPECENUM_VALUE129NAME "HP_Regen_Min"
 // keep this last
 #define SPECENUM_COUNT EFT_COUNT
 #include "specenum_gen.h"
--- a/data/alien/effects.ruleset
+++ b/data/alien/effects.ruleset
@@ -13,7 +13,7 @@
 
 [datafile]
 description="Alien World effects data for Freeciv"
-options="+Freeciv-ruleset-Devel-2017.Jan.02"
+options="+Freeciv-ruleset-Devel-2017.Jan.02 +HP_Regen_Min"
 format_version=20
 
 ; /* <-- avoid gettext warnings
--- a/data/alien/effects.ruleset
+++ b/data/alien/effects.ruleset
@@ -594,6 +594,45 @@ reqs	=
 type	= "Pollu_Pop_Pct"
 value	= -100
 
+; Basic HP regen.
+; All unit-types' hitpoints are divisible by 10, so no need to worry about rounding
+[effect_hp_regen_base]
+type    = "HP_Regen"
+value   = 10
+reqs    = {}
+
+; When hardcoded, effect_hp_regen_base was applied after the HP_Regen_Min
+[effect_hp_regen_min_base]
+type    = "HP_Regen_Min"
+value   = 10
+reqs    = {}
+
+; Units in cities regenerate at least 1/3 of their HP regardless
+[effect_hp_regen_min_city]
+type    = "HP_Regen_Min"
+value   = 33
+reqs    =
+    { "type", "name", "range"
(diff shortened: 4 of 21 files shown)


lmoureaux:
Looking at this more, it looks like the `max()` can already be implemented in `effects.ruleset`. It requires less C++ but maybe a lot more ruleset code?

Code itself looks good. Nitpicking that you could use `std::max()` instead of `MAX()` but `MIN()` and `MAX()` are used everywhere anyway, I'll probably do a mass replace sometime.

Use the following to your description to trigger [Github automation rules](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword): `Closes #593`.

Next on the todo list from my point of view:

* [ ] Add the new effect to `docs/Modding/Rulesets/effects.rst`
* [ ] Decide on a strategy for ruleset compatibility. Since 1) we're still in alpha, 2) SIM30 uses ruleup, and 3) I trust Avi will be updated in a timely manner, I think a breaking change is acceptable (@jwrober thoughts?). This would be done by updating shipped rulesets and bump the ruleset capstring. We'll want to update the modpack installer content immediately after the release.
  The alternative is to call this a new format version and enable automatic `ruleup` in the server. I'm not sure why it's currently disabled, any background info on that would be welcome.
* [ ] Ruleup uses [this god function](https://github.com/longturn/freeciv21/blob/master/server/rscompat.cpp#L652)

jwrober:
The classic ruleset seems to be updated. So if we don't update any of the other ones that ship with the two packages they would break with this change? I'm not sure I am a fan of that. Looks like it is a bit of CnP into the other ruleset files to bring them up to speed and not break existing rules on our way to Beta 1.

lmoureaux:
> The classic ruleset seems to be updated. So if we don't update any of the other ones that ship with the two packages they would break with this change?

It is my understanding that this will be done once the rest is decided upon, this PR is still a draft. My concern is more about compatibility with other rulesets (SIM and Avi).

ec429:
@lmoureaux is right, I plan to update the rest of the bundled rulesets for the 'final' PR.  I just wanted to get something up as an RfC so folks could start reviewing the code side.

Regarding ruleup, how about adding in compat effects if there is no `EFT_HP_REGEN_MIN` defined in the ruleset?  Rulesets wanting to suppress that but still not have the `MIN` could have an effect with value 0.  It wouldn't be perfect, as there's no way I can see to automatically test for `hp_loss_pct!=0`, but it'd mean un-fixed rulesets were _mostly_ compatible and wouldn't need a capstring break / flag day.

Regarding Avi, I plan to use this to make city HP regen depend on having a suitable building in the city — jets will require the same Air Base that's needed to build them, while for heavy props (e.g. Lancasters) I'll have a new Concrete Runway impr (that won't be required for build, since upgrade would bypass that).

ec429:
It occurs to me that the `Unit_Recover` we use to simulate rounding-up behaviour isn't quite a perfect match, because it applies even if unit moved this turn.  Possible alternatives:

- Lots of extra HP_Regen effects conditioned on unit flags, that add just the right amount of extra percents to make each unit DTRT.  (This is ugly.)
- Change `hp_gain_coord()` to round up when applying the regen: `hp = (base - 1) * bonus / 100 + 1;`.  This is technically a semantic change, would e.g. make 10-hp units in Fortresses (regen=25%) gain 3HP/turn instead of 2.  (But I think that's acceptable.)

lmoureaux:
I think that off-by-one differences is an acceptable tradeoff they don't change 1 -> 0 or 0 -> 1.

> Regarding ruleup, how about adding in compat effects if there is no EFT_HP_REGEN_MIN defined in the ruleset? Rulesets wanting to suppress that but still not have the MIN could have an effect with value 0. It wouldn't be perfect, as there's no way I can see to automatically test for hp_loss_pct!=0 (edit: or hitpoints not divisible by 10), but it'd mean un-fixed rulesets were mostly compatible and wouldn't need a capstring break / flag day.

This solution looks overly complicated and somewhat counterintuitive for ruleset authors. Testing for `hp_loss_pct != 0` (resp. `HP % 10 != 0`) is definitely possible by iterating over unit classes (resp. unit types). I would suggest to using the proper migration feature that one of the Freeciv devs grafted us with years ago and that was never really used: capabilities. Add a new optional capability to the ruleset capstring, and automigrate rulesets that don't define it. We can then make it implicit for ruleset version 30 when it comes. Getting this to work will likely require small changes all over the place, I'm happy to help.

I also noticed that the path finding code has some support for units losing HP (helicopters in classic rulesets). I don't think it's worth adding support for the new effect there (it is broken in other ways regarding effects: for instance, in LTT units will happily end their turn on mountains because it's safer, but all calculations are off because the code doesn't know that it also means losing one MP for the next turn).

ec429:
I've implemented the "round up" behaviour, and changed the ruleset patches accordingly.
I'll leave automigration for someone else to do, because while capabilities sound like the right thing, it also sounds like they've likely bitrotted and I don't feel like wrestling that.
I've also still not properly tested this (other than messing about in a SP Avi game bombing the AI and watching units heal or not).  I'm not super sure how to rigorously test it.

Regarding hp_loss_pct, I almost feel like that should be replaced with an HP_Loss effect (which could then be given reqs like "unitclass heli; citytile center !present; extra airbase !present" to replicate the old behaviour).  But that's just a suggestion for the future if someone wants to do something more detailed (because unhardcoding is a Good Thing).

lmoureaux:
> I also noticed that the path finding code has some support for units losing HP (helicopters in classic rulesets)…

-> #824 will pick it up automatically, but the AI won't be aware until it's ported away from the old PF

lmoureaux:
Added the migration code, I think you missed `CityCenter` in the requirements for `HP_Regen_Min`.

lmoureaux (approved):
Elsewhere:
> technically 4 and 5 don't need the CityTile Center req, because if it's not a city then `EFT_HP_REGEN` and `EFT_HP_REGEN_MIN` will be the same (e.g. 20 if fortified and not `hp_loss_pct`).  Which is why I didn't add them to the existing rulesets

Sounds good :+1:

jwrober (approved):

PR #822, opened 2022-01-15
Allow @ in filenames and valid UTF letters in usernames

zekoz:
Testing plan:
* Try to use fancy usernames with any characters. Letters and numbers should be allowed, punctuation etc should not. Things like chat should still behave as expected.
* Try to copy a ruleset as `ruleset@abcd` and load it. Should work.

### DIFF
--- a/common/player.cpp
+++ b/common/player.cpp
@@ -16,6 +16,8 @@
 #endif
 
 #include <QBitArray>
+#include <QString>
+
 // utility
 #include "fcintl.h"
 #include "log.h"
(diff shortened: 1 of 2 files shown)


lmoureaux:
What's going on with the CI?

common/player.cpp
```diff
@@ -1796,8 +1797,30 @@ int player_in_territory(const struct player *pplayer,
  */
 bool is_valid_username(const char *name)
 {
-  return (strlen(name) > 0 && !QChar::isDigit(name[0]) && is_ascii_name(name)
-          && fc_strcasecmp(name, ANON_USER_NAME) != 0);
+  QString username(name);
+
+  if (username.length() == 0) {
```
lmoureaux, line 1802: `isEmpty()`

utility/support.cpp
```diff
@@ -501,6 +501,33 @@ size_t fc_strlcpy(char *dest, const char *src, size_t n)
   }
 }
 
+size_t fc_strlcpy(char *dest, const QStringView &src, size_t n)
```
lmoureaux, line 504: It would make sense to call this function from the `const char *` overload above

common/player.h
```diff
@@ -28,7 +28,7 @@
 /* If changing this, be sure to adjust loading of pre-2.6 savegames
  * which depend on saved username to tell if user (or ranked_user) is
  * anonymous. */
-#define ANON_USER_NAME N_("Unassigned")
+const QString kAnonUserName("Unassigned");
```
lmoureaux, line 31: Use `QStringLiteral` to avoid an allocation before even `main` is called.

lmoureaux (commented):
I like the conversion of a preprocessor macro to a `QString` and the overloading of `fc_strlcpy`. Few code nitpicks.

I think the errors spotted in the CI are easy to fix.

lmoureaux:
Made the patch simpler to avoid `QString` <-> `char *` issues. Should be good to go now.

jwrober (approved):

Issue #618, opened 2021-09-04
Goto code cannot generate "Load" and "Unload" orders


lmoureaux:
**Describe the bug**
"Load" and "unload" (on the coast) are now orders different from classic goto but the path finding code wasn't updated to generate the corresponding orders.

**To Reproduce**
Try to use goto ("G") to unload some unit from a transport.

**Expected behavior**
The unit is issued the appropriate order to load or unload. The calculation of MP cost is correct (idk if this is currently the case).

**Platform and version (please complete the following information):**
 - OS: Linux
 - Freeciv21 version: Observed around 3.0-alpha3, no changes to goto code since
 - Ruleset/Longturn game (if applicable): `LTEx24`

**Additional context**
Sveinung or cazfi may have fixed this in their version.

lmoureaux:
It works when moving to an adjacent tile, but not when moving farther away.

lmoureaux:
Fixed in #824.

PR #824, opened 2022-01-16
New path finding implementation

lmoureaux:
This PR overhauls the path finding code used by the client (leaving the AI untouched). The goal is to address the shortcomings of the previous implementation, for instance #618 and bugs with fueled units. With this PR is complete, fueled units with HP loss and `damage_slows` should work fine, attacking non-native units shouldn't require the arrow keys and various load/unload scenarios should be solved. Even the special abilities of paratroopers should be taken into account.

Since this is quite ambitious and goto is a critical feature, I attach [a set of saves](https://github.com/longturn/freeciv21/files/7966619/saves.zip) I used while testing. It's also a good idea to use an old save from an advanced game of yours.

## Major missing features

The following are missing to reach feature parity:

* [x] Final orders other than load are not supported (attack, bribe, conquer city, etc.)
* [x] Patrol (`Q`) is not supported
* [ ] "Connect with" commands are not supported
* [x] Prevent sending fueled unit to death

## Known issues

* ~~The displayed end-of-turn locations can be off by one, especially the first one is often missed~~ fixed :heavy_check_mark: 
* The displayed path looks bad when there are actions jumping tiles (paradrop)
* ~~Found a case where a wrong path is generated~~ fixed (was an issue with unload orders) :heavy_check_mark: 

### Compatibility

Tilesets not using the standard `misc/colors.spec` will need to add
```
mapview_unsafe_goto = {"r", "g", "b"
    255,   0,   0
}
```
No migration path is added because I don't know of any tileset tweaking the colors.

## Performance

There is no visible lag for everyday life situations.

Some preliminary performance testing: with a debug build (no optimizations), it takes about 6s on my laptop to generate a path 175 tiles long with the whole map visible and a couple AI cities that could be attacked. Looking at the profile, I expect a significant speedup with optimizations. It's also a rather extreme case: the map has 100k tiles and is fully revealed. It already takes a minute to generate such a map. Memory usage remains invisible for a non-fueled Destroyer (and would scale linearly with fuel).

### Basic path finding

Nearly every unit falls in this category. Finding paths with roads, rails, terrain in the normal rulesets.

Status: :heavy_check_mark: works

### Multiple units

Same as above but multiple units selected.

Status: :heavy_check_mark: works

### Royale mountains

In Royale, Explorers ending their turn on a mountain lose 3 moves for the next turn. The old goto will happily end its turn on a mountain and turn predictions will be off afterwards. The new one should disfavor ending turn on a mountain, and predictions should still be correct when it's the only solution.

Status: :question: untested

### Fuel

Fueled units should refuel as appropriate when following a generated path. Moving to targets from where the unit couldn't possibly come back should be allowed but also warned about.

Status:
* Basic fueled units (`classic` Fighters, `Aviation` planes) work :heavy_check_mark:
* They'll also refuel at air bases :heavy_check_mark: 
* When the unit can't come back to a refuel point, the path turns to red: :heavy_check_mark:
  image

### HP loss

Similar to fuel, but for HP losing units such as `classic` helicopters. HP+fuel units should also work (though no existing ruleset has them and I bet it will become quite slow).

Status:
* `killunhomed` units are limited in range :heavy_check_mark:
* I'm confident that Helicopters will regenerate HP as needed, but only performed a synthetic test :question: 
* Diminishing MP across the path should be taken into account but currently untested :question: 
* Fueled + HP losing units untested :question: 

### Load and unload

Goto to and from a transport should work, even if not the last item in a path. Action enablers and `Action_Success_Actor_Move_Cost` effects should be taken into account. When needed, units should automatically refuel in transports.

image

image  image

Status:
* Tested single load/unload to and from own units in both `classic` (Transport) and `aviation` (several moves involving Paratroops and Twenties Transport), works :heavy_check_mark: 
  * This involves both action enablers and `Action_Success_Actor_Move_Cost`
  * `ACTION_TRANSPORT_ALIGHT` and `ACTION_TRANSPORT_BOARD` were never generated in my tests
* Transport chains: untested :question:
* Anything involving allied units: untested :question: 
* Refuel: untested :question: 

### Waypoints

Waypoints are added by hitting `G` again and removed by hitting `Esc`. They force the path to go through the tile under the cursor. A small flag is displayed on the map.

Status:
* Tested adding and removing waypoints for a fueled unit. Path finding and refueling seem to work as expected :heavy_check_mark: 
* Minor bug: removing a waypoint doesn't update the path immediately. :x:
* Markers appear at the correct location :heavy_check_mark: 

### Patrol

Finding paths for patrols quickly becomes slow, but works. The implementation is quite dumb and there are obvious paths for optimization -- I just don't consider it essential enough for this PR.

Status:
* Can find path for patrols and send them to the server. The unit will follow them. :heavy_check_mark: 
* It's slow, especially for units that can move to many tiles like ships, because it restarts the search from scratch every time the target tile changes (and possibly more often) :snail: 

### Paratroopers

Units that can paradrop will do so when it makes them reach their destination faster. They can even do it several times.

image

Status:
* Path finding fully functional :heavy_check_mark: 
* UI not optimal: the mouse could use the paradrop pointer when paradrop is the last action (instead of the normal goto pointer) and the path looks strange :arrow_right: later 

### Final orders

When sending a unit to a city or enemy unit stack, it should generate a special order that will make it ask for more information once it gets there (for instance, attack or poison city). To save CPU cycles, this is generated only on tiles with a visible city or unit stack at the time when the tile is visited (if e.g. an allied unit moves to uncover a stack, one needs to restart goto from scratch).

Status:
* Works for units that can attack :heavy_check_mark: 
* Untested for spies and caravan actions :question:


lmoureaux:
There's a conflict with #816, #816 should go in first.

lmoureaux:
I recall now that paths across the unknown are not allowed in the new implementation. I think this feature would be good to restore.

jwrober (approved):
I was able to test the 3 scenarios in my local test game.

PR #837, opened 2022-01-19
Add server ability to execute lua scripts on the client

zekoz:


server/scripting/script_server.cpp
```diff
@@ -275,6 +280,29 @@ static void script_server_code_save(struct section_file *file)
   }
 }
 
+/**
+   Dump the bytecode of function passed and send to client to execute.
```
lmoureaux, line 284: Lua bytecode isn't portable between LuaJIT and PUC Lua and probably also between Lua versions, so this would impose tight constraints on the versions used. I would prefer using the source instead.

common/networking/packets.h
```diff
@@ -36,6 +36,8 @@ struct data_in;
 // Used in network protocol.
 #define MAX_LEN_MSG 1536
 #define MAX_LEN_ROUTE 2000 /* MAX_LEN_PACKET / 2 - header */
+const int MAX_LEN_LUA_SCRIPT = 1
+                               << 20; /* 1 MB should be enough for anyone. */
```
lmoureaux, line 40: Does 1MB fit into a packet? We can use at most `MAX_LEN_PACKET - header - 4`.

server/scripting/script_server.cpp
```diff
@@ -275,6 +280,29 @@ static void script_server_code_save(struct section_file *file)
...
+  struct player *pplayer = player_by_number(player);
+  if (pplayer == nullptr) {
+    return sol::make_object(lua, "player not found");
+  }
+  if (!pplayer->is_connected) {
+    return sol::make_object(lua, "player not connected");
+  }
+
+  const std::string bytecode = lua["string"]["dump"](f);
+  std::unique_ptr<packet_execute_lua> pack(new packet_execute_lua);
```
lmoureaux, line 299: Prefer `make_unique` over `operator new`.

server/scripting/script_server.cpp
```diff
@@ -275,6 +280,29 @@ static void script_server_code_save(struct section_file *file)
...
+/**
+   Dump the bytecode of function passed and send to client to execute.
+ */
+static sol::object execute_client_lua(sol::this_state s, int player,
+                                      sol::function f)
+{
+  sol::state_view lua(s);
+  struct player *pplayer = player_by_number(player);
+  if (pplayer == nullptr) {
+    return sol::make_object(lua, "player not found");
```
lmoureaux, line 292: Can we raise a lua error instead? (same on line 295)

client/luascript/script_client.cpp
```diff
@@ -307,3 +307,14 @@ static void script_client_signal_create()
...
+
+/**
+   Execute client lua.
+ */
+void handle_execute_lua(int size, const unsigned char *bytecode)
+{
+  luascript_do_bytecode(main_fcl, reinterpret_cast<const char *>(bytecode),
+                        size, "remote_execution");
+
+  return;
```
lmoureaux, line 319: No need for `return` here.

server/scripting/script_server.cpp
```diff
@@ -294,6 +322,9 @@ bool script_server_init()
     return false;
   }
 
+  sol::state_view sol_state(fcl_main->state);
```
lmoureaux, line 325: What kind of permissions does the client lua state have?
Is it reset when disconnecting?
Using the same state as for client scripts allows a broken server to mess up with it, I would prefer a different one.

lmoureaux (changes requested):
Do we need a way for the client to report execution failures?

server/scripting/script_server.cpp
```diff
@@ -275,6 +280,29 @@ static void script_server_code_save(struct section_file *file)
...
+    return sol::make_object(lua, "player not found");
+  }
+  if (!pplayer->is_connected) {
+    return sol::make_object(lua, "player not connected");
+  }
+
+  const std::string bytecode = lua["string"]["dump"](f);
+  std::unique_ptr<packet_execute_lua> pack(new packet_execute_lua);
+  pack->size = bytecode.size();
+  memcpy(pack->bytecode, bytecode.data(), bytecode.size());
```
lmoureaux, line 301: I think Codacy is right that a buffer overflow can happen here?

lmoureaux:
Closing so as not to clutter the list of open PRs -- please feel free to reopen when you get back to it

PR #838, opened 2022-01-20
Auto Revision

### LINKED ISSUE #444: Automatically set the version at build-time from CMakeLists.txt
Conversation in discord started by me asking about how we keep the version of fc21 up to date over time. We have been at 3.0.1337.1-haxx for some time.

Three files note the version
- `./utility/fc_version.h`
- `./CMakeLists.txt`
- `./fc_version`

General consensus is to use a three tiered version system. 
- Major Version
- Minor Version
- GitHub Commit Hash + Date of Hash

Major and Minor version would come from a tag of the code base at a certain point in time. This idea is to use code from `./dist/generate_gitrev.sh` and call from CMakeLists.txt to create `fc_version.h` so that when `--target package` is called a tagged tarball is created and on Msys2 an properly tagged NSIS installer is created.

jwrober:
This commit establishes the basics of getting some status from `git` and using that information when configuring and compiling freeciv21. Does not support the tagged tarball at this point. Right now this assumes you have a working repository and `git` installed.

Eventually closes #444


.github/workflows/build.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
lmoureaux, line 17: I think we'll need the same in `coverity.yaml`

jwrober:
At this point I think this PR is close to ready. Need to figure out how to handle if `git` is not installed and there is no repository to query (effectively a person is compiling from a tarball). If there was a way for the tarball generator to create a file with two values in it, then I could have CMake look that data up as a fall back.

lmoureaux:
> At this point I think this PR is close to ready. Need to figure out how to handle if `git` is not installed and there is no repository to query (effectively a person is compiling from a tarball). If there was a way for the tarball generator to create a file with two values in it, then I could have CMake look that data up as a fall back.

Have a look at https://git-scm.com/docs/gitattributes#_export_subst

cmake/AutoRevision.cmake
```diff
@@ -0,0 +1,62 @@
...
+
+# Auto Revision needs git to work
+find_package(Git)
+
+if(Git_FOUND)
+
+  # get the value of the git hash at HEAD to 5 chars
+  execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --short=5 HEAD
+                  OUTPUT_VARIABLE FC21_REV_HEAD_HASH_H
+                  OUTPUT_STRIP_TRAILING_WHITESPACE)
```
lmoureaux, line 13: What if git is found, but this is not a git repo (eg building from a tarball)?

lmoureaux (commented):
In order to test this, could you make a new branch in your fork and create a release there?

cmake/AutoRevision.cmake
jwrober: Oh that is a good point. I need another if() layer.

jwrober:
One the tests complete, I think this PR is ready to merge. Only "issue" per se is the duplicate of the source tarball and zip at release time. We can point users to the defined assets in the release notes.

cmake/AutoRevision.cmake
```diff
@@ -0,0 +1,73 @@
+# This file runs a collection of git commands to help automate the process
+# of generating the build revision. It takes an update to 3 files by hand
+# to do it manually and automates it.
+
+# Auto Revision needs git to work
+find_package(Git)
+
+if(Git_FOUND)
+
+  # Run git status to see if we are in a real local repo
```
lmoureaux, line 10: Can you specify `--git-dir` in case someone extracts a tarball within some other git-manage area?

lmoureaux (changes requested):
Thought of yet another edge case :1234:
Will hopefully be easy.

cmake/AutoRevision.cmake
jwrober: Pushed another update

lmoureaux (approved):
Let's polish it once it hurts someone

PR #841, opened 2022-01-21
Add support for CPack Debian Linux Package to be built in CI

jwrober:


lmoureaux:
Do we want the package built in `RelWithDebInfo` mode?

Issue #611, opened 2021-08-31
"Present units" clashes with unit orders when investigating cities


lmoureaux:
**Describe the bug**

The "Present units" panel overlaps with the unit description and orders panel when investigating a city.

**To Reproduce**
1. Investigate a city

**Expected behavior**

There is no overlap.

**Platform and version (please complete the following information):**
 - OS: some Linux
 - Freeciv21 version: 3.0-alpha4
 - Ruleset/Longturn game (if applicable): `LTEx24`

**Additional context**

Reported by soundnfury on Discord.

lmoureaux:
Also related are various instances where managing units from the city screen result in the same overlap problem.

lmoureaux:
Got better with #2060, now the city screen can be closed, the unit moved, and the city screen reopened (provided no RTS).

lmoureaux:
Cannot reproduce in current `master`.

PR #862, opened 2022-01-31
Fix performing actions from the city screen

### LINKED ISSUE #819: Disbanding units via right click in city isn't working
**Describe the bug**

@Corbeau-1 on Discord:

> Disbanding units from in-city (right-click) isn't working.

**To Reproduce**
1. See above

**Expected behavior**

Works

 - OS: Windows 10
 - Ruleset/Longturn game (if applicable): `SIM30Test`

lmoureaux:
The condition in most if(can_issue_orders()) was reversed.
Closes #819.

------------

Remains to be seen if I'm the only one not getting the context menu at all. If not, I'll also add the fix to this PR.

### DIFF
--- a/client/gui-qt/citydlg.cpp
+++ b/client/gui-qt/citydlg.cpp
@@ -750,7 +750,7 @@ bool unit_list_item::can_issue_orders() const
  */
 void unit_list_item::disband()
 {
-  if (!can_issue_orders()) {
+  if (can_issue_orders()) {
     auto *units = unit_list_new();
     unit_list_append(units, m_unit);
     popup_disband_dialog(units);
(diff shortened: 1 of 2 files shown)


lmoureaux:
> Remains to be seen if I'm the only one not getting the context menu at all. If not, I'll also add the fix to this PR.

Added anyway.

jwrober:
I went into a city with 3 units - Stealth Bomber, Howitzer and a AWACS. I right clicked on the Howitzer and picked set home city and then the units bar at the bottom disappeared and was replaced with the unit selection bar and I have to close the city dialog and come back in to see the units bar again.
image

jwrober:
Same is true if I do Sentry Unit command. Activate unit does what I expect it to by closing the city dialog. If I do Disband unit, the unit bar stays there, but all the units go into a kind of greyed out state and are not clickable.

lmoureaux:
> I went into a city with 3 units - Stealth Bomber, Howitzer and a AWACS. I right clicked on the Howitzer and picked set home city and then the units bar at the bottom disappeared and was replaced with the unit selection bar and I have to close the city dialog and come back in to see the units bar again.

Another instance of #611

jwrober:
> Another instance of #611

Ok in that case, I can indeed disband units inside cities with a right-click.

jwrober (approved):

PR #867, opened 2022-02-03
Bombard limit effect

### LINKED ISSUE #731: Change the minimum HP to which bombardment can reduce a unit
Bombarding always reduces the unit HP to the hardcoded 1hp. It would be good to generalize that. 

an effect could change the minimum HP. 

The minimum HP can either be as a % of health of the unit or an absolute value

**Additional context**
Keeping it as an effect allows ruleset makers to have improvements and techs which defend against bombarding.

ec429:
Closes #731.

In my testing it appears that if there's no limit specified (so the min is 0), bombing a unit down to 0HP doesn't kill it immediately but it dies on the next TC (with the message "Your <unit_type> has run out of hitpoints").

### DIFF
--- a/docs/Modding/Rulesets/effects.rst
+++ b/docs/Modding/Rulesets/effects.rst
@@ -678,3 +678,11 @@ National_History
 Infra_Points
     City increases owner's infra points by value each turn. If overall points are negative after all cities
     have been processed, they are set to 0.
+
+Bombard_Limit_Pct
+    Bombardment may only reduce units to amount percent (rounded up) of their total hitpoints.  Unit
+    requirements on this effect are the defending unit itself.
+
+    .. note::
+        This effect is added automatically with a value of 1 and no reqs. This behavior can be turned
+        off by requiring the ``+Bombard_Limit_Pct`` option in ``effects.ruleset``.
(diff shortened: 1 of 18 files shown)


ec429:
Unsure if we need to update the shipped rulesets given that rscompat does the right thing with them all.  What's the protocol?

lmoureaux:
> Unsure if we need to update the shipped rulesets given that rscompat does the right thing with them all. What's the protocol?

The idea is to eventually make the capability mandatory, so let's upgrade them now. Also shows how to do it (by looking at the diff).

docs/Modding/Rulesets/effects.rst
```diff
@@ -678,3 +678,7 @@ National_History
 Infra_Points
     City increases owner's infra points by value each turn. If overall points are negative after all cities
     have been processed, they are set to 0.
+
+Bombard_Limit_Pct
+    Bombardment may only reduce units to amount percent (rounded up) of their total hitpoints.  Unit
+    requirements on this effect are the defending unit itself.
```
lmoureaux, line 684: Could you mention the `+Bombard_Limit_Pct` capstring?

PR #870, opened 2022-02-06
Improve zoom

### LINKED ISSUE #295: Zooming artifacts
Seen with hexamplio (notice how hex edges are visible):

image

### LINKED ISSUE #296: Zooming is slow
Zooming in or out takes several seconds on my machines. This is most likely because the whole tileset is being reloaded.

This should be sped up as much as possible. The Gtk client is 2.6 manages to do it instantly, we should at least match that.

### LINKED ISSUE #729: Changing the zoom level while in a city dialog causes the dialog to close
**Describe the bug**
Changing the zoom level while in a city dialog causes the dialog to close

**To Reproduce**
1. Go to default zoom (ctrl-backspace) -- or use the view menu
2. Open a city dialog box
3. Zoom in one level (shift-+) -- or use the view menu
4. Notice the dialog closes and then the map zooms in one level

**Expected behavior**
The map zooms in one level and so does the city center and the city dialog's size shifts to match the new dimensions.

**Screenshots**
N/A

 - OS: All
 - Freeciv21 version: 3.0-Alpha9

**Additional context**
None

### LINKED ISSUE #864: Depiction of worked tiles on the main map is faulty
When you shift-alt-rightclick on the main map, you get city tiles coloured if they are worked by a city. However, in one case, I got this:

image

The real layout is different, there should be definitely more tiles and, as visible in the image, some tiles are only half painted. This could be an issue of city size and map zoom level.

The issue starts fixing itself if I click on surrounding cities.

image

More zoom:

image

Less zoom:

image

lmoureaux:
It no longer reloads the tileset every time, resulting in much faster zooms. Instead, zoom is handled at the mapview level by pretending that the canvas is smaller (or larger) than it really is, and scaling the drawn canvas afterwards. The main drawback is that drawing the map when zoomed out is rather slow. It is, however, quite fast when zoomed in.

City names and production aren't as sharp as they used to be when zooming. Text layers will need to be taken into account when designing the next-gen renderer.

This started as a test to see how far it could go but it turned out to be rather easy so I went ahead.

Closes #295.
Closes #296.
Closes #729.
Closes #864
  (I think).

### Test plan

* Play with zoom in and zoom out.
* Watch out for possible offsets between the tile hovered with the mouse and the tile picked up by the client.
* Check trade route lines, in particular trade planner lines.

### DIFF
--- a/client/citybar.cpp
+++ b/client/citybar.cpp
@@ -460,10 +460,8 @@ QRect traditional_citybar_painter::paint(QPainter &painter,
     return QRect();
   }
 
-  // Select the tileset to grab stuff from
-  const auto t = (gui_options.zoom_scale_fonts || !unscaled_tileset)
-                     ? tileset
-                     : unscaled_tileset;
+  // Get the tileset to grab stuff from
+  const auto t = get_tileset();
 
   // Get some city properties
   const citybar_sprites *citybar = get_citybar_sprites(t);
@@ -621,10 +619,8 @@ QRect polished_citybar_painter::paint(QPainter &painter,
   const bool should_draw_trade_routes =
       can_see_inside && gui_options.draw_city_trade_routes;
 
-  // Select the tileset to grab stuff from
-  const auto t = (gui_options.zoom_scale_fonts || !unscaled_tileset)
-                     ? tileset
-                     : unscaled_tileset;
+  // Get the tileset to grab stuff from
+  const auto t = get_tileset();
 
   // Get some city properties
   const citybar_sprites *citybar = get_citybar_sprites(t);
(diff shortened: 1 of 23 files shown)


jwrober:
Middle-click popup window moves away as you zoom out
1. default zoom
image

2. one zoom out
image

3. a second zoom out
image

jwrober:
Inside the city, the panel does not move when you zoom in
1. default zoom
image

2. zoom in a couple times
image

jwrober:
Other than those two issues, rest looks fine to me.

lmoureaux:
Good catch, fixed both!

PR #871, opened 2022-02-06
Wasm v0 client

zekoz:


client/clinet.cpp
```diff
@@ -143,13 +143,11 @@ static int try_to_connect(const QUrl &url, char *errbuf, int errbufsize)
           client.conn.used = false;
         });
   }
+  QObject::connect(client.conn.sock, &QTcpSocket::connected,
+                   [userName = url.userName()] {
+                     make_connection(client.conn.sock, userName);
+                   });
```
lmoureaux, line 149: This will break autoconnect retry. The function is called as follows:
```cpp
  if (try_to_connect(url, errbuf, sizeof(errbuf)) == 0) {
    // Success! Don't call me again
    autoconnecting = false;
    return FC_INFINITY;
  } else {
    // All errors are fatal
    qCritical(_("Error contacting server \"%s\":\n %s\n"),
              qUtf8Printable(url.toDisplayString()), errbuf);
    exit(EXIT_FAILURE);
  }
```

Error reporting from `connect_to_server` will also be broken by this change. I suggest passing a callback as the last parameter.

lmoureaux (changes requested):

lmoureaux (approved):
Code looks good

PR #872, opened 2022-02-06
Don't generate assertion failures when opening the help dialog

lmoureaux:
**Test plan**

* Open help dialog
* No assertions


jwrober (approved):
Works as advertised.

PR #885, opened 2022-02-08
Conditionally write fc_version.h and fc_version

### LINKED ISSUE #879: Autorevision triggers a full rebuild after every cmake run
**Describe the bug**

Autorevision unconditionally overwrites `utility/fc_version.h` every time CMake is run. Since this file is included pretty much everywhere, touching it triggers a full rebuild.

**To Reproduce**
1. Build
2. `touch CMakeLists.txt`
2. `cmake --build build`
3. Watch it do a full rebuild even though nothing has changed
4. Do something else while the compiler is working
5. Don't implement a new feature

**Expected behavior**

Nothing is rebuilt if nothing changed.

 - OS: any
 - Freeciv21 version: `master`

jwrober:
Closes #879

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,12 +99,17 @@ install(
   COMPONENT freeciv21)
 
 # Use Auto Revision variables to convert some templates to real files at build time
-configure_file(${CMAKE_SOURCE_DIR}/utility/fc_version.h.in
-               utility/fc_version.h
-               @ONLY NEWLINE_STYLE UNIX)
+if(NOT EXISTS "${CMAKE_BINARY_DIR}/utility/fc_version.h")
+  configure_file(${CMAKE_SOURCE_DIR}/utility/fc_version.h.in
+                utility/fc_version.h
+                @ONLY NEWLINE_STYLE UNIX)
+endif()
+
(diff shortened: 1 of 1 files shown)


lmoureaux:
So it will only update on a new checkout... is it the intended behavior?

jwrober:
> So it will only update on a new checkout... is it the intended behavior?

So it will work from a new created build directory and it will work with code changes post build to only rebuild those files that need updating based on the changes. It won't work if you run a subsequent configure in an existing build directory. I don't see that as much of an issue per se. I think typical workflow is to delete the build directory after a fresh pull or checkout so it works in that scenario too. All that said, I do think it will have the intended effect.

lmoureaux:
I tend to never delete my build directories, so in a year from now I might still be using the same version...

jwrober:
> I tend to never delete my build directories, so in a year from now I might still be using the same version...

Interesting. I'm surprised you don't run into configuration issues with all the changes we have made. Do you want to handle this with a different approach?

lmoureaux:
I think merging this would be good to get rid of #879. Eventually we may want to:

1. Move the values to a `.cpp` file
2. Generate a temporary file, compare it with the existing one and replace only if they're different

PR #886, opened 2022-02-09
Remove the minimap thread.

zekoz:
Removes zooming and panning of minimap view, but adds wraparound for viewport rectangle.

### DIFF
--- a/client/gui-qt/minimap.cpp
+++ b/client/gui-qt/minimap.cpp
@@ -45,9 +45,6 @@ minimap_view::minimap_view(QWidget *parent) : fcwidget()
   rw = new resize_widget(this);
   rw->put_to_corner();
   pix = new QPixmap;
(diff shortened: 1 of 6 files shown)


client/gui-qt/minimap.cpp
```diff
@@ -113,177 +113,54 @@ void minimap_view::draw_viewport(QPainter *painter)
     return;
```
lmoureaux, line 113: The variables `i`, `src_x`, `src_y`, `dst_x`, and `dst_y` (declared above) are no longer used.

lmoureaux (changes requested):
Artifacts accumulate along the edges (`WRAPX|WRAPY|HEX`):
image

A stupidly small map or sufficient zooming out can give this result:
image

lmoureaux (approved):

PR #888, opened 2022-02-10
Local sockets

lmoureaux:
This PR does two things:

* Fix the `--bind` server parameter
* Implement local sockets and use them to communicate with the server spawned by the client for local games

Code is a bit ugly because Qt doesn't leverage OO programming enough in its `QIODevice` hierarchy.

### DIFF
--- a/client/connectdlg_common.cpp
+++ b/client/connectdlg_common.cpp
@@ -239,7 +239,7 @@ bool client_start_server(const QString &user_name)
     return false;
   }
 
-  // Unique name for this game
+  // Generate a (random) unique name for the local socket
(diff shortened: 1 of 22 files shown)


client/connectdlg_common.cpp
```diff
@@ -238,15 +239,17 @@ bool client_start_server(const QString &user_name)
     return false;
   }
 
+  // Unique name for this game
+  auto uuid = QUuid::createUuid().toString(QUuid::WithoutBraces);
```
zekoz, line 243: What's this for?

zekoz (approved):

PR #896, opened 2022-02-12
Remove the web-compatible capability from rulesets

### LINKED ISSUE #863: Drop web-compatible ruleset capabilities
**Describe the bug**

Some rulesets have the `web-compatible` capability to signal that the JavaScript Freeciv-web client understands them. Since the JS client is being phased out in favor of emscripten, we should drop these caps.

**To Reproduce**
1. Check capabilities of the `classic` ruleset.

lmoureaux:
Our rules are not meant to be used with the JS client.

Closes #863.

### DIFF
--- a/data/civ2civ3/buildings.ruleset
+++ b/data/civ2civ3/buildings.ruleset
@@ -12,7 +12,7 @@
 
 [datafile]
 description="Civ2Civ3 buildings data for Freeciv"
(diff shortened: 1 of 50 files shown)


lmoureaux:
Rulesets still load, it's removing dead "code" and quite straightforward so merging.

PR #907, opened 2022-02-13
Try once again to fix stdin handling on Windows

### LINKED ISSUE #850: Server started from Windows Explorer blocks on input
**Describe the bug**

Apparently we still have this bug.

**To Reproduce**
1. Start a server from Windows Explorer
2. Start a client, go to network page
3. Don't see your server
4. Go to server window, hit a key
5. Go back to client, see your server listed
6. Try to connect, seems to block
7. Go to server window, hit a key
8. Connect unblocks

**Expected behavior**

Client seamlessly detects the server and connects.

 - OS: Windows 10

**Additional context**

The server event loop is blocked waiting for input on stdin. Move input to a thread.

lmoureaux:
On Windows, there is apparently no unified non-blocking API to wait for 
input on stdin.  Thus, call the blocking API from a worker thread.

Tested the following use cases (*before rebasing*):

* Starting from the Windows console
* Starting from Windows Explorer (should be the same, but who knows)
* Starting from the client (I guess stdin is a named pipe in this case)
* Starting from the MSYS console (no idea what it uses, maybe a file)

The thread is only used on Windows, no reason for the extra overhead to
propagate to sane OSes.

Closes #850.

This should also be tested on Linux even though I don't expect differences there.

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -1,3 +1,5 @@
+set(CMAKE_AUTOMOC true)
+
 # Generate the packet handling code
 add_custom_command(
   OUTPUT
--- a/server/server.cpp
+++ b/server/server.cpp
@@ -30,10 +30,7 @@
 
 // Stuff to wait for input on stdin.
 #ifdef Q_OS_WIN
-#include <QtCore/QWinEventNotifier>
-#include <array>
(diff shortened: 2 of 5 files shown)


jwrober (approved):
Works for me!

PR #912, opened 2022-02-15
Themeable icons

zekoz:


client/gui-qt/icons.cpp
```diff
@@ -39,21 +44,84 @@ fcIcons *fcIcons::instance()
...
+namespace {
+
+class icon_engine : public QIconEngine {
+public:
+  explicit icon_engine(const QString &path) : path(path) {}
+
+  icon_engine *clone() const override { return new icon_engine(path); }
+
+  void paint(QPainter *painter, const QRect &rect, QIcon::Mode mode,
+             QIcon::State state) override
```
lmoureaux, line 56: I guess this is only ever used with `state = QIcon::On`? Do we want a different color for `mode = QIcon::Active`, maybe the active text color (`QPalette::Active`)?
Add `Q_UNUSED` to avoid compiler warnings.

client/gui-qt/icons.cpp
```diff
@@ -39,21 +44,84 @@ fcIcons *fcIcons::instance()
...
+      pix.fill(Qt::transparent);
+      if (!path.isEmpty()) {
+        QSvgRenderer renderer;
+        QFile f(path);
+        QByteArray bytes;
+        if (f.open(QIODevice::ReadOnly)) {
+          bytes = f.readAll();
+        }
+        if (!bytes.isEmpty()) {
+          bytes.replace("#000", col.name().toLatin1());
```
lmoureaux, line 73: This will break if I add a path with color `#000FFF`.

client/gui-qt/icons.cpp
```diff
@@ -39,21 +44,84 @@ fcIcons *fcIcons::instance()
  */
 void fcIcons::drop() { NFCN_FREE(m_instance); }
 
+namespace {
+
+class icon_engine : public QIconEngine {
```
lmoureaux, line 49: It would be good to add a few words about what this icon engine does differently, maybe "Icon engine that loads SVG images and changes the color of black (`#000`) paths to match the text color."

PR #917, opened 2022-02-19
Remove qtg_ indirection for remaining canvas functions

lmoureaux:
This brings `QPainter` into the core drawing code, removing the indirection layer and shaving another 400 lines of code.

The affected areas in the clients are:

* The main map (sprites, goto lines, trade routes, … virtually everything but city names)
* The minimap (visible area polygon)
* The tech tree (everything)
* The spaceship report


jwrober (approved):
I ran into no issues with the game while testing this patch. Did find another issue, but its not releated.

PR #920, opened 2022-02-20
Hex corner terrain sprite type

lmoureaux:
The existing "corner" method doesn't work well in hexagonal mode because for N
matching groups ("match type"), it requires N^4 sprites. It gues without saying
that this explodes very quickly. In hexagonal mode, each corner is the meeting
point of only three tiles, making a O(N^3) method a better match.

This patch implements such a method, requiring 2 * N^3 sprites for N groups. It
is limited to ISO|HEX to prevent keep code complexity under control.

The sprite names take the following form:

    t.l[N].hex_cell_[left|right]_[g]_[g]_[g],

where the [g] are the first letters of the matching groups. It is expected that
the sprites are cut along the dotted lines in the following schema, which also
shows what `left` and `right` correspond to:

    ____
    .:..\...:.
     : R \__:_
    .:.../..:.
    _:__/ L :
    .:..\...:.
     : R \__:_
    .:.../..:.
    _:__/   :

Tested with Vectron dev version; shouldn't affect other modes in any way.

### DIFF
--- a/client/layer_terrain.cpp
+++ b/client/layer_terrain.cpp
@@ -310,7 +310,6 @@ void layer_terrain::initialize_terrain(const terrain *terrain)
       break;
     }
     break;
-    break;
   case CELL_HEX_CORNER:
     if (tileset_topo_index(tileset()) != TS_TOPO_ISOHEX) {
       tileset_error(tileset(), LOG_ERROR,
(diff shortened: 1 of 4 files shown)


lmoureaux:
I'm writing documentation, but this will come in another PR.

jwrober (approved):

PR #921, opened 2022-02-23
Improve the documentation of tileset terrain support

lmoureaux:
I wrote documentation of the terrain features of the map drawing engine based on my experience with Vectron, including the new `hex_corner` mode. I don't claim that it's fully correct, but it's certainly better than what we had (which was both outdated and unclear). Hopefully this reduces the amount of dark magic involved in authoring tilesets.

Since I was adding REUSE info for the new images and files, I also added them to some areas where I was the main contributor.


jwrober:
I will have a chance to fully review this over the weekend. Lots of great content in there.

jwrober (approved):

PR #923, opened 2022-02-24
Handle fuel effects of unit conversion/upgrade

ec429:
When transforming a non-fuel-using unit to a fueled unit, fill up the
 tanks.  When transforming between units with different fuel, add the
 difference between their fuel values (thus keeping the ullage constant
 and ensuring that a unit that was fully fueled before is fully fueled
 after), except never reduce fuel below 1.
If a unit is converting in the same turn it runs out of fuel, and the
 conversion would either increase fuel or turn it into a non-fueled
 unit, then don't kill the unit; conceptually, let the conversion
 happen before the fuel exhaustion.  This makes for a less confusing UX
 as players can start converting on the turn that the unit shows as
 'low on fuel'.

### DIFF
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -434,6 +434,34 @@ static void do_upgrade_effects(struct player *pplayer)
 }
 
 /**
+   Return true if unit is about to finish converting to a unittype
+   with more/no fuel, rescuing it from imminent death
+ */
+static bool converting_fuel_rescue(struct unit *punit)
+{
+  const struct unit_type *from_type, *to_type;
+
+  if (punit->activity != ACTIVITY_CONVERT)
+    /* not converting */
+    return false;
+  if (punit->activity_count + get_activity_rate_this_turn(punit)
+      < action_id_get_act_time(ACTION_CONVERT, punit, unit_tile(punit),
+                               punit->activity_target))
+    /* won't be finished in time */
+    return false;
+  if (!unit_can_convert(punit))
+    /* can't actually convert */
+    return false;
+  from_type = unit_type_get(punit);
+  to_type = from_type->converted_to;
+  if (utype_fuel(to_type) && utype_fuel(to_type) <= utype_fuel(from_type))
+    /* to_type doesn't have more fuel, so conversion doesn't help */
+    return false;
+  /* we're saved! */
+  return true;
+}
+
+/**
    1. Do Leonardo's Workshop upgrade if applicable.
 
    2. Restore/decrease unit hitpoints.
@@ -580,7 +608,8 @@ void player_restore_units(struct player *pplayer)
   // 7) Check if there are air units without fuel
   unit_list_iterate_safe(pplayer->units, punit)
   {
-    if (punit->fuel <= 0 && utype_fuel(unit_type_get(punit))) {
+    if (punit->fuel <= 0 && utype_fuel(unit_type_get(punit))
+        && !converting_fuel_rescue(punit)) {
       notify_player(pplayer, unit_tile(punit), E_UNIT_LOST_MISC, ftc_server,
                     _("Your %s has run out of fuel."),
                     unit_tile_link(punit));
@@ -1506,6 +1535,16 @@ void transform_unit(struct unit *punit, const struct unit_type *to_unit,
     punit->moves_left = unit_move_rate(punit);
   } else {
     punit->moves_left = punit->moves_left * unit_move_rate(punit) / old_mr;
+  }
+
+  /* Update unit fuel state */
+  if (utype_fuel(to_unit) && utype_fuel(old_type)) {
+    /* Keep the amount of fuel used constant, except don't kill the unit */
+    int delta = utype_fuel(to_unit) - utype_fuel(old_type);
+    punit->fuel = MAX(punit->fuel + delta, 1);
+  } else {
+    /* When converting from a non-fuel unit, fully refuel */
+    punit->fuel = utype_fuel(to_unit);
   }
 
   unit_forget_last_activity(punit);


server/unittools.cpp
```diff
@@ -1508,6 +1537,16 @@ void transform_unit(struct unit *punit, const struct unit_type *to_unit,
     punit->moves_left = punit->moves_left * unit_move_rate(punit) / old_mr;
   }
 
+  /* Update unit fuel state */
+  if (utype_fuel(to_unit) && utype_fuel(old_type)) {
+    /* Keep the amount of fuel used constant, except don't kill the unit */
+    int delta = utype_fuel(to_unit) - utype_fuel(old_type);
+    punit->fuel = MAX(punit->fuel + delta, 1);
```
lmoureaux, line 1544: This looks strange. If I have 1/10 fuel, I end up with 91/100. If I have 100/100 fuel, I end up with 1/10.

ec429: No you don't?  In your second case, `delta = 10 - 100` = -90, so you end up with `MAX(100 - 90, 1)` i.e. 10/10.  (A fully-fueled unit should always remain fully-fueled.)

lmoureaux: Eeeh right.

lmoureaux:
Auto clang-format

ec429:
Auto clang-format

Lexxie9952:
it covers almost every case better than before, (which was: punit->fuel just remained unchanged)

but it does not catch every possible case in the logic tree fairly or cleanly.

Lexxie9952:
> Thanks @Lexxie9952. Can you explain the edge cases you are solving ? It helps us document the code better if we are using your suggestion.

Yeah sure. First to clarify the vars:
```
  int old_fu = utype_fuel(old_type),
      new_fu = utype_fuel(to_unit),
      cur_fu = punit->fuel;
```

Then, I know the one line of code is pretty amazing to cover it all in one line, but not easy to understand. So here I added some comments above it
```
  /* Scale Fuel. Avoid "unfair" death by covering every case in the logic tree:    
   * CASE 1. Old type has no fuel? Fill up with 100% fuel.  ==(Don't start with no fuel)
   * CASE 2. New type has more fuel? Add the difference.    ==(Same amount of burnt fuel)
   * CASE 3: New type has less fuel? Keep max.legal amount. ==(Avoid unfair death) */ 
  punit->fuel
    = (!old_fu) ? new_fu : MAX(cur_fu + (new_fu - old_fu), MIN(cur_fu, new_fu));
```
I don't want to say this is the ONLY right way, since CASE 2 is potentially arguable, but you have to make a strong argument against one of those three cases not working out the way it's currently coded to operate. The other proposed solution, while certainly better than before, has one case where it's worse than before, and also doesn't handle CASE 3 properly. So it has two problems.

lmoureaux:
Considering the following hypothetical units:

| Unit type | Fuel |
|:-|:-|
| A | 10 |
| B | 20 |

Let's say we have actions that let one convert back and forth between the types. Now considering the following conversion sequence, I get the following for the two proposals:

| Step | ec429 | Lexxie |
|:-|:-|:-|
| Start with A | 8/10 | 8/10 |
| Convert to B | 17/20 | 17/20 |
| Convert back to A | 6/10 | 9/10 |

Units start at 8/10 because I decided so, and lose 1 fuel per step because the conversion takes one turn.

It looks like Lexxie's formula allows weird cases where a unit ends up with more fuel than it started with. While it's hard to determine what's the "right" thing to do here, the possibility of gaining fuel seems very much "wrong".

pranavtifr:
While, it seems like it makes sense to keep fuel from going from a high fueled unit to a low fuel one. I think there is a fundamental problem when you are going from a low fueled unit to a high fueled one. You are technically creating fuel out of thin air, since you are upgrading your unit but matching fuel *used* not the fuel the unit has. 

This will obviously create problems where you can go back and forth a unit and it seems to create fuel magically. snf's implementation seemed to alleviate this a bit by being too punishing on going back from high fuel to a low fueled unit, it compensates for the magical fuel gained when back from low fuel to a high fueled unit. So, I think the bigger problem is the assumption we need to keep ullage constant, in both the implementations, whereas if you dont want to gain fuel magically. You want to keep fuel level constant irrespective of max fuel.

ec429:
Certainly the ullage approach isn't perfect.  But I chose it because it generalises from the following special case: a fully-fuelled unit should convert to a fully-fuelled unit.  Is that "creating fuel out of thin air", or does the new unit just have a slower fuel consumption?  The ideal way to represent the latter would be by preserving the fuel _fraction_ on conversion, but not only does that raise further design questions (which way should you round?), it _still_ doesn't prevent creating fuel from thin air, because (still) it's necessary to never leave a unit with less than 1 fuel.

I'm not dead set on any particular answer, because the only cases my ruleset cares about are:
* either `new` or `old` is 0 (submarines converting to/from surfaced)
* `new` == `old` (bomber conversions like BH/BD or LM/FH)
* upgrades, which can only happen in a city (so all that matters is result ≤ `new`)

for all of which the right answer is uncontroversial.  But I can't see a good way to add support for those while not codifying behaviour for the rest (e.g. if you try to reject rulesets that don't stick to these cases, you have to ensure that the Upgrade Unit actionenabler always requires a city if any unit's fuel changes on upgrading, and what about the `Upgrade_Unit` effect, and…) so we have to pick _something_.

Lexxie9952:
Dance baby dance. It is true the burnt fuel thing is arguable. How often do fuel units get transformed, it's a question that non-ESFP ask themselves every day, or probably not. This will tire you out forever with no resolution until you realise I'm right. I'm fine with doing it because u know how it goes... or not ? n'est-ce pas ? On peut se tutoyer ou non ? 
Anyway you're welcome for saving you from the other case, you know the one !

Lexxie9952:
> Certainly the ullage approach isn't perfect. But I chose it because it generalises from the following special case: a fully-fuelled unit should convert to a fully-fuelled unit. Is that "creating fuel out of thin air", or does the new unit just have a slower fuel consumption? The ideal way to represent the latter would be by preserving the fuel _fraction_ on conversion, but not only does that raise further design questions (which way should you round?), it _still_ doesn't prevent creating fuel from thin air, because (still) it's necessary to never leave a unit with less than 1 fuel.
> 
> I'm not dead set on any particular answer, because the only cases my ruleset cares about are:
> 
> * either `new` or `old` is 0 (submarines converting to/from surfaced)
> * `new` == `old` (bomber conversions like BH/BD or LM/FH)
> * upgrades, which can only happen in a city (so all that matters is result ≤ `new`)
> 
> for all of which the right answer is uncontroversial. But I can't see a good way to add support for those while not codifying behaviour for the rest (e.g. if you try to reject rulesets that don't stick to these cases, you have to ensure that the Upgrade Unit actionenabler always requires a city if any unit's fuel changes on upgrading, and what about the `Upgrade_Unit` effect, and…) so we have to pick _something_.

I'm tricking you into slowly realising something by first objecting then realising I'm right... I'm not your enemy but have to first make you understand it...

lmoureaux:
Did the `clang-format` by hand

lmoureaux:
Since we didn't find any convincing reason why the proposed implementation would be "bad" and none of the other proposals is unambiguously "better", I'm merging. We can still change the formula later if we find a critical issue, it's not like we're going to fuse the code in PROM.

Lexxie9952:
I warned that this is a rabbit-hole of nightmares to get into and tried to just give you the clean solution that handles everything in the best way possible.

My fear was that from personal antipathy, Louis would think hard to find a very trivial edge cases in my solution. And he did. He found one I knew about that only matters if it's sci-fi Klingon spaceships that can tranform in the middle of outer space from type A to type B AND then do a "reverse Leonardo" back again from B to A, all in the middle of the trip and not at the Spaceship yards where ships normally get upgraded (and downgraded too!), on Planet Xogumu!!!) Let me be clear, the flaw he found is an accurate one, even if non-existent problem in any ruleset likely to ever be written. So for this, good job.

What's not so good? He used this example of saving the Klingons as his reason to go with the formula that causes numerous REAL edge cases in REAL rulesets that are a lot easier to find than the "Klingon exception". The edge cases he chooses to keep have a much greater negative impact on REAL EXISTING RULESETS such as MP2, LTT, and Aviation rulesets. Why search so hard for very difficult non-existent case (which you could also fix with an if-else nest), and ignore the real problem cases that were easier to find? When it comes to server code we need more objectivity than personal bias.

I was aware of this edge case (and another "non-existing" edge case). Apparently Louis is not aware that all solutions have edge cases in them. This is a problem space that goes into billions of distinct situations. The existence of an edge case cannot be what decides the solution. I'll tell you why. Metaphysical/conceptual flaws in freeciv mechanics make it impossible to have no edge  case EXCEPT under one situation, a hive of nested if-else for literally billions of cases, where each and every one of the billions of cases is decided optimally by itself. Therefore, when deciding a solution, it can't be his m.o. of "find a single trivial edge case and use it to dismiss the superior solution." I would propose the professional conduct here is rather common sense to most people. It is something like a weighing/scoring of the solution to give superiority index. Something more like Score S = P - E - C,
where P = [superior rationality,balance,realism, and playability of the solution in
the great majority of cases], and
E = [(quantity of edge cases) * (impact severity of those cases)], and
C = complexity of the solution for players to understand and figure in their head, as they play
the game.

I argue that this analogy of a decisionmaking process is better than S = "I don't like you, therefore I analyse a non-existing edge case and use it as excuse to dismiss the superior solution for one that has some obvious edge case problems, while 
not even spending 1/10th the time finding the more easier-to-find and larger problems in the solution I chose."

Sorry for length of response, I'm just in shock at the brilliance of finding the edge case, the non-brilliance of ignoring its non-applicability to any ruleset ever written, the ignoring of how easy it is to solve that edge case and keep all superior mechanics that avoid the problems he is chosing to ignore here.

lmoureaux:
I understand that facing the judgement of others is not always easy. Here we try to judge code and game mechanics, not people.

The code I merged works, is readable, and was provided in a way that makes it easy to handle. It may not be bug-free but I have yet to find a real issue with it. The code you proposed also works, but it needs 4 lines of comments to understand what it does, which is a red flag that it should be simplified; furthermore it wasn't proposed in a standard way [[1]](https://longturn.readthedocs.io/en/latest/Contributing/pull-request.html). The game mechanics aren't very different and no compelling argument was given to prefer your solution -- had this been the case I'm quite sure ec429 would have changed his code.

For future contributions I would advise you to open issues (for things you'd like to report without providing code) or pull requests (if you have code). This will dramatically increase the chances that your contributions be accepted. If there is an issue in an open pull request feel free to take part in the review by explaining the problems with the code and we'll try to find a solution together --- it may be something nobody thought of in the first place. One last tip: try to be specific in your communication, vague statements are usually not useful.

Thank you for your understanding.

Lexxie9952:
Your own example put it in the light of day what's wrong. Only someone in the mindset of kangaroo court wouldn't see his own example proving the case! LOL.  Your own people commenting: "While, it seems like it makes sense to keep fuel from going from a high fueled unit to a low fuel one" then you use the words "we didn't find anything" when, it's right there. Your own chart shows a unit LOSING fuel when it downgrades... this causes edge cases of "unfair death". Right there in your own example staring at you, but you were so busy making the case for a ruleset of "Origami" units in a super special case that has not nor will not ever exist, without realising right in the example you give is the proof of all the REAL rulesets that will have bad cases. Not from origami units who change like mechatron-bot transformers while out in the air or outer space on fuel missions... but in a real game out on patrol and Leonardo came down and gave them a death sentence. 
LOOK AGAIN
<html>
<body>


Step | ec429 | Lexxie
-- | -- | --
Start with A | 8/10 | 8/10
Convert to B | 17/20 | 17/20
Convert back to A | 6/10 | 9/10

<p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(201, 209, 217); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Units start at 8/10 because I decided so, and lo</p>
</body>
</html>

Lexxie9952:
ec429 and I are not in a competition and both want the better result.

Unit_type A: Plane with 4 fuel, and 2 fuel left, flying home ... start with A
Unit_type B: Plane with 2 fuel


<html>
<body>


Step | Louis| Lexxie
-- | -- | --
Start with A | 2/4 | 2/4
Leonardo to to B | DEAD| 1/2

</body>
</html>
Anyone who tried can see this in 1/100th the effort of coming up with strange ruleset of Klingon Origami warships who change from A to B to A again in outer space without even landing at the spaceship upgrade factories. So what we're revealing here appears to be a bias in my opinion, to use an example that proves my case right in the open, but has to slant to the rare subset of "Origami" units who are so special they don't exist:

I will quote from my own long extensive white paper on the hours of theoretic work I did on this:
G8. {Origami Units} = This is a very special theoretical set of units. They are the units that...
a. Have 3 or more turns of fuel and/or convert to a type that does.
b. Both the types it starts as, AND the type it converts to, has fuel.
c. The type it that it converts to, also converts to a type that also has fuel.
d. The type it starts from can legally be given the order to do the activity "Convert Unit",
by the player, to change its type, outside a city or refueling tile, while out on a fueled mission.
e. The type it converts to can also legally be given the order to do the activity "Convert Unit",
by the player, to change its type, outside a city or refueling tile, while out on a fueled mission.
f. The type it converts to (A) and the type which that type converts to, (B) have...
g. different amounts of fuel, and,
h. the type that was converted to can now also convert to another type which has less
   fuel than the converted type, and,
i. that amount of fuel for the second converted type is less than the first converted type, and,
j. the amount of fuel in the original type is less than the first type it converts to, and,
k. the difference in the amount of fuel between all the types is enough such that another conversion
   first converted type to the second converted type with less fuel would create more fuel IF:
l. we follow the rule that a unit should never lose fuel if the new type of unit it converts to has
   enough fuel capacity to hold the amount of fuel that it had in its tanks before the conversion.
g. None of the types it starts from or converts to have 1 turn of fuel.
********** Number of known "Origami" units in existing freeciv rulesets: 0
             "     "   "       "       "    " likely freeciv rulesets: 0

This is what's being used by you that you somehow came up with, and even in your own example it's staring right back the much more natural cases of real rulesets that will have St. Leonardo, patron saint of plane crashes, giving out random death sentences by stealing fuel from a tank even when the the new fueltank has enough capacity to keep the fuel we already had. i.e., THAT IS, even worse in a way than leaving punit->fuel how it was before you ever accepted the pull request !

ec429:
tl;dr: Go home Lexxie, you're drunk.

The "origami unit" is not all that theoretical.  Aviation already has units (submersibles/U-boats) that can Convert Unit in the field, back-and-forth between two unittypes.  Only one of the two has nonzero fuel, so it doesn't fulfil all the conditions for this to matter.

The "unfair death" you're so worked up about requires all of the following conditions:

1. A fueled unittype (A) has an upgrade.
2. The upgrade (B) is also fueled.
3. B has less fuel than A.
4. The ruleset has a Leonardo (by which, to be precise, we mean an `Upgrade_Unit` Effect, which is typically but not necessarily provided by a Wonder).
5. Effect `Upgrade_Unit` can apply to units for which the `Upgrade Unit` Action is not actionenabled (this appears to be true of current `freeciv` server versions, but I would argue this is a BUG), **or** an `Upgrade Unit` actionenabler allows air units to upgrade in-flight (currently most rulesets do, but only because `ACTION_UPGRADE_UNIT` has `ATK_CITY` and thus already can only happen in a city so there's been no need to softcode that).

Now, let us consider existing rulesets.  In core rulesets and LTT (1) and (2) are true (planes have fuel) but AFAIK (3) is not.  In Aviation, (1-3) all hold (e.g. Twenties Light Bomber upgrading to Thirties Light Bomber), but (4) does *not* (this was a deliberate decision, there are several ways in which upgrades in Avi can involve some minor capability loss, so I decided it should never happen without user intervention).

The only ruleset I'm aware of for which there is a potential issue is LTX, in which the 7-fuel unique Longboat upgrades to the 6-fuel Caravel.  But I think (5) should be fixed, because it makes no sense for Leonardo to upgrade units in situations where they couldn't be upgraded with gold.  At which point the problem goes away, because upgrading a unit in a city is always safe from a fuel perspective.

As for all the semantic arguments about "stolen fuel" and "miraculously from God awarding fuel out of thin air to avoid death"… I think people are being mislead by the name.  It's not really _fuel_ as such, it's the _endurance_ of a unit; the number of turns it can spend in the field.  (This is literally how the generated helptext describes it: "unit must be on a carrier after _n_ turns" or whatever.)  And the "ullage" is precisely the number of turns since it was last in a base.  If I'd been implementing freeciv from the start, instead of `unit->fuel` counting down to zero, it would count _up_ to `unittype->fuel`.

And the case when we "award fuel out of thin air" is intended specifically for the situation where a unit which has just arrived in a city, and thus potentially has low fuel, is upgraded; I wasn't confident that the game would correctly handle a unit that already has 0 fuel at the start of `player_restore_units()`.  (On re-reading it, I _think_ it might actually be fine, in which case line 1544 could just skip the `MAX()` entirely; but I'd have to test that.)

Unless you're referring to the `converting_fuel_rescue()`, but that isn't "awarding fuel" at all.  It's simply changing the logical order of the "convert unit" and the "kill units with empty fuel tanks" processing, because players will **expect** to be able to convert a unit on its last turn of fuel, rather than having to do so the turn before.  And if we didn't do that, then they'd **definitely** feel the resulting death was "unfair".


All of which is somewhat of a **side-issue**.  The real reason no-one cares about your opinions, Lexxie, is that the way you're presenting them does you no favours.  There's a thread of supreme arrogance running through your posts, to the effect that obviously your solutions are Correct and thus the only purpose of the conversation is for the rest of us to See The Light.  (Not to mention that you seem to think it's "pretty amazing" that you managed to perl-golf your way to a one-liner; no-one is impressed by your ability to use the ternary operator.)  And you keep on bringing in unrelated personal remarks and grandiose accusations.  Calling someone a "kangaroo court" does not endear you to them.  Nor does your assumption that simply because I've been polite to you thus far I therefore "will side with what [you're] saying" (to be abundantly clear, I don't).  For the record, Louis didn't find the edge case that breaks in your solution; I did, and communicated it privately, because I thought this discussion would involve less noise, acrimony, and shouting that way.  Boy did I call _that_ one wrong ;)  I didn't have any "personal antipathy" to you going into this, but I sure do now, because you've demonstrated yourself to be petulant, narcissistic, and generally not the sort of person I want to expend time and energy collaborating with on software projects which we are, after all, all doing _for fun_ in our free time.  And that even if your solution _were_ technically superior, I'd now recommend against adopting it, simply because you'd interpret your having "won" as license to continue and extend your toxic behaviour, to the detriment of the wider project and community.

If you want to be taken seriously, try to contribute more light than heat.

Lexxie9952:
Well for all this rant, I lack any antipathy or personal issue with you, and wish you had not started any with me since I did not start with you. 

Your units are not our definition of an "Origami unit." None of those exist in any ruleset. Converting in the field is certainly not the qualifier for such, it's a WHOLE lot more, any more than a hole in the ground qualifies as a cosmological wormhole.  In freeciv development, I respect a person strictly on the quality of their output, its innovation and special usefulness for solving outstanding balance problems. Therefore, I respect you in spite of all this personal attack, which just gets in the way of maximized collaborative unified MindProduct. You're doing and looking at things that I have been trying to get other devs to look at for years, but only get arrogant dismissal.

If you can keep the personal attacks out of the issue and the "I wouldn't use a better solution for no other reason than it's a certain person" also out of it -- I would certainly never stoop so low myself -- let's get back to the main subject. This is a greater community where such "metaphysical mechanics" should strive to conformity and collaboration across all the intelligentsia of various sites. The case where Leonardo upgrades a fuel unit in the field, is not for us to decide or debate since neither of us makes all rulesets. Due to the well-known "molasses effect" that punishes sea units in the game, we should be very thoughtful about any arbitrary server changes that demand others' rulesets to intensify a known balance issue in the game. We may or may not agree on it but at FCW we've come to prioritise in the order of balance, playability, realism; so a server change that forces rulesets to #3 over #2 and #1 instead of giving a solution with maximum universality and flexibility, hmmm, I'll let you reconsider and not pin it on you that you're espousing such.

Leonardo-type upgrades in the game are quite common, and are about to get a lot more common as certain under-utilised systems in freeciv start to get used more to find subtly improved balance. Those are: use of fuel in more types of units, and amplified "Upgrade_Unit" effects to cure the lag of 500 year old units lingering too long.

What is not going to stay around too long are the set of {Origami units}. We're like those astrophysicists talking about whether in theory some kind of wormhole exists or not, LOL. Because Origami units have never been seen yet.  If you're really interested I'll post all 9 of the metaphysical concept definitions we made in order to ultimately solve this punit->fuel problem. You'd be interested a lot on the concept of freeciv-fuel vs real fuel, I think, since your reply quoted a little of what we came up with. 

It turns out we simultaneously and very similarly, came up with a punit->fuel solution and think similarly on other things which no one else in the Freeciv world has bothered with for decades. This ought to be cause for celebratory camaraderie rather than unilateral personal attack from you onto me. Please reconsider your stance of antipathy > collaboration, and if you hold it, divergence > convergence, "go home" > "welcome and thanks for taking a passionate interest in this!"

Cheers.

PR #924, opened 2022-02-25
Unit subsurface vision (sonar)

### LINKED ISSUE #893: Support vision_subs_sq on units
Bases can grant vision of submarine units at range, with `vision_subs_sq`; but units can't have the same thing.

Add a unittype field `vision_subs_sq` that sets the unit's vision range for `vision_layer = "Subsurface"` units.

Could add a `vision_layer` req to the `Unit_Vision_Radius_Sq` effect; but this would require ruleset compat to add `"Vision Layer", "Surface", "Local", TRUE` to all existing instances of the effect.

**Additional context**
I'm fairly sure upstream freeciv has unit.vision_subs_sq as of 3.1.

ec429:
Add a requirement type `Vision_Layer` that allows `Unit_Vision_Radius_Sq` or `City_Vision_Radius_Sq` effects to apply to `Subsurface` or `Stealth` layers instead of `Main`; make those bonuses apply _after_ subs/invis vision range is capped to 2; and fix a couple of bugs in vision handling.
Add a ruleset effects capability `+Vision_Layer`; and compatibility for old rulesets to retain existing behaviour.  Update the shipped rulesets to retain their old behaviour (except in the case of `stub` where we omit the req on city vision for simplicity).

Closes #893.

### DIFF
--- a/common/effects.cpp
+++ b/common/effects.cpp
@@ -691,7 +691,8 @@ int get_player_bonus(const struct player *pplayer,
 /**
    Returns the effect bonus at a city.
  */
-int get_city_bonus(const struct city *pcity, enum effect_type effect_type)
+int get_city_bonus(const struct city *pcity, enum effect_type effect_type,
+		   enum vision_layer vlayer)
 {
   if (!initialized) {
     return 0;
@@ -699,7 +700,7 @@ int get_city_bonus(const struct city *pcity, enum effect_type effect_type)
 
   return get_target_bonus_effects(
       nullptr, city_owner(pcity), nullptr, pcity, nullptr, city_tile(pcity),
-      nullptr, nullptr, nullptr, nullptr, nullptr, effect_type);
+      nullptr, nullptr, nullptr, nullptr, nullptr, effect_type, vlayer);
 }
 
 /**
--- a/common/effects.cpp
+++ b/common/effects.cpp
@@ -692,7 +692,7 @@ int get_player_bonus(const struct player *pplayer,
    Returns the effect bonus at a city.
  */
 int get_city_bonus(const struct city *pcity, enum effect_type effect_type,
-		   enum vision_layer vlayer)
+                   enum vision_layer vlayer)
 {
   if (!initialized) {
     return 0;
@@ -843,10 +843,9 @@ int get_unittype_bonus(const struct player *pplayer,
     pcity = nullptr;
   }
 
-  return get_target_bonus_effects(nullptr, pplayer, nullptr, pcity, nullptr,
-                                  ptile, nullptr, punittype, nullptr,
-                                  nullptr, nullptr, effect_type,
-                                  vision_layer);
+  return get_target_bonus_effects(
+      nullptr, pplayer, nullptr, pcity, nullptr, ptile, nullptr, punittype,
+      nullptr, nullptr, nullptr, effect_type, vision_layer);
 }
 
 /**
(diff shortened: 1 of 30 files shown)


lmoureaux:
Network compat: since this is the first new requirement kind I see, I'm curious how old client react to effects using it

ec429:
> I'm curious how old client react to effects using it

Connecting an old client to a new server that uses VUT_VISIONLAYER, the client logs assertions `universal_number(): invalid source kind -1.` and `Invalid source kind -1.` (the latter from `universal_name_translation()`), but the actual vision handling appears to still operate correctly (though of course Help>Effects displays the req as an empty string).

common/effects.cpp
```diff
@@ -601,26 +601,15 @@ bool is_building_replaced(const struct city *pcity,
...
 int get_target_bonus_effects(
     struct effect_list *plist, const struct player *target_player,
     const struct player *other_player, const struct city *target_city,
     const struct impr_type *target_building, const struct tile *target_tile,
     const struct unit *target_unit, const struct unit_type *target_unittype,
     const struct output_type *target_output,
     const struct specialist *target_specialist,
-    const struct action *target_action, enum effect_type effect_type)
+    const struct action *target_action, enum effect_type effect_type,
+    enum vision_layer vision_layer)
```
lmoureaux, line 612: This new function would need a docstring

common/effects.cpp
```diff
@@ -809,18 +827,11 @@ int get_building_bonus(const struct city *pcity,
...
-   For instance with EFT_DEFEND_BONUS the attacker's unittype and the
-   defending tile should be passed in.  Slightly counter-intuitive!
-   See doc/README.effects to see how the unittype applies for each effect
-   here.
- */
 int get_unittype_bonus(const struct player *pplayer,
                        const struct tile *ptile,
                        const struct unit_type *punittype,
-                       enum effect_type effect_type)
+                       enum effect_type effect_type,
```
lmoureaux, line 848: This new function would need a docstring

server/rscompat.cpp
```diff
@@ -1075,6 +1075,24 @@ void rscompat_postprocess(struct rscompat_info *info)
   autoadjust_ruleset_data();
 }
 
+/**
+ * Adds <VisionLayer, Main, Local, True> req to all unit vision reqs,
+ * as compat for missing CAP_VUT_VISIONLAYER
+ */
+static bool rscompat_vision_effect_cb(struct effect *peffect, void *data)
+{
+  struct rscompat_info *info __attribute__((unused)) =
```
lmoureaux, line 1084: `__attribute__` is a GNU extension. Use `Q_UNUSED(variable_name);` or the standard `[[maybe_unused]]` instead.

common/effects.cpp
```diff
@@ -715,6 +715,21 @@ int get_city_bonus(const struct city *pcity, enum effect_type effect_type)
...
+int get_city_bonus(const struct city *pcity, enum effect_type effect_type,
+		   enum vision_layer vlayer)
+{
+  if (!initialized) {
+    return 0;
+  }
+
+  return get_target_bonus_effects(NULL, city_owner(pcity), NULL, pcity,
+                                  NULL, city_tile(pcity), NULL, NULL, NULL,
+                                  NULL, NULL, effect_type, vlayer);
```
lmoureaux, line 730: We transitioned to `nullptr` everywhere... I can do this one

lmoureaux (commented):
Still not tested, but went through the code and it looks reasonable. Few stylistic comments.

ec429:
auto clang-format

lmoureaux:
Tested submarines in the `classic` ruleset, appear to work. I consider the new effects experimental until we can play-test them (so behavior changes are still possible).

PR #937, opened 2022-03-06
Support ruleset-defined custom actions in client

### LINKED ISSUE #936: Qt client does not support User Actions
**Describe the bug**
The three ruleset-defined actions (User Action 1-3) can't be performed from the action selection dialog.

**To Reproduce**
1. Load a ruleset with a User Action
2. Attempt to perform the action with a unit that should be able to do it
3. Action does not appear in the dialog and cannot be ordered.

**Expected behavior**
Action and its probability should appear in the dialog, like any other action.

**Additional context**
Needs handlers adding to the `af_map` in `client/gui-qt/dialogs.cpp`.

ec429:
User Action 1 through 3 may be defined by the ruleset, allow ordering
 them through the action selection dialog.
Action Target Kind depends on the ruleset, so dispatch on it at
 runtime to determine the semantics of qv2.

Closes #936.

### DIFF
--- a/client/gui-qt/dialogs.cpp
+++ b/client/gui-qt/dialogs.cpp
@@ -143,6 +143,9 @@ static void transport_alight(QVariant data1, QVariant data2);
 static void transport_unload(QVariant data1, QVariant data2);
 static void keep_moving(QVariant data1, QVariant data2);
 static void pillage_something(QVariant data1, QVariant data2);
+static void user_action_1(QVariant data1, QVariant data2);
+static void user_action_2(QVariant data1, QVariant data2);
+static void user_action_3(QVariant data1, QVariant data2);
 static void action_entry(choice_dialog *cd, action_id act,
                          const struct act_prob *act_probs,
                          const QString custom, QVariant data1,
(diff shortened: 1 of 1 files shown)


client/gui-qt/dialogs.cpp
```diff
@@ -3114,6 +3122,83 @@ static void pillage_something(QVariant data1, QVariant data2)
...
+    break;
+  case ATK_SELF:
+    user_action_unit_vs_self(actor_id, target_id, act_id);
+    break;
+  case ATK_COUNT:
+    fc_assert_msg(ATK_COUNT != action_id_get_target_kind(act_id),
+                  "Bad target kind");
+    break;
+  }
+  return;
```
lmoureaux, line 3184: Isn't the `}` enough?

lmoureaux (commented):
Added to `game.ruleset` in `classic`:
```ini
; Blank ruleset defined user actions.
; See the section "Ruleset defined actions" is doc/README.actions
; Example: set up "User Action 1"
ui_name_user_action_1                = _("%sDisrupt Supply Lines%s")
user_action_1_target_kind            = "tiles"
user_action_1_min_range              = 1
user_action_1_max_range              = 3
user_action_1_actor_consuming_always = FALSE

[actionenabler_user_1]
action = "User Action 1"
```
-> it works.

Some functions are crying for a docstring and I don't want them to feel lonely in the code. Since this is C++, `user_action_N` could have been a template but it's fine as it is.

client/gui-qt/dialogs.cpp
ec429: Yes.

PR #942, opened 2022-03-16
Update paths-ignore in build.yaml

### LINKED ISSUE #629: Don't run GitHub actions for non-code commits
Every commit includes a full compile of the code. That is wasteful, especially for non-code directories such as `doc` or docs`

Make the actions smarter for our code base and only run checks that matter

None. Leaving things as is seems like a waste of time and resources.

**Additional context**
None

jwrober:
Closes #629

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,7 +1,10 @@
 name: Build
 on:
   push:
-    paths-ignore: docs
+    paths-ignore:
+      - 'data/**'
+      - 'dist/**'
+      - 'docs/**'
   pull_request: {}
   schedule:
     - cron: "0 5 * * MON"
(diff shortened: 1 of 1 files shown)


lmoureaux:
Files from `data` and `dist` are installed/used for the installer and I think this should be tested.

lmoureaux:
So what would happen is:

* PR (a) gets merged without testing the Windows installer and introduces a new bug that prevents building it
* `master` isn't rebuilt either since the changes are on the ignore list
* If the next PR with code changes, (b), was branched before (a), it can be merged. Once it is, CI builds start failing on `master`
* If (b) was instead branched after (a), it can't be merged before the bug is solved (in the branch of (b)!)

So we'd need an emergency fix to `master`. I think we really want to avoid this situation...

PR #947, opened 2022-03-19
Remove tolua, replace with sol2 everywhere

zekoz:


lmoureaux (changes requested):
```
/rulesetdir classic
/set revealmap start
/set hut 500
/start
```
Map labels aren't placed and huts do nothing -> signal handlers aren't executed. There in no visible error message

lmoureaux:
Rebased

lmoureaux:
Fixed signals. Basic use cases appear to work now, but I'm still trying to figure out a way to test more rigorously.

common/scriptcore/api_game_methods.cpp
```diff
@@ -40,209 +43,86 @@
...
-                                         Building_Type *pbuilding)
-{
-  LUASCRIPT_CHECK_STATE(L, false);
-  LUASCRIPT_CHECK_SELF(L, pbuilding, false);
+template <> struct unqualified_getter<::QString> {
+  static ::QString get(lua_State *L, int index)
+  {
+    return QString(stack::get<const char *>(L, index));
+  }
+};
```
lmoureaux, line 61: I guess this should go to a header and then we might be able to drop the `toStdString` in the code calling signals.

common/scriptcore/luascript.cpp
```diff
@@ -372,7 +263,14 @@ static void luascript_exec_resource(lua_State *L, const QString &filename)
 
   // We trust that it loads.
   sol::state_view lua(L);
-  lua.script(data.data(), in.fileName().toStdString());
+  auto result = lua.safe_script(data.data(), in.fileName().toStdString());
+  if (!result.valid()) {
+    sol::error err = result;
+    std::cout << "An error (an expected one) occurred: " << err.what()
```
lmoureaux, line 269: Use `qCritical` and friends for error messages like this. A bundled script not loading is quite serious so it should be `qFatal` if you want to check

common/scriptcore/luascript.cpp
```diff
@@ -569,170 +375,49 @@ bool luascript_check_function(struct fc_lua *fcl, const char *funcname)
...
-  if (status) {
-    luascript_report(fcl, status, str);
-  } else {
-    status = luascript_call(fcl, 0, 0, str);
+  sol::state_view lua(fcl->state);
+
+  auto result = lua.safe_script(str);
+  if (!result.valid()) {
+    sol::error err = result;
+    std::cout << "An error (an expected one) occurred: " << err.what()
```
lmoureaux, line 392: Same here, but this one shouldn't be `qFatal`

common/scriptcore/luascript.cpp
```diff
@@ -569,170 +375,49 @@ bool luascript_check_function(struct fc_lua *fcl, const char *funcname)
...
-
-  return stop_emission;
-}
+  auto result = lua.safe_script_file(
+      filename, [](lua_State *, sol::protected_function_result pfr) {
+        // pfr will contain things that went wrong, for either loading or
+        // executing the script the user can do whatever they like here,
+        // including throw. Otherwise...
+        sol::error err = pfr;
+        std::cout << "An error (an expected one) occurred: " << err.what()
```
lmoureaux, line 414: `qCritical` again

lmoureaux (commented):
Had a look at `common/`:

* I see that `sol::state_view` (40 bytes) is passed around by value. Intuitively I would use references, is there a specific reason to use values?
* `Signal` and `ent_handle` would need some docstrings.

More specific comment below

PR #957, opened 2022-03-22
Rewrite loop in research_get_reachable_rreqs

pranavtifr:
The previous implementation wasnt making full use of the capabilites of cpp.

### DIFF
--- a/common/research.cpp
+++ b/common/research.cpp
@@ -393,27 +393,22 @@ static bool research_get_reachable_rreqs(const struct research *presearch,
                                          Tech_type_id tech)
 {
   bv_techs done;
-  std::vector<Tech_type_id> techs;
-  techs.reserve(game.control.num_tech_types);
-  int techs_num;
-  int i;
-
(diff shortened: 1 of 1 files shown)


common/research.cpp
```diff
@@ -422,14 +417,13 @@ static bool research_get_reachable_rreqs(const struct research *presearch,
 
     // Check if required techs are research_reqs reachable.
     for (int req = 0; req < AR_SIZE; req++) {
-      Tech_type_id req_tech = advance_required(techs[i], tech_req(req));
+      // log_debug("iter_tech %d", techs[i]);
```
lmoureaux, line 420: Drop commented code

lmoureaux (commented):
Looking good! Spotted commented out code, otherwise I'd have merged immediately

common/research.cpp
pranavtifr: Ah, Thanks. My bad.

PR #961, opened 2022-03-23
Fixes for ASAN

pranavtifr:
Fix a few alloc-dealloc mismatches, and change all NULL to nullptr in cpp files.


common/aicore/path_finding.cpp
```diff
@@ -3264,7 +3264,6 @@ pf_path_new_to_start_tile(const struct pf_parameter *param)
 void pf_path_destroy(struct pf_path *path)
 {
   if (NULL != path) {
-    delete[] path->positions;
```
lmoureaux, line 3267: This one is needed, it's a raw pointer and `pf_path` doesn't have a destructor.

lmoureaux (commented):
One `free` too much… I trusted you when changing between macros.

There are some `NULL`s hidden in `generate_packets.py` and `generate_specenum.py`.

common/aicore/path_finding.cpp
```diff
@@ -3266,6 +3266,7 @@ pf_path_new_to_start_tile(const struct pf_parameter *param)
 void pf_path_destroy(struct pf_path *path)
 {
   if (nullptr != path) {
+    delete path->positions;
```
lmoureaux, line 3269: This will be an alloc-dealloc mismatch when allocated with `new[]`. Change all allocations to `new[]` using `new[1]` when a single element is needed

pranavtifr: Fixed. I didnt (also 500turns with ASAN) didn't spot anything else.

PR #966, opened 2022-03-25
Begone realloc

pranavtifr:
Changed a few easy reallocs to vectors

### DIFF
--- a/ai/aitraits.cpp
+++ b/ai/aitraits.cpp
@@ -33,8 +33,7 @@ void ai_traits_init(struct player *pplayer)
 {
   enum trait tr;
 
-  pplayer->ai_common.traits = static_cast<ai_trait *>(fc_realloc(
-      pplayer->ai_common.traits, sizeof(struct ai_trait) * TRAIT_COUNT));
+  pplayer->ai_common.traits = std::vector<ai_trait>(TRAIT_COUNT);
 
   for (tr = trait_begin(); tr != trait_end(); tr = trait_next(tr)) {
     int min = pplayer->nation->server.traits[tr].min;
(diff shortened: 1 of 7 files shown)


ai/aitraits.cpp
```diff
@@ -33,8 +33,7 @@ void ai_traits_init(struct player *pplayer)
 {
   enum trait tr;
 
-  pplayer->ai_common.traits = static_cast<ai_trait *>(fc_realloc(
-      pplayer->ai_common.traits, sizeof(struct ai_trait) * TRAIT_COUNT));
+  pplayer->ai_common.traits = std::vector<ai_trait>(TRAIT_COUNT);
```
lmoureaux, line 36: This doesn't seem to be ever resized. `TRAIT_COUNT` is 3 and `ai_trait` is just two `int`s. Would this make more sense as a `std::array<ai_trait, TRAIT_COUNT>`?

client/gui-qt/cityrep.cpp
```diff
@@ -219,24 +219,23 @@ bool city_model::setData(const QModelIndex &index, const QVariant &value,
 QVariant city_model::headerData(int section, Qt::Orientation orientation,
                                 int role) const
 {
-  struct city_report_spec *spec = city_report_specs + section;
+  struct city_report_spec spec = city_report_specs[section];
```
lmoureaux, line 222: Suggestion: `const auto &`

lmoureaux (commented):
Haven't run it yet. Some `city_report_specs` code could be simplified further by declaring variables inside loops and using `auto`, but the reduction in manual memory management is already a significant improvement.

ai/aitraits.cpp
pranavtifr: I was considering the same, also we have waay too much ram these days to have a lot more traits still in stack,  Ended up doing the same in heap because I thought the difference would be negligible, and array didnt have clear.

Issue #958, opened 2022-03-22
Memory leaks in client 


pranavtifr:
**Describe the bug**
We have a bunch of memory leaks in client found using `-fsanitize=leak` which we need to fix.

**To Reproduce**
- Compile using `-fsanitize=leak` flag. 
- Let an automated game run. 
Settings for the automated game
```
set gamelog 30
set aifill 4
set endt 500
set timeout 1
set autosave "INTERRUPT" 
set minp 0
set gameseed 42
set mapseed 42
start
```
- Connect using a client and observe and then exit the client.

**Expected behavior**
Ideally, we should not have any leaks. 

**Platform and version (please complete the following information):**
 - OS:Linux
 - Freeciv21 version: v3.0-beta.1
 - Ruleset/Longturn game (if applicable): classic

**Additional context**
Information on the leaks. See attached log file. 
[log.txt](https://github.com/longturn/freeciv21/files/8326722/log.txt)

lmoureaux:
Using this issue to document overall progress.

# Improved

I don't think I caught all 92 leaks but at least 4 per tileset load.
```
Direct leak of 2944 byte(s) in 92 object(s) allocated from:
    #0 0x7f82589e1811 in operator new(unsigned long) /usr/src/debug/gcc/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x5638c187106d in crop_sprite(QPixmap const*, int, int, int, int, QPixmap const*, int, int) /home/pranav/freeciv/freeciv21/client/gui-qt/sprite.cpp:76
    #2 0x5638c256fdaf  (/home/pranav/freeciv/freeciv21/build/freeciv21-client+0x43a6daf)

Indirect leak of 590968 byte(s) in 94 object(s) allocated from:
    #0 0x7f82589dfdd9 in __interceptor_malloc /usr/src/debug/gcc/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f8257889d87 in QImageData::create(QSize const&, QImage::Format) (/usr/lib/libQt5Gui.so.5+0x184d87)

Indirect leak of 144992 byte(s) in 92 object(s) allocated from:
    #0 0x7f82589e1811 in operator new(unsigned long) /usr/src/debug/gcc/libsanitizer/asan/asan_new_delete.cpp:99
[...]

PR #970, opened 2022-03-29
Fix some memory leaks

lmoureaux:
See #958.


lmoureaux:
You can merge PRs now

PR #971, opened 2022-03-30
Support windows clang and fixes

NIKEA-SOFT:


server/server.cpp
```diff
@@ -31,9 +31,12 @@
 // Stuff to wait for input on stdin.
 #ifdef Q_OS_WIN
 #include <QMutexLocker>
+#include <io.h>
 #else
 #include <QtCore/QSocketNotifier>
+#include <unistd.h>
```
lmoureaux, line 37: Why do we need `unistd` on Linux and Mac?

NIKEA-SOFT: > Why do we need `unistd` on Linux and Mac?

https://github.com/longturn/freeciv21/blob/f779ae6b04739d613b5077c47aebc550f9657fc1/server/server.cpp#L276

lmoureaux: Thanks :) It's obviously already included transitively but better do it explicitly indeed.

.github/workflows/build.yaml
```diff
@@ -120,7 +120,48 @@ jobs:
...
             build/Windows-x86_64/Freeciv21-*-Windows-x86_64.exe
             build/Windows-x86_64/Freeciv21-*-Windows-x86_64.exe.sha256
-
+  windows_clang_msvc:
+    name: "Windows Clang"
+    runs-on: windows-latest
+    needs: clang-format
+    strategy:
+      matrix:
+        architecture: [Win32, x64]
```
jwrober, line 129: Do we need both architectures here? Why not just x86?

ai/default/aitools.cpp
```diff
@@ -743,7 +743,6 @@ void dai_unit_new_task(struct ai_type *ait, struct unit *punit,
       if (unit_owner(missile) == unit_owner(punit)
           && def_ai_unit_data(missile, ait)->task != AIUNIT_ESCORT
           && !unit_transported(missile)
-          && unit_owner(missile) == unit_owner(punit)
```
jwrober, line 746: This does not seem to be related to the build tools update

lmoureaux: It's the same condition as in the first line

PR #973, opened 2022-03-30
Refactor/path finding class

pranavtifr:
Make Pf_path into a class

- Add many constructors for situations which the path is created to be immediately destroyed
- Overload the subscript operator [] to accept negative values. Negative values count backwards from end of array
- positions arent directly exposed and you reach positions using the overloaded [] operator
- move memory management to under the hood in as many places as possible
- algos which return nullptrs instead return an empty path. Checks now use, path.empty() to differentiate from other nullptrs.
- All functions which dont modify the path, now get the path by value. For functions which modify the path the pointer is passed. 
(this helps us differentiate both functions at the point of calling the functions)

### DIFF
--- a/ai/classic/classicai.cpp
+++ b/ai/classic/classicai.cpp
@@ -39,6 +39,7 @@
 
 #include "classicai.h"
 
+class Pf_path;
 /* This function is only needed as symbol in module. Nevertheless
    gcc requires that there is prior prototype. */
 const char *fc_ai_classic_capstr();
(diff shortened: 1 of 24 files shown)


pranavtifr:
Regarding CodeFactor issue: I'll leave the refactoring of the `pf_danger_map_construct_path` function  to reduce complexity for later. It is quite annoying and involves touching quite a bit of AI code. 

Regarding Codacy issue: I am not sure what to do. Casting as void is needed if you want to explicitly print the pointer, but it is more error prone if we dont use const in the << operator. 

I'll just ask for review as is. Since both the issues arent that important and I would like a review of the class first.

pranavtifr:
Remember to not connect the client midway differently when observing, I have noticed that sometimes changes the results. Not really sure why.

lmoureaux:
Seems like I won't find the motivation to test this thoroughly. Merging.

PR #975, opened 2022-03-31
Package Release code (no debugging symbols)

### LINKED ISSUE #954: Standardize what is included in the install packages
For the Windows package we compile "release" code and package it up into a 51 MB archive. For Linux we compile with "RelwithDebInfo" and package it up into a 215 MB archive. For MacOS we do the same as Windows and package it up into a 101 MB archive

All supplied packages should do the same thing. Would suggest to go with no-debugging symbols in released code. The size of the executables are very large.

None

**Additional context**
None

jwrober:
Closes #954

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -67,14 +67,14 @@ jobs:
         if: |
           startsWith(github.ref, 'refs/tags/')
             && matrix.compiler == 'gcc'
-            && matrix.build_type == 'RelWithDebInfo'
+            && matrix.build_type == ''
         with:
           files: |
             build/Linux-x86_64/freeciv21_*_amd64.deb
             build/Linux-x86_64/freeciv21_*_amd64.deb.sha256
 
   windows:
-    name: "Windows"
+    name: "Windows MSYS2 (gcc)"
     runs-on: windows-latest
     needs: clang-format
     defaults:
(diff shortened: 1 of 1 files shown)


lmoureaux:
`Release` code would be `CMAKE_BUILD_TYPE=Release` (i.e., an optimized build). I suggest to either rephrase the commit message or add an optimized build. We could also `strip` debug symbols out of the binaries and save them somewhere -- though I'm not sure how Windows loads debug symbols at runtime.

jwrober:
Linux is/was the only platform packaging using `RelWithDebInfo`, Windows MSYS2 and Mac OS were doing `Release` for the build type. So all this does is keep all of them the same for the package.

lmoureaux:
`Release` is a specific CMake build type that enables compiler optimization (`-O3`). By default or with `CMAKE_BUILD_TYPE=""`, CMake pulls its configuration from the environment. If nothing is found in the environment then no flags are passed to the compiler and it's an unoptimized build (`-O0`). A Linux distro would set build flags roughly equivalent to `RelWithDebInfo` (`-O2 -g`) and later split out the debug symbols into a `-dbg` package.

Mac builds use `Release`, which is a sensible thing to do for production level code (even though it makes debugging crashes very difficult): https://github.com/longturn/freeciv21/blob/cc96a43adb2d6d5c5704bd326ebf2b418829a29c/CMakePresets.json#L16
I think your switches Debian/Ubuntu to the default CMake build type (which is different from `Release`). To avoid confusion with the CMake `Release` mode, I would avoid using this name in the PR and commit message.

lmoureaux (approved):
Let's squash the two commits

PR #987, opened 2022-04-22
Refactoring and fixes

NIKEA-SOFT:
List of changes:
1) Change CMake files for support Visual Studio and debug definition for multi-configurations
2) Remove unused/deprecated files
3) Remove GUI_FUNC_PROTO wrapper with callbacks
4) Fix incorrect delete


lmoureaux:
There was a reason not to use GLOB in the CMakeLists files. The documentation states:

> **Note**  We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

NIKEA-SOFT:
> 

I am aware, but GLOB_RECURSE cannot be used, and enumerating each file manually is the same as using GLOB

lmoureaux:
I prefer the first method (for the reasons explained in the note; you don't even specify `CONFIGURE_DEPENDS` so currently it's quite heavily broken with most generators). The duplication between `foo.h` and `foo.cpp` can be resolved with a bit of `string(REPLACE)` magic in a function.

lmoureaux:
The `clang-format` error is imported from `master`: #988 (you don't need to merge `master` unless there is a merge conflict, in which case I prefer rebase).

Please squash [fc174fe](https://github.com/longturn/freeciv21/pull/987/commits/fc174fed153aa8bff18351c788ef2386c60e8904) into [4360b65](https://github.com/longturn/freeciv21/pull/987/commits/4360b65e4ada6bc52f041a85537e3c3470380d46) to clean up the history.

NIKEA-SOFT:
> The `clang-format` error is imported from `master`: #988 (you don't need to merge `master` unless there is a merge conflict, in which case I prefer rebase).
> 
> Please squash [fc174fe](https://github.com/longturn/freeciv21/pull/987/commits/fc174fed153aa8bff18351c788ef2386c60e8904) into [4360b65](https://github.com/longturn/freeciv21/pull/987/commits/4360b65e4ada6bc52f041a85537e3c3470380d46) to clean up the history.

I had to perform a rebase to move these two commits closer to each other, and then perform a second rebase to merge the commits. I've finished making corrections.

PR #995, opened 2022-05-01
Add move constructor and move assignment for PFPath

### LINKED ISSUE #989: Coverity complaining about PFPath
**Describe the bug**

```
** CID 1488326:  Low impact quality  (MISSING_MOVE_ASSIGNMENT)
/common/aicore/path_finding.h: 319 in ()


________________________________________________________________________________________________________
*** CID 1488326:  Low impact quality  (MISSING_MOVE_ASSIGNMENT)
/common/aicore/path_finding.h: 319 in ()
313     
314       enum direction8 dir_to_next_pos; // Used only in 'PFPath'.
315       enum direction8 dir_to_here;     // Where did we come from.
316     };
317     
318     // Full specification of a path.
>>> CID 1488326:  Low impact quality  (MISSING_MOVE_ASSIGNMENT)
>>> Class "PFPath" may benefit from adding a move assignment operator. See other events which show the copy assignment operator being applied to rvalues, where a move assignment may be faster.
319     class PFPath {
320       std::vector<pf_position> positions;
321     
322     public:
323       PFPath();
324       PFPath(int size);
```

pranavtifr:
It improves performance in places a default copy constructor is used instead (Fixes #989)

Additionally, Remove the implementation of some constructors and use the default keyword instead

### DIFF
--- a/ai/default/aiferry.cpp
+++ b/ai/default/aiferry.cpp
@@ -675,7 +675,7 @@ bool dai_amphibious_goto_constrained(struct ai_type *ait, struct unit *ferry,
        * has run out of movement points */
       struct tile *next_tile;
 
-      if (!path.pf_path_advance(unit_tile(passenger))) {
+      if (!path.advance(unit_tile(passenger))) {
         /* Somehow we got thrown away from our route.
          * This can happen if our movement caused alliance breakup. */
         return unit_is_alive(pass_id);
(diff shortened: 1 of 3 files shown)


lmoureaux:
Looks good to me, though I still need to test. Out of curiosity, where is the move constructor used?

pranavtifr:
As far I know, nowhere, I havent used the copy constructor explicitly anywhere either, With the move assignment operator, the places which by default used copy constructor wont use it any longer. It is just good practice to include it because, incase in the future we have classes with paths as members, then the default move constructor there will look for a move constructor here (which all just eventually boils down to the move constructor of std::vector for pf_positions)

PR #997, opened 2022-05-01
Add the autogame serv file for testing

pranavtifr:


lmoureaux:
```
Option 'gamelog' not recognized.
```

lmoureaux:
Autogame went on for 500 turns as expected. No player died but there was a civil war, spaceship/planes/paratroopers probably not tested. No leaks in server, 1MB in client.

PR #1000, opened 2022-05-03
Make the gold indicator widget more informative

### LINKED ISSUE #983: Polish top bar gold indicator
The gold indicator in the top bar needs to be polished. Clicking on the widget gives access to the economics report.

In most cases, the widget should show the following:
* Gold
* Income

For global observers, it should show something else. I'm not even sure there is a working economics report in that case.

Proposal for the contents:
image

From top to bottom:
* Positive income
* When space constrained, amounts can collapse using SI suffixes
* Negative income
* Low on funds
* Global observer

Left click on the widget toggles the economics report in full screen. Right click does nothing (or open the taxes dialog? open for discussion).

**Additional context**
#940

lmoureaux:
The top bar widget that shows the current treasury and gold income gains two features:

* It turns red when the income is negative.
* Its outline turns red when low on funds.

This should help newbies figure out when something is wrong.

Closes #983.
Changes with respect to the proposed solution in the issue:

* The whole text turns red when income is negative (for a technical reason), and
* The global observer state is not implemented because the economy screen doesn't exist in that mode.

Only Nightstalker is implemented pending UI comments.

Screenshots of the new states:

image
image

### DIFF
--- a/client/gui-qt/page_game.cpp
+++ b/client/gui-qt/page_game.cpp
@@ -227,9 +227,11 @@ void pageGame::updateInfoLabelTimeout()
   if (client.conn.playing != nullptr) {
     sw_economy->set_gold(client.conn.playing->economic.gold);
     sw_economy->set_income(player_get_expected_income(client.conn.playing));
+    sw_economy->setEnabled(true);
   } else {
     sw_economy->set_gold(0);
     sw_economy->set_income(0);
+    sw_economy->setEnabled(false);
   }
 
   sw_indicators->update();
--- a/client/gui-qt/page_game.cpp
+++ b/client/gui-qt/page_game.cpp
@@ -97,8 +97,7 @@ pageGame::pageGame(QWidget *parent)
   sw_science->setIcon(
       fcIcons::instance()->getIcon(QStringLiteral("research")));
   sw_science->setCheckable(true);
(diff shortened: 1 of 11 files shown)


jwrober:
Is it possible to have the red outline "pulse" to help bring your eye to it? Red is hard on color blind folks, but that is ok.

lmoureaux:
> Is it possible to have the red outline "pulse" to help bring your eye to it? 

A custom style could do that, Qt stylesheets are too limited. I'm not sure a blinking widget in peripheral vision is a good idea, it would be too annoying.

> Red is hard on color blind folks, but that is ok.

It's not the only way to convey the information, though the minus sign occupies much less screen area. The contrast ratio with the background does fail Web accessibility standards... Color could be tweaked for better contrast.

jwrober:
If we can go with a more contrasting color that meets web accessibility standards that would be ideal.

lmoureaux:
The problem is actually the background, it's too bright to provide good contrast with anything but super washed out colors. Changing it would be going way overboard (NightStalker needs a complete revision if we want to meet the standards).

lmoureaux:
Implemented other themes. Didn't spend much time picking the perfect color, but tried to use something legible.

jwrober (approved):

PR #999, opened 2022-05-02
Document Visual Studio

jwrober:
Closes #976

PR #1005, opened 2022-05-05
Refactoring code and fixes: cmake, crashes and codecy

NIKEA-SOFT:


common/aicore/cm.h
```diff
@@ -22,6 +22,22 @@ struct cm_parameter {
 
   int factor[O_LAST];
   int happy_factor;
+
+  // Copy assignment operator
```
lmoureaux, line 26: Docstrings should start with three slashes (`///`) or `/**` because that how Doxygen "sees" them (this is very common for documentation generators, with a long history dating back to at least 1995 and the first release of Java).

common/aicore/path_finding.cpp
```diff
@@ -95,8 +95,14 @@ struct pf_map {
...
 
+template <typename T> struct pf_map_trait {
+  using Type = std::conditional_t<std::is_const_v<std::remove_pointer_t<T>>,
+                                  const pf_map, pf_map>;
+};
+
 // Down-cast macro.
-#define PF_MAP(pfm) ((struct pf_map *) (pfm))
+#define PF_MAP(pfm)                                                         \
+  (reinterpret_cast<typename pf_map_trait<decltype(pfm)>::Type *>(pfm))
```
lmoureaux, line 105: 1. Thinks macros are ugly
2. Hides templates in a macro

Nice solution to the const-correctness problem! Would have skipped the struct and written `using as_pf_map_t = ...` directly, but it works.

lmoureaux (commented):
Very easy to review! It's clear that the second commit changes nothing, which let me focus on the rest of the code. I have two minor comments on the last commit, I leave them to you (i.e. do as you prefer).

Apparently we see in #999 that adding the header files to `CMakeLists` isn't necessary for VS 2022. I think we can still merge this PR while we figure out whether this is indeed the case, since the change could be reverted without breaking the blame too much (I would prefer to have only the `.cpp` if possible, since that's how CMake is intended to work and it's less verbose).

common/aicore/path_finding.cpp
lmoureaux: Thinking about it now, this would be a slightly simpler solution (if I understood the problem correctly):
```cpp
/**
 * Casts to `pf_map`
 *
 * \fixme Capitalized because this used to be a macro.
 */
pf_map *PF_MAP(void *x) { return reinterpret_cast<pf_map *>(x); }

/**
 * Casts to `pf_map`
 *
 * \fixme Capitalized because this used to be a macro.
 */
const pf_map *PF_MAP(const void *x) { return reinterpret_cast<const pf_map *>(x) }
```

jwrober:
When I run a build of this PR I am seeing a lot more warnings that from what we see on master branch right now.
```
common/packets_gen.cpp:8284:48: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_info’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
common/packets_gen.cpp:8766:32: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_info’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
common/packets_gen.cpp:12111:48: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_manager’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
common/packets_gen.cpp:12161:32: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_manager’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
../client/governor.cpp:439:46: warning: ‘void* memcpy(void*, const void*, size_t)’ writing to an object of type ‘struct cm_parameter’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
../server/citytools.cpp:2609:66: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct cm_parameter’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
../client/packhand.cpp:170:66: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct player’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
```

Not sure if this is an issue or not, just making mention of it.

NIKEA-SOFT:
> When I run a build of this PR I am seeing a lot more warnings that from what we see on master branch right now.
> 
> ```
> common/packets_gen.cpp:8284:48: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_info’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
> common/packets_gen.cpp:8766:32: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_info’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
> common/packets_gen.cpp:12111:48: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_manager’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
> common/packets_gen.cpp:12161:32: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct packet_city_manager’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
> ../client/governor.cpp:439:46: warning: ‘void* memcpy(void*, const void*, size_t)’ writing to an object of type ‘struct cm_parameter’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
> ../server/citytools.cpp:2609:66: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct cm_parameter’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
> ../client/packhand.cpp:170:66: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct player’ with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
> ```
> 
> Not sure if this is an issue or not, just making mention of it.

Thank you for writing about this. 
The problem is that the copy assignment operator defined by me changed the structure and it ceased to be a POD.
Fixed.

PR #1008, opened 2022-05-09
Don't include headers in CMakeLists.txt after all

lmoureaux:
The headers were added because Microsoft Visual Studio's "Solution" view
doesn't show the headers.  However, VS' built-in CMake handling doesn't use the
solution view and is based on folders (much like CMake itself).  There is no
recommendation from the CMake side on how to handle this problem with CMake's
built-in generators for Visual Studio.

Following the VS documentation and the instructions added in #999, one can get
(at least) decent VS support with IntelliSense working.  This doesn't require
keeping the headers in the CMakeLists, so remove them for more idiomatic CMake
code.

This reverts part of commit 6e38a15ac21b75dc616c0c30cec760d026f2598a.
See #1005.

### DIFF
--- a/ai/CMakeLists.txt
+++ b/ai/CMakeLists.txt
@@ -2,12 +2,9 @@ add_library(
   ai
   STATIC
   aitraits.cpp
-  aitraits.h
   classic
   difficulty.cpp
-  difficulty.h
   handicaps.cpp
-  handicaps.h
 )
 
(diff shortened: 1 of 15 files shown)


jwrober (approved):
I see not change in behavior with or without the `.h` files included. I'm glad they are being removed.

PR #1011, opened 2022-05-10
Fix bugs found when porting dino's ruleset

lmoureaux:
See commit messages.


lmoureaux:
MWE for the second commit (`script.lua`):
```lua
function crash()
  for tile in whole_map_iterate() do
    if tile:has_extra("Wheat") then
      tile:remove_extra("Wheat")
    end
  end
end

signal.connect("map_generated", "crash")
```

jwrober (approved):
Works for me

PR #1017, opened 2022-05-14
Support target install in Visual Studio

### LINKED ISSUE #1006: Install target fails on Visual Studio on Windows
**Describe the bug**
The CMake install target for Windows is built for MSYS2 and has a hard coded path in it for SSL DLLs based on the MSYS2 path and not Visual Studio. When invoking `--target install` then fails when using Visual Studio.

**To Reproduce**
1. Install Visual Studio on Windows
2. Run configure and build commands
3. Run the command to install

**Expected behavior**
We will not support CPack and build a package installer due to licensing issues, however the install target should at least combine everything together for debugging and testing purposes.

**Screenshots**
N/A

 - OS: Windows
 - Freeciv21 version: 3.0-Beta1

**Additional context**
None

jwrober:
Closes #1006

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ set(FREECIV21_VERSION "${FREECIV21_VERSION_MAJOR}.${FREECIV21_VERSION_MINOR}.${F
 # Set project
 project(freeciv21 VERSION ${FREECIV21_VERSION} LANGUAGES C CXX)
 
-# Gather all the things we need for Windows builds early.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -227,7 +226,34 @@ if(WIN32 OR MSYS OR MINGW)
           list(REMOVE_AT _files 0 1)
       endwhile()
     ]] COMPONENT freeciv21)
+  elseif(WIN32)
```
lmoureaux, line 229: I think this block is indented one step too far to the right

PR #1018, opened 2022-05-14
Add Ubuntu 22.04 (Jammy) to CI builds

lmoureaux:
The main development target is still 20.04, but testing on 22.04 as well is a
first step towards a complete switch.  I don't expect any breakage since some
of us have been using Arch for development (-> newer versions of everything).

Include a single configuration for now, building with GCC in debug mode to
avoid using too many free Github CPU cycles.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,15 +11,20 @@ on:
 jobs:
   ubuntu:
     name: "Ubuntu"
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


lmoureaux:
Let's see if it builds :sweat_smile:

lmoureaux:
...and it did not! Hopefully on the second attempt...

jwrober (approved):
Looks good to me

PR #1023, opened 2022-05-27
Fix: server does not terminate the process

NIKEA-SOFT:
Server may not shutdown the process, it gets stuck in a loop, waiting for input to the console.

### DIFF
--- a/server/server.cpp
+++ b/server/server.cpp
@@ -222,9 +222,14 @@ std::pair<QTcpServer *, bool> srv_prepare()
  */
 detail::async_readline_wrapper::async_readline_wrapper(bool interactive,
                                                        QObject *parent)
-    : QThread(parent), m_interactive(interactive)
+    : QThread(parent), m_interactive(interactive), m_stop{}
 {
 }
+
+/**
+ * Set stop flag for thread
+ */
+void detail::async_readline_wrapper::stop() { m_stop = true; }
 
 /**
  * Blocks until a line of input can be read from stdin, then emits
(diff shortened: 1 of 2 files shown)


server/server.cpp
```diff
@@ -354,8 +364,11 @@ server::~server()
     auto notifier =
         qobject_cast<detail::async_readline_wrapper *>(m_stdin_notifier);
     if (notifier) {
+      notifier->stop();
+      notifier->quit();
```
lmoureaux, line 368: `quit()` does nothing because the thread isn't running an event loop

server/server.h
```diff
@@ -21,6 +21,7 @@
 
 // Qt
 #include <QObject>
+#include <atomic>
```
lmoureaux, line 24: I don't think `<atomic>` is part of Qt... can we put an empty line between them?

server/server.cpp
lmoureaux: oh yes, my bad, it does run an event loop

server/server.h
NIKEA-SOFT: > I don't think `<atomic>` is part of Qt... can we put an empty line between them?
Why, if std::atomic is used there, we do not need to guess whether atomic is included somewhere or not

lmoureaux: I'm simply suggesting this:
```C++
// Qt
#include <QObject>

#include <atomic>
```
So if in the future we include more of Qt or the STL, it doesn't become inconsistent.

NIKEA-SOFT: Done

lmoureaux:
Requesting a test from @jwrober -- start a local game, end it, the server should be gone

jwrober (approved):

Issue #1001, opened 2022-05-03
Add a button to show/hide the messages widget


lmoureaux:
**Is your feature request related to a problem? Please describe.**

The messages widget is currently always shown. It can be "folded", but the title bar keeps using valuable space on the map.

**Describe the solution you'd like**

I would like the messages widget to be togglable with a small button integrated in the top bar. The right edge looks like a good location:
![image](https://user-images.githubusercontent.com/22327575/166586427-876c1669-ca40-4eca-bcfe-4545fd399a2c.png)

The button should be icon-only with a colorized SVG icon. The icon could represent a bell. A tooltip should be included. A keyboard shortcut could also be added.

The button should have two modes: toggled and not toggled (Qt buttons support this natively). When toggled, the messages widget should be visible; otherwise, it should remain hidden. The messages widget should appear immediately under the button, like a dropdown menu:
![image](https://user-images.githubusercontent.com/22327575/166586835-4b177646-831c-4978-afb0-7cd2c5c75800.png)
Moving it should be forbidden, but resizing can be kept (though the active sides should be changed to left and bottom). Care should be taken to keep it in the corner when the window is resized.

When new messages appear and the messages panel is closed, the button should change its icon to include a small colored circle. The icon should come back to normal when the panel is opened.

**Describe alternatives you've considered**

* Merging the chat and messages streams to free up space (#952). Reception on Discord was poor, but it could still be done after this PR.
* A system of "bubbles" was mentioned. While I like it in principle, currently it's still very abstract. Its implementation would certainly need a lot more work than the solution proposed here.
* Always making the panel full height. I'm still considering it actually, but keeping the current form factor sounds easier as a first step.

**Additional context**

Want to free up space at the top left for more important things.

lmoureaux:
Done in #1026

PR #1026, opened 2022-06-11
Add a button to show/hide the messages widget #1001

NIKEA-SOFT:
Task from #1001


lmoureaux:
Simple change that improves the UI, I like it! Few comments:

* Don't worry about the wasm build, it's not your fault
* I can still move the messages using the small arrow at the top left. I think we should remove it.
* The expand/collapse button in the title bar duplicates the new show/hide functionality in the top bar. I would also remove it.
* The new icons won't follow the color scheme because they are in png format. With svg icons [this code](https://github.com/longturn/freeciv21/blob/0e679bcce2f538dc024c5a38cc405e1d5cc23c63/client/icons.cpp#L85-L88) will automatically adjust the color (it's quite dumb: it replaces `#000` everywhere in the file with another color). This way the icons can match the theme.
* Speaking of icons, we should add `.license` files for the new files (look at `data/themes/gui-qt/icons` for examples)
* Could we have a red dot on the `notify` icon? I think it would make it more distinct

PR #1027, opened 2022-06-12
Reset tileset data when the ruleset changes

### LINKED ISSUE #974: Client renders terrain specials that aren't there
**Describe the bug**
If there are more than a certain number of extras defined in the ruleset, the client renders some of them with a special overlaid.  In the case of the current Aviation ruleset, this consists of Gold on rails and Iron on rivers.

**To Reproduce**
Download the latest Aviation ruleset from http://jttlov.no-ip.org/fc-aviation-latest/aviation.json
Start a game.
Observe the terrain.  There will be iron bars drawn on all tiles with rivers, despite these tiles not having the Iron special.

**Expected behavior**
Specials should only appear on tiles that actually have them.

**Screenshots**
fcgold
fcgoldbug
Example of a tile with rails, showing nonexistent gold.

 - OS: Linux (Debian testing)
 - Freeciv21 version: `e9f5b020e9`
 - Ruleset/Longturn game (if applicable): Aviation `8afef23`

**Additional Context**
The bug first appeared when I added the Spotted extras to `terrain.ruleset` and `aviation.tilespec`.

lmoureaux:
Some cached data needs to be updated when the ruleset changes and the tileset
remains the same.  Add a new virtual method in the layer class to trigger this
reset.

Fixes #974.

### DIFF
--- a/client/layer.h
+++ b/client/layer.h
@@ -173,6 +173,11 @@ public:
     Q_UNUSED(terrain);
   }
 
+  /**
+   * Resets cached data that depends on the ruleset.
+   */
+  virtual void reset_ruleset() {}
+
   mapview_layer type() const { return m_layer; }
 
 protected:
--- a/client/layer_special.cpp
+++ b/client/layer_special.cpp
@@ -80,4 +80,13 @@ std::vector<drawn_sprite> layer_special::fill_sprite_array(
   return sprites;
 }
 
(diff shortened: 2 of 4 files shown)


lmoureaux:
@ec429

jwrober (approved):
It seems to resolve the issue as described for me.

PR #1032, opened 2022-06-17
Don't auto-return units to refuel if they're converting

ec429:
A converting unit may (e.g. Submersibles in Aviation ruleset) save
 itself from fuel exhaustion by conversion.  In that case, there is
 no need to override the unit's existing orders at turn-end and
 auto-move it to a nearby refueling point (city, carrier, etc.).
So, condition that auto-move code on !converting_fuel_rescue().


pranavtifr:
Merging since it is  a very small and reasonable change.

PR #1039, opened 2022-06-23
Remove data/misc/cityicon.png

### LINKED ISSUE #985: Drop the old city dialog icon

We have an icon for the city dialog, located at `data/misc/cityicon.png`. It is no longer in use.

Delete it.

Keep it in case we want to make the city screen windowed again. We'd probably not use it anyway.

**Additional context**

#978 #984

jwrober:
Closes #985


lmoureaux:
When I wrote "no longer used", it may have been an overstatement. I see that it's still referenced from `misc/icons.spec`. `tileset.cpp` has an enumerator named `ICON_CITYDLG` that I bet must still load it.

jwrober:
> When I wrote "no longer used", it may have been an overstatement. I see that it's still referenced from `misc/icons.spec`. `tileset.cpp` has an enumerator named `ICON_CITYDLG` that I bet must still load it.

Ok so maybe leave the file alone? I'm not going to touch the code.

lmoureaux:
> Ok so maybe leave the file alone? I'm not going to touch the code.

I'm trying to push you into it :joy: I can do the coding part, I'll push to your branch (assuming you kept the "allow edits by maintainers" box checked)

jwrober:
I'll look at the `misc/icons.spec` and you can do the `tileset.cpp` fix

lmoureaux:
Did the C++ changes and rebased on top of `master` so the actions can succeed.

PR #1041, opened 2022-06-23
Fix running the wasm CI

lmoureaux:
The CI started to fail because of an outdated cache after the latest release of
emsdk. Update the cache key to pull a fresh cache.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -327,7 +327,7 @@ jobs:
         name: Checkout extra-cmake-modules
         with:
           repository: KDE/extra-cmake-modules
-          ref: v5.89.0
+          ref: v5.96.0
           path: ecm
       - name: Install ecm
         run: |
@@ -338,7 +338,7 @@ jobs:
         name: Checkout karchive
         with:
           repository: KDE/karchive
-          ref: v5.89.0
+          ref: v5.96.0
           path: karchive
       - name: Install karchive
         run: |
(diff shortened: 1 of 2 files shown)


lmoureaux:
It doesn't seem to like the latest KArchive. @zekoz is this easy for you to fix?

jwrober:
@lmoureaux - can you rebase this to master and see what happens?

lmoureaux:
Rebased, let's see

lmoureaux:
```

CMake Error at emsdk/upstream/emscripten/cache/sysroot/lib/cmake/KF5Archive/KF5ArchiveTargets.cmake:56 (set_target_properties):
-- Generating done
  The link interface of target "KF5::Archive" contains:
    PkgConfig::LibZstd
  but the target was not found.  Possible reasons include:
    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
Call Stack (most recent call first):
  emsdk/upstream/emscripten/cache/sysroot/lib/cmake/KF5Archive/KF5ArchiveConfig.cmake:55 (include)
  cmake/FreecivDependencies.cmake:121 (find_package)
  CMakeLists.txt:72 (include)
```

jwrober:
I suggest we disable WASM CI until we can get zekoz help to determine what is going on with KF5Archive

lmoureaux:
> I suggest we disable WASM CI until we can get zekoz help to determine what is going on with KF5Archive

Even better, we can fix it ourselves!

PR #1060, opened 2022-06-26
Fix: Optimize code in ASan build

pranavtifr:
Fixes warning, _FORTIFY_SOURCE requires compiling with optimization
Also change the build type to RelWithDebInfo

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -44,9 +44,9 @@
       "generator": "Ninja",
       "binaryDir": "${sourceDir}/build",
       "cacheVariables": {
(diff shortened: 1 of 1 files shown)


lmoureaux:
Why are the C and CXX flags so different btw?

lmoureaux (approved):

pranavtifr:
> Why are the C and CXX flags so different btw?

Idk. These are the default flags added by `makepkg` in `/etc/makepkg.conf`. I just took the ones which look important and skipped stuff like `--arch`

PR #1067, opened 2022-06-26
Bugfix/remove unused functions

pranavtifr:
And also unnecessary parameters

### DIFF
--- a/common/aicore/caravan.cpp
+++ b/common/aicore/caravan.cpp
@@ -84,8 +84,6 @@ static void caravan_result_init(struct caravan_result *result,
 
   result->value = 0;
   result->help_wonder = false;
-  if ((src != nullptr) && (dest != nullptr)) {
-  }
 }
 
 /**
--- a/common/aicore/caravan.cpp
+++ b/common/aicore/caravan.cpp
@@ -60,30 +60,6 @@ void caravan_parameter_init_from_unit(struct caravan_parameter *parameter,
 }
 
 /**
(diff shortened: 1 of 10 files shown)


common/aicore/caravan.cpp
```diff
@@ -110,10 +85,6 @@ static void caravan_result_init(struct caravan_result *result,
   result->value = 0;
   result->help_wonder = false;
   if ((src != nullptr) && (dest != nullptr)) {
-    result->required_boat =
-        tile_continent(src->tile) != tile_continent(dest->tile);
-  } else {
-    result->required_boat = false;
   }
```
lmoureaux, line 88: Empty `if`?

common/aicore/path_finding.cpp
```diff
@@ -3439,28 +3380,6 @@ inline bool operator==(const pf_parameter &e1, const pf_parameter &e2)
   return true;
 }
 
-inline uint qHash(const pf_parameter &key, uint seed)
```
lmoureaux, line 3442: Note to self: check that it's really not used

lmoureaux: It was added when the hash in `pf_reverse_map` was turned to `QHash`. This hasn't changed since. Maybe it's used indirectly in template instantiations?

lmoureaux: Looks like `-O2` also removes it, so it's really unused.

PR #1078, opened 2022-06-28
Remove function  player_slot_count()

pranavtifr:
It always returns a constant, 

Inspired from https://github.com/freeciv/freeciv/commit/62fb8c8f66baf67bbd7eebcb03982ffc5c024702, why call it once when you dont have to call it at all.


lmoureaux (approved):
Yeah, cleaning up a 16-years-old function!

PR #1072, opened 2022-06-27
Draft: Bugfix/remove unused functions2

pranavtifr:
I'll rebase after #1067  is merged.

PR #1079, opened 2022-06-29
Bugfix/remove unused functions3

pranavtifr:
I'll rebase after merging #1072

### DIFF
--- a/server/fcdb.cpp
+++ b/server/fcdb.cpp
@@ -29,27 +29,20 @@ _   ._       Copyright (c) 1996-2021 Freeciv21 and Freeciv contributors.
 
 #include "fcdb.h"
 
-enum fcdb_option_source {
-  AOS_FILE, // Read from config file
-};
-
 struct fcdb_option {
-  enum fcdb_option_source source;
   char *value;
 };
 
 QHash<QString, fcdb_option *> fcdb_config;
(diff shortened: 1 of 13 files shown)


server/fcdb.cpp
```diff
@@ -30,9 +30,7 @@ _   ._       Copyright (c) 1996-2021 Freeciv21 and Freeciv contributors.
 #include "fcdb.h"
 
 enum fcdb_option_source {
-  AOS_DEFAULT, // Internal default, currently not used
-  AOS_FILE,    // Read from config file
-  AOS_SET      // Set, currently not used
+  AOS_FILE, // Read from config file
```
lmoureaux, line 33: Has this enum just become useless? Can we remove its usage?

server/handchat.h
```diff
@@ -12,4 +12,5 @@
 
 struct connection;
 
+// This function is called from the generated code in generate_packets.py
```
lmoureaux, line 15: Yeah it's obvious since it starts with `handle_` :P

server/fcdb.cpp
pranavtifr: Actually Yes, Looks like it has just been assigned somewhere and forgotten after that.

PR #1081, opened 2022-07-01
Widget Resizable now support multiple types of resize

NIKEA-SOFT:
Now we can support several types of "resizable" at the same time
Example:
```cpp
  setResizable(resizable_flag::top | resizable_flag::topLeft
               | resizable_flag::topRight | resizable_flag::bottom
               | resizable_flag::bottomLeft | resizable_flag::bottomRight
               | resizable_flag::left | resizable_flag::right);
```

### DIFF
--- a/client/widgetdecorations.cpp
+++ b/client/widgetdecorations.cpp
@@ -171,6 +171,35 @@ void close_widget::notify_parent()
 }
 
 /**
+   Set resizable flags
+ */
+void resizable_widget::setResizable(QFlags<resizable_flag> flags)
+{
+  resizeFlags = flags;
+}
+
+/**
+   Get resizable flags of wdiget
+ */
+QFlags<resizable_flag> resizable_widget::getResizable() const
+{
+  return resizeFlags;
+}
+
+/**
+   Remove all resizable flags
+ */
+void resizable_widget::removeResizable() { resizeFlags = {}; }
+
+/**
+   Check if resizable flag is active
+ */
+bool resizable_widget::hasResizable(resizable_flag flag) const
+{
+  return resizeFlags.testFlag(flag);
+}
+
+/**
    Checks if info_tab can be moved
  */
 void resizable_widget::mousePressEvent(QMouseEvent *event)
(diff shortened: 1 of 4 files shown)


client/widgetdecorations.h
```diff
@@ -74,16 +74,38 @@ class fcwidget : public QFrame {
...
+  enum resizable_type {
+    top,
+    topLeft,
+    topRight,
+    bottom,
+    bottomLeft,
+    bottomRight,
+    left,
+    right
+  };
```
lmoureaux, line 92: Would this benefit from a `QFlags` declaration?

client/widgetdecorations.cpp
```diff
@@ -179,77 +203,180 @@ void resizable_widget::mousePressEvent(QMouseEvent *event)
...
-    resize(event->x(), this->geometry().height() - newheight);
-    to_move = event->globalPos() - cursor;
-    move(this->x(), to_move.y());
-    setCursor(Qt::SizeBDiagCursor);
+
+  if (event->buttons() & Qt::LeftButton) {
+    if (drag_event != -1) {
+      // constants of min size
+      constexpr auto min_width = 20;
+      constexpr auto min_height = 20;
```
lmoureaux, line 297: Why not `minimumSizeHint`?

lmoureaux (commented):
Two questions from a first look -- will do a more detailed test later on

client/widgetdecorations.cpp
NIKEA-SOFT: > Why not `minimumSizeHint`?

`min_width` and `min_height` mean minimum width and height dimensions

client/widgetdecorations.h
NIKEA-SOFT: > Would this benefit from a `QFlags` declaration?

Um, why?
What advantages does this give?

lmoureaux: You don't need to invent operations with the `flags` array and you can just OR enumerators like in C (but still in a type-safe manner). See https://doc.qt.io/qt-5/qflags.html#details

NIKEA-SOFT: > You don't need to invent operations with the `flags` array and you can just OR enumerators like in C (but still in a type-safe manner). See https://doc.qt.io/qt-5/qflags.html#details

done

client/widgetdecorations.cpp
NIKEA-SOFT: > Why not `minimumSizeHint`?

thank's, fixed

client/widgetdecorations.cpp
```diff
@@ -170,6 +170,32 @@ void close_widget::notify_parent()
...
+   Remove the active resizable_type from list
+ */
+void resizable_widget::removeResizable() { resizeFlags = {}; }
+
+/**
+   Check the active resizable_type or not
+ */
+bool resizable_widget::hasResizable(resizable_flag flag) const
+{
+  return resizeFlags.testFlag(flag);
```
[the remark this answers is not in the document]
NIKEA-SOFT: > I guess it could be made inline

Are you talking about the inline specifier?
```cpp
inline bool resizable_widget::hasResizable(resizable_flag flag) const {}
```

client/widgetdecorations.cpp
```diff
@@ -170,6 +170,32 @@ void close_widget::notify_parent()
   fcw->update_menu();
 }
 
+/**
+   Add the resizable_type to the list (multiple types support)
```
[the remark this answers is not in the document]
NIKEA-SOFT: > I think the comment needs an update

Oh, yes, I forgot about the comments after change api

client/widgetdecorations.cpp
```diff
@@ -170,6 +170,32 @@ void close_widget::notify_parent()
...
+ */
+void resizable_widget::setResizable(QFlags<resizable_flag> flags)
+{
+  resizeFlags = flags;
```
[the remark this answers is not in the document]
NIKEA-SOFT: > What happens if we call this in the middle of a resize?

Nothing, this variable is only responsible for which flags are supported. 
Another variable is used to store active events

client/widgetdecorations.cpp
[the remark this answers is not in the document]
lmoureaux: The specifier is mostly ignored by compilers nowadays, but we could help them by moving the implementation to the header. Anyway not a critical issue :)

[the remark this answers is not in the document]
NIKEA-SOFT: > The specifier is mostly ignored by compilers nowadays, but we could help them by moving the implementation to the header. Anyway not a critical issue :)

Yeah

client/widgetdecorations.cpp
```diff
@@ -170,6 +170,32 @@ void close_widget::notify_parent()
...
+}
+
+QFlags<resizable_flag> resizable_widget::getResizable() const
```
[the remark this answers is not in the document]
NIKEA-SOFT: > Let's add a docstring. Does it make sense to inline it?

The fact is that there is no point in doing inline in this place, this function is not called so often.  
And as you said earlier, it makes sense to do inline only in header files.

lmoureaux:
Spotted a strange behavior:

* Grab the bottom left corner of the chat window
* Move the mouse up until above the top of the chat window. The chat window hits its minimum height, which is fine.
* Move the mouse down again. The chat window expands, but the corner is "disconnected" from the mouse.

Also found that it's possible to move the chat window out of the screen if the main window is smaller than the screen:

* Grab the left side, move it to the left (out of the window area)
* Grab the right side, move it to the left (also out of the window area)

I suggest to prevent moving any edge out of the screen.

lmoureaux:
Could also use `Qt::Edges` instead of a custom enum

jwrober:
Merging with associated issues.

PR #1082, opened 2022-07-05
Bugfix/remove unused functions4

pranavtifr:
I'll rebase after merging #1079

### DIFF
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -1188,8 +1188,6 @@ bool city_has_building(const struct city *pcity,
                        const struct impr_type *pimprove)
 {
   if (nullptr == pimprove) {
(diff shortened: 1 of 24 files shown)


common/city.cpp
```diff
@@ -1188,8 +1188,6 @@ bool city_has_building(const struct city *pcity,
                        const struct impr_type *pimprove)
 {
   if (nullptr == pimprove) {
-    /* Callers should ensure that any external data is tested with
-     * valid_improvement_by_number() */
```
lmoureaux, line 1192: This seems out of scope for unused code cleanup

pranavtifr: Thats cuz, I deleted valid_improvement_by_number()

Issue #429, opened 2021-04-13
Unit colors that match the player color


lmoureaux:
It would be nice if we could replace some color in unit sprites by the player color. This would make opposing armies more distinct from each other, adding graphical depth to the game. Same for city colors.

I'm not sure how this is usually done to allow shading etc, so some research will be needed.

lmoureaux:
Demo of the WIP implementation in #1095 
![image](https://user-images.githubusercontent.com/22327575/178082128-c92a668e-7b3f-46b2-97f2-b67beab450f8.png)

PR #1083, opened 2022-07-06
Refactor unit map layers

lmoureaux:
This PR refactors the code that draws units on the map and some code around it. I wanted to implement #429 but figured that some refactoring was needed first. This PR introduces no behavior change.

**Things to check**
* Are units still drawn where they should?
* Do options like drawing only the focused unit still work as intended?
* Does it crash?

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -82,6 +82,7 @@ add_library(
   layer_darkness.cpp
   layer_special.cpp
   layer_terrain.cpp
+  layer_units.cpp
   layer.cpp
   luaconsole.cpp
   luaconsole_common.cpp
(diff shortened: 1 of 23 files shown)


pranavtifr (approved):
Take this chance to move layers to a separate folder ? I think something like, client/layers/layer_*.cpp/h is better. But I'll approve anyways, since that can be done in a different PR, where we reorganize everything within client. Apart from that, rest look good to me.

lmoureaux:
Thanks for the review! I agree that layer classes should eventually be moved to another folder, but I'm not sure how to call it: it could be for instance `tileset/layers` or simply `tileset` next to related classes. It's the only folder that I'm sure of, so I'm leaving this for later.

PR #1095, opened 2022-07-08
Add tilespec support for more colorful units

lmoureaux:
Implements #429 for units. No tileset supports it yet (I have WIP changes for amplio2 but it will take more time and a different kind of review).

Will rebase once we get #1083 

**Testing it**
* Open `units.png` for your favorite tileset and draw in magenta over the units (`#ff00ff` or any color with the same HSL hue)
* Set `replaced_hue = 300` in the `tilespec`
* Load a game with many units and many nations and check how units are colored with the player color!
* Look around other places where units are displayed to check for weird displays (I don't think any of them uses the player color)

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -56,6 +56,7 @@ add_library(
   climap.cpp
   climisc.cpp
   clinet.cpp
+  colorizer.cpp
   colors.cpp
   colors_common.cpp
   connectdlg.cpp
--- a/client/colorizer.cpp
+++ b/client/colorizer.cpp
@@ -1,0 +1,66 @@
+/*
+ * SPDX-FileCopyrightText: 2022 Louis Moureaux <m_louis30@yahoo.com>
+ *
+ * SPDX-License-Identifier: GPLv3-or-later
+ */
+
+#include "colorizer.h"
+
+#include <QBrush>
+#include <QPainter>
+
+namespace freeciv {
+
+/**
+ * @class colorizer Swaps colors in QPixmap
+ *
+ * Starting from a base pixmap, this class generates new pixmaps with one
+ * color replaced (for instance, all pink pixels changed to green).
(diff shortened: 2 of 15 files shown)


pranavtifr:
Seemed to work for me in a quick test. 
A few overall comments not related to the code. The code looks good to me!  
1) I think we need to have atleast one good tileset to go along with this PR where it is immediately seen by the user, so that it doesnt endup as an unused feature. 
2) We need to document this
3) Add a client option to disable this/make the colour based on nationality instead of ownership. I dont think there is one right answer to that and it is upto preference. 
4) Do the same thing to cities to have a unified look (We can leave that for later, just need to create an issue then)

lmoureaux:
> 1. I think we need to have atleast one good tileset to go along with this PR where it is immediately seen by the user, so that it doesnt endup as an unused feature.

I started adjusting `amplio2` to use this, but it will take some time. I think `hexemplio` and `trident` should also be adapted. -> New issue

> 2. We need to document this

Another PR (it's actually "documented" just like most other tileset options: in the shipped tilesets). -> New issue

> 3. Add a client option to disable this/make the colour based on nationality instead of ownership. I dont think there is one right answer to that and it is upto preference.

These hypothetical users can still use a tileset that doesn't do it (or edit the tileset they like). Freeciv had this culture of making everything an option, and ended up with many unused options.

pranavtifr:
> These hypothetical users can still use a tileset that doesn't do it (or edit the tileset they like). Freeciv had this culture of making everything an option, and ended up with many unused options.
Good point. I agree, what seems more important is the defaults. Very few people play around with options.  I guess a tilespec option then. To use nationality or ownership.

lmoureaux:
> Do the same thing to cities to have a unified look (We can leave that for later, just need to create an issue then)

Will do some refactoring first -> new PR

PR #1103, opened 2022-07-09
Fix minimap show/hide

### LINKED ISSUE #1096: Behaviour of "hide minimap" is broken
**Describe the bug**
Using the hide minimap option from the menu or "ctrl+M" must just hide the minimap, but it causes unintended behaviour

**To Reproduce**
1. Hide minimap from view menu, or Ctrl + M
2. Notice, that the widget is still left in the bottom as a blank space
3. It also disables moving widgets around the screen . You cant move unit-actions, chat window or any widgets around

**Expected behavior**
Just the minimap disappears along with the empty space, and the zoom buttons stay above the turn done button. It shouldnt affect the behaviour of widgets

**Screenshots**
image
Notice that the move handle in the corner of widgets have disappeared. 

 - OS:Linux
 - Freeciv21 version:master

lmoureaux:
Closes #1096.

I chose not to remove the option (for now), instead implementing a more or less sensible behavior.

### DIFF
--- a/client/icons.cpp
+++ b/client/icons.cpp
@@ -68,6 +68,10 @@ public:
     Q_UNUSED(mode)
     Q_UNUSED(state)
 
+    if (rect.width() <= 0 || rect.height() <= 0) {
+      return;
+    }
+
     auto col = QApplication::palette().color(QPalette::ButtonText);
     QString key = path + "-" + QString::number(rect.width()) + "-"
                   + QString::number(rect.height()) + "-" + col.name();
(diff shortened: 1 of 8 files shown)


pranavtifr (changes requested):
If you dont lock the interface, it resets the position of the chat widget to the default position. That is jarring, and isnt the expected behaviour. The minimap should hide/show without involving the other widgets.

lmoureaux:
> If you dont lock the interface, it resets the position of the chat widget to the default position. That is jarring, and isnt the expected behaviour. The minimap should hide/show without involving the other widgets.

Fixed. Yes, I used a `switch` with intentional `[[fallthrough]]` (and the C++17 attribute).

pranavtifr:
Hmm! MacOS build failed!didn't expect that! 😄

pranavtifr (approved):

PR #1112, opened 2022-07-10
Refactor/remove functions 6

pranavtifr:


client/mapctrl_common.cpp
```diff
@@ -66,24 +66,6 @@ struct city *city_workers_display = nullptr;
 static void clipboard_send_production_packet(struct city *pcity);
 static void define_tiles_within_rectangle(bool append);
 
-/**
-   Called when Right Mouse Button is depressed. Record the canvas
-   coordinates of the center of the tile, which may be unreal. This
```
lmoureaux, line 71: These two sentences were pure gold.

client/options.cpp
```diff
@@ -113,13 +113,11 @@ struct client_options gui_options = {
     true,                           //.show_previous_turn_messages =
     false,                          //.concise_city_production =
     false,                          //.auto_turn_done =
-    true,                           //.meta_accelerators =
     true,                           //.ask_city_name =
     true,                           //.popup_new_cities =
     true,                           //.popup_actor_arrival =
     true,                           //.popup_attack_actions =
     true,                           //.popup_last_move_to_allied =
-    true,                           //.update_city_text_in_refresh_tile =
```
lmoureaux, line 122: You need to remove the corresponding initializers or everything will be shifted.

lmoureaux (changes requested):

client/options.cpp
pranavtifr: I did. Thats why i deleted these.

pranavtifr: Check line 93,99 in options.h

lmoureaux: Oooh they already had no
```cpp
    GEN_BOOL_OPTION(
        auto_turn_done, N_("End turn when done moving"),
        N_("Setting this option makes your turn end automatically "
           "when all your units are done moving."),
        COC_INTERFACE, false, nullptr),
```

lmoureaux (approved):

PR #1129, opened 2022-07-13
Refactor/remove stuff10

pranavtifr:


lmoureaux (approved):
Would be good to have an explanation of why you removed `Q_UNREACHABLE()`

pranavtifr:
> Would be good to have an explanation of why you removed `Q_UNREACHABLE()`

I have modified the commit to include an explanation.

PR #1132, opened 2022-07-14
Refactor/remove stuff11

pranavtifr:
Still have to simplify further and remove some intermediate variables. 
Test: 
There should be no behavioral change. Just play for a few turns, load a save and the lot and see if anything happens.

### DIFF
--- a/client/diplodlg.cpp
+++ b/client/diplodlg.cpp
@@ -39,9 +39,6 @@
 #include "sprite.h"
 #include "top_bar.h"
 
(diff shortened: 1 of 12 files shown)


jwrober (approved):

lmoureaux:
Can some of the loops on `research_array` be turned to `const` references?

pranavtifr:
> Can some of the loops on `research_array` be turned to `const` references?

Done

PR #1164, opened 2022-07-30
Fix build with NLS enabled

lmoureaux:
We need to include fc_config.h in fcintl.h to control enabling NLS. Else we're
defining "textdomain" as nothing, which doesn't play nice with libintl.h.


jwrober (commented):
Approved

PR #1165, opened 2022-07-30
Get rid of fileinfo and fileinfo_list

lmoureaux:
Use QFileInfo and QFileInfoList instead. While we're at it, simplify the API a
little.


lmoureaux:
The CI needs #1164

pranavtifr (approved):
Loading a save seems to work just fine.

Issue #115, opened 2020-11-22
Qt client is "ugly"


lmoureaux:
This is the adjective I get most often when people discover the Qt client. So we should work on it.

mir3x:
something can be uglier than this ??
![image](https://user-images.githubusercontent.com/43116166/99916354-1fb7e380-2d0a-11eb-9425-c1306569253a.png)

lmoureaux:
Heck, ask them. They're too used to Gtk :crying_cat_face:

pranavtifr:
So, I have been recently using qt and gtk together. I think predominantly the complaint comes from the fact that, qt shows a lot more information on screen. It is great once you are used to it. But the fact remains that, people would like to switch on and off certain information, and it feels really cluttered initially.

Second factor is the defaults. qt is a great client, but I think many people are thrown off by the defaults. 

Here are my few suggestions which would make qt less "ugly". They thematically follow the notion of removing certain unnecessary addons and provide an option to remove certain elements from the interface. 


Before I start, the first thing I need to say is, the view window in qt is the best. The smooth scroll makes everything more convenient, it would be nice if the other elements dont obstruct this. 

Anyways, the first change IMO would be to change the default theme. A lot of people are thrown away by the nightstalker theme (including me). I was able to use qt only because it supports the system theme, and people who are forced to use qt quickly change themes. It is better if the first view is not nightstalker and maybe give classic to people who are used to gtk. 

Secondly, Qt likes to keep floating windows in order to not move away from the view tab. I think it is a valid and great design choice, but this need to be done better. A few changes on this front IMO would make the client "cleaner" are, 

- Remove those icons next to view, units, nations etc. They occupy space and they don't look good on nightstalker theme. In the extra space from the lack of icons, I would like a "chat" tab which turns on and off the floating chat tab. It is too inconvenient to keep on screen, and also the minimap tab which can be used to turn on and off the minimap. This lets people use the view window to the fullest. 

- There should be a setting to make the chat window not transparent. The transparency makes it more cluttered. This problem is there in the units window and the demographics window too. 

[...]

PR #1166, opened 2022-07-30
Improve the layout of the "load game" page

lmoureaux:
This PR changes the layout of the "load game" screen to jump around less. There are still huge improvements to be made but I think it's a good start.

The new layout looks like this:
image

After #1165.
Related to #115.

### DIFF
--- a/client/page_load.cpp
+++ b/client/page_load.cpp
@@ -15,6 +15,7 @@
 // Qt
 #include <QDateTime>
 #include <QFileDialog>
+#include <qpushbutton.h>
 // utility
 #include "fcintl.h"
 // common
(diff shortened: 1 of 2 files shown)


jwrober:
So with the default NightStalker theme, the dialog overlaps the LongTurn logo as shown above and here.

image
image

Looks unprofessional. I would suggest increasing the size the the dialog so the image shows uncovered in the lower right corner. Since the other shipped themes don't have an image there it should be fine.

Also if you open the dialog, click cancel and then load it again you will briefly see the new game dialog and then it is replaced by the load game dialog.

Lastly, if you open the dialog, pick a saved game to show the preview, click cancel and then load saved game again, the preview of the game you clicked on the first try still shows, but the save game file in the table is not selected. I think it should be clean every time.

lmoureaux:
> Also if you open the dialog, click cancel and then load it again you will briefly see the new game dialog and then it is replaced by the load game dialog.

I don't think this one is new.

jwrober:
> > Also if you open the dialog, click cancel and then load it again you will briefly see the new game dialog and then it is replaced by the load game dialog.
> 
> I don't think this one is new.

I can recreate in beta.3, so yea not new. But if can be fixed here while the code is being worked on would be good.

lmoureaux:
> > > Also if you open the dialog, click cancel and then load it again you will briefly see the new game dialog and then it is replaced by the load game dialog.
> > 
> > 
> > I don't think this one is new.
> 
> I can recreate in beta.3, so yea not new. But if can be fixed here while the code is being worked on would be good.

Looks like a timing-dependent issue, it doesn't happen very often for me.

lmoureaux:
Should have addressed most concerns. The LT logo will still be partially covered at small window sizes, but now it should at least be fine when the window is big enough.

jwrober (approved):
Looks good to me.

lmoureaux:
OS X CI: https://gitlab.freedesktop.org/ is down

pranavtifr (approved):
Looks good to me. Can merge after rebasing.

PR #1167, opened 2022-07-30
Don't try to set the ruleset based on the tileset

lmoureaux:
The last used tileset was saved in the options file, but that value wasn't shown anywhere in the interface. It was then used *instead* of the tileset configured in the settings when the topology matched. This was a source of confusion. Get rid of this behavior by removing the hidden option.

---

The tileset in pregame isn't well defined because the topology isn't known yet. Always start servers with the default ruleset, and load a tileset based on that.

If the user want to spawn a client with a given ruleset, they can use the `-r` option.

### DIFF
--- a/client/options.h
+++ b/client/options.h
@@ -62,9 +62,8 @@ struct client_options {
   bool save_options_on_exit;
 
   /** Migrations **/
-  bool first_boot;                /* There was no earlier options saved.
-                                   * This affects some migrations, but not all. */
-  char default_tileset_name[512]; // pre-2.6 had just this one tileset name
+  bool first_boot; /* There was no earlier options saved.
+                    * This affects some migrations, but not all. */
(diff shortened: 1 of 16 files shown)


client/options.h
```diff
@@ -62,9 +62,8 @@ struct client_options {
   bool save_options_on_exit;
 
   /** Migrations **/
-  bool first_boot;                /* There was no earlier options saved.
-                                   * This affects some migrations, but not all. */
-  char default_tileset_name[512]; // pre-2.6 had just this one tileset name
+  bool first_boot; /* There was no earlier options saved.
```
pranavtifr, line 65: Why reformat comments ? :/

pranavtifr (approved):

client/options.h
lmoureaux: Clang-format because it aligns them

pranavtifr: Argh, breaks git blame. But ok I guess.

PR #1168, opened 2022-07-30
Remove static calls to units_reports

lmoureaux:
`units_reports` is the units list widget. This replaces static functions to pop it open and close it with Qt signals and slots.

Widgets are better managed by their parents: this way the parent knows exactly
when and why its children change (well, signals and slots might break this in
principle, but the parent knows how it connects them).

### DIFF
--- a/client/dialogs.cpp
+++ b/client/dialogs.cpp
@@ -64,7 +64,6 @@
 extern void popdown_all_spaceships_dialogs();
 extern void popdown_players_report();
 extern void popdown_economy_report();
-extern void popdown_units_report();
 extern void popdown_science_report();
 extern void popdown_city_report();
 extern void popdown_endgame_report();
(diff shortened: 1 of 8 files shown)


jwrober:
I cannot load a saved game on this one from master.

jwrober:
I would like to get the PR for the saved game loading done and merged and then rebase that into this to make sure the issue is resolved. Other than that, units widget looks unchanged on a scenario and single player game I tried. Might be good to also login to the current AVI game to see if there is any issue with UWT table.

lmoureaux:
> I cannot load a saved game on this one from master.

What happens? In principle this doesn't touch anything related to saved game loading so it might be unrelated

jwrober:
> > I cannot load a saved game on this one from master.
> 
> What happens? In principle this doesn't touch anything related to saved game loading so it might be unrelated

I get a new game dialog, not the saved game

jwrober (approved):
Load game issue on my end.

PR #1169, opened 2022-07-30
Preliminary cleanup of mapview_common

lmoureaux:
This PR contains a collection of small commits that improve the quality of the `mapview_common` code that handles drawing the map and updating it when needed. The commits are mostly independent from each other and I'm happy to take out controversial ones and discuss them in another PR. No behavior change is expected.

I strongly suggest to review this commit by commit. See individual commit messages for more context.

PR #1170, opened 2022-07-30
Misc small refactorings

lmoureaux:
Same as #1169 but at random locations (loosely related to the pregame code).


client/mapctrl.cpp
```diff
@@ -116,35 +114,31 @@ void map_view::keyPressEvent(QKeyEvent *event)
     case Qt::Key_Up:
     case Qt::Key_8:
       if (is_shift) {
-        recenter_button_pressed(queen()->mapview_wdg->width() / 2 / scale,
```
pranavtifr, line 119: LOL

PR #1171, opened 2022-07-30
Introduce a map renderer class

lmoureaux:
It is meant to incorporate the functionality of mapview_common needed to draw
the map. Right now it delegates most (all) of its work to mapview_common.

Developed on top of many patches from #1169 but auto rebase worked fine so I'm making a PR to `master`.

### DIFF
--- a/client/renderer.h
+++ b/client/renderer.h
@@ -1,0 +1,41 @@
+/*
+ * SPDX-FileCopyrightText: 2022 Louis Moureaux <m_louis30@yahoo.com>
+ *
+ * SPDX-License-Identifier: GPLv3-or-later
+ */
+
+#pragma once
+
+#include <QObject>
+#include <QPainter>
+#include <QRect>
+#include <QRegion>
+#include <QSize>
+
+namespace freeciv {
+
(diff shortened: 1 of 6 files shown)


client/renderer.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <QObject>
+#include <QPainter>
+#include <QRect>
+#include <QRegion>
+#include <QSize>
+
+namespace freeciv {
+
+class renderer : public QObject {
```
pranavtifr, line 17: Now that we are converting more and more stuff to classes, we need to think a bit about the naming convention. We wanted to use CamelCase to fit with Qt's convention which we use everywhere. I'm approving this, but is the naming convention finalized. If yes, we need to document it in contributor's guidelines.

pranavtifr (approved):

client/renderer.h
lmoureaux: I've been using `lower_snake` everywhere, which also matches existing client classes. I don't think the convention is documented anywhere or implemented, so it's not finalized :yum: We'll also want to add an automatic check.

PR #1174, opened 2022-07-30
Refactor how map updates are propagated to widgets

lmoureaux:
Map changes (e.g. units moving, terrain changing, etc.) are propagated to the map view and the minimap using static functions in `mapview_common.cpp`. These functions had to (magically) know which widgets were waiting for updates and send them as appropriate. This made the updates hard to track, because the widget code on its own didn't "subscribe" to any notification.

Change this to use the `listener<>` template. `listener<>` is meant to handle the case of C-like static code calling into C++-like object-oriented code in a way that respects encapsulation, so it's natural to use it here.

After the change, mapview functions called upon updates no longer need to call into the minimap directly because the minimap receives the updates directly (thus we get separation of purposes).

On top of #1169.


lmoureaux:
Rebased

PR #1175, opened 2022-07-31
Fix new Coverity warnings

lmoureaux:
A handful of Coverity warnings popped up in the last month. This should clear them all.

Special thanks to Synopsis for 55336e7b7bc37cb046c78a259fa6a99dbb985266: *The client was crashing when Ctrl+Alt+clicking outside of the map (this shortcut is bound to the "create link" action).*

### DIFF
--- a/server/auth.cpp
+++ b/server/auth.cpp
@@ -35,11 +35,6 @@ _   ._       Copyright (c) 1996-2021 Freeciv21 and Freeciv contributors.
 #define GUEST_NAME "guest"
 
 #define MIN_PASSWORD_LEN 6 // minimum length of password
-#define MIN_PASSWORD_CAPS                                                   \
-  0                         /* minimum number of capital letters required   \
-                             */
-#define MIN_PASSWORD_NUMS 0 // minimum number of numbers required
-
 #define MAX_AUTH_TRIES 3
 #define MAX_WAIT_TIME 300 // max time we'll wait on a password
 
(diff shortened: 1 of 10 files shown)


server/auth.cpp
```diff
@@ -35,11 +35,6 @@ _   ._       Copyright (c) 1996-2021 Freeciv21 and Freeciv contributors.
 #define GUEST_NAME "guest"
 
 #define MIN_PASSWORD_LEN 6 // minimum length of password
```
pranavtifr, line 37: Cant we remove MIN_PASSWORD_LEN and is_good_password() entirely (just check for ascii and garbage and keep the function as is_password_valid() or something) ? If we dont want to have custom servers for password quality, we shouldnt bother with length too imo.

tools/civmanual.cpp
```diff
@@ -644,6 +644,7 @@ static bool manual_command(struct tag_types *tag_info)
       break;
 
     case MANUAL_COUNT:
```
pranavtifr, line 646: Why do we need to keep this MANUAL_COUNT again ? Some backward compatibility ? If yes, we need to comment it.

pranavtifr:
The fails dont look like garbage, WA and Win doesnt seem to like the `std::array` conversion.

tools/civmanual.cpp
lmoureaux: It's the upper bound for the loop iterating over all manual types.

server/auth.cpp
lmoureaux: Let's say that such a discussion is out of the scope of this PR which is to fix Coverity warnings about always-true conditionals. We can open an issue if we want to remove it.

pranavtifr:
All the checks have passed now. Merging.

PR #1183, opened 2022-08-01
Use compile_commands.json with clang-tidy

lmoureaux:
Otherwise it can't find headers. Also reduce the number of steps using features
of clang-tidy-review.

Hopefully this works...

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -4,49 +4,17 @@ on: [pull_request]
 jobs:
   build:
     name: clang-tidy
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        compiler: [gcc]
-        build_type: [Debug]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
-      - name: Set base_dir
-        run: echo "base_dir=$(pwd)" >> $GITHUB_ENV
-
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Install GCC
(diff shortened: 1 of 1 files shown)


pranavtifr:
Set base dir is probably not needed. I added that to try and use the artifacts from Ubuntu build..but doesn't work that way.

lmoureaux:
Yeah I removed it, but I added `build_dir` to match what we do elsewhere. I'm not even sure if we support in-source builds :sweat_smile:

pranavtifr:
Uh oh. You broke it 😛
And i meant 
```
      - name: Set base_dir
        run: echo "base_dir=$(pwd)" >> $GITHUB_ENV
```
These aren't needed either.

pranavtifr:
A nice improvement would be if we can cache clang-tools. Then this would a fast check!

pranavtifr:
Can you bump version to 0.9.0 ? I think that fixes it.

pranavtifr:
Still fails :/

lmoureaux:
```
clang-tidy failed with return code 1 and error:
Error while trying to load a compilation database:
Could not auto-detect compilation database from directory "build"
No compilation database found in /github/workspace/build or any parent directory
fixed-compilation-database: Error while opening fixed database: No such file or directory
json-compilation-database: Error while opening JSON database: No such file or directory
Running without flags.
```

pranavtifr:
Ya, I'm just looking at the integration bug. Clang-tidy seems to work ok in my PR. (I did that so that we can have some extra cpp changes to check with), 
The integration bug wasnt there yesterday :/

lmoureaux:
Let's merge this so we get a green check mark at least?

PR #1201, opened 2022-08-06
Improve the city dialog unit lists

### LINKED ISSUE #805: Activating units in city window
**In city window:**

Current state:
double-clicking or right-click-"activate" on a unit activates unit and closes city window

Problem: 
If I want to activate more units from the same city, I have to reopen the city window for each unit

Desired behaviour: 
Same as in GTK: have separate commands
"Activate" - activates unit but does not exit city window
"Activate and exit" - what it says

### LINKED ISSUE #1200: Right-clicking the supported units panel interacts with the present units panel
**Describe the bug**

In the city dialog, right-clicking on a unit in the list of supported units pops up the menu in the list of present units.

**To Reproduce**
1. Open a city with both present and supported units
2. Right click on a supported unit
3. See the menu appearing next to the present unit at the same index in the list

**Expected behavior**

The menu opens near the mouse

 - OS: any
 - Freeciv21 version: 3.0-beta3
 - Ruleset/Longturn game (if applicable): any and all

lmoureaux:
Closes #805
Closes #1200

**Test plan**
* Try to repro the above issues

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -60,20 +60,24 @@ extern QString cut_helptext(const QString &text);
 /**
    Constructor
  */
-icon_list::icon_list(QWidget *parent) : QListWidget(parent), oneliner(true)
+unit_list_widget::unit_list_widget(QWidget *parent) : QListWidget(parent)
 {
   // Make sure viewportSizeHint is used
   setSizeAdjustPolicy(QAbstractScrollArea::AdjustToContents);
   setWrapping(true);
   setMovement(QListView::Static);
+
+  setContextMenuPolicy(Qt::CustomContextMenu);
+  connect(this, &QWidget::customContextMenuRequested, this,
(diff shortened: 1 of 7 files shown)


jwrober (approved):
1200 is completely fixed, however 805 is only partially done. The issue asks for an activate unit and activate unit and close option similar to what fc gtk client supports. Right now we get activate, which when selected closes the city dialog. I am ok with approving this PR, but we should open a subsequent issue enhancement asking for a change to the selection menu.

lmoureaux:
> 805 is only partially done

I thought we had agreed in #805 that multiple selection also solved the use case?

jwrober:
> > 805 is only partially done
> 
> I thought we had agreed in #805 that multiple selection also solved the use case?

Oh I missed that. I didn't scroll down. I am good then. I will need to update the manual, this is not totally obvious

PR #1204, opened 2022-08-06
Add the city size and citizen happiness next to other city stats

### LINKED ISSUE #438: Add city size and unhappy to general tab
Classic freeciv GTK 3 client shows it under info section. Makes it super easy to see how large a city is while working in the city dialog.

image

lmoureaux:
Closes #438.
The citizen types (happiness and entertainers) is shown in the tooltip.

The patch itself is rather small, but there's a lot of refactoring to make it less error-prone in the end.


jwrober (approved):
Looks great.

PR #1206, opened 2022-08-07
Expand the text for empire size unhappiness

lmoureaux:
Improve the city dialog tooltip text that explains how many citizens are
unhappy because of the number of cities. Write two paragraphs: the first
describing the rules for the current player, and the second the status of the
city at hand. The total number of citizens that can be made content without
improvements is also shown; see #269.

### classic, size 14, communism

All cities start with 4 content citizens. Once you have more than 12 cities, a citizen becomes unhappy.

You have 36 cities, resulting in a maximum of 3 content citizens. In this city of size 14, **3 citizens are content.**

### classic, size 2, communism

All cities start with 4 content citizens. Once you have more than 12 cities, a citizen becomes unhappy.

You have 36 cities, resulting in a maximum of 3 content citizens. In this city of size 2, **all citizens are content.**

### Aviation, size 4, monarchy

All cities start with 4 content citizens. Once you have more than 10 cities, a citizen becomes unhappy. Afterwards, for every 9 additional cities, a content citizen becomes unhappy. If there are no more content citizens, an unhappy citizen becomes angry instead.

You have 11 cities, resulting in a maximum of 3 content citizens. In this city of size 4, **3 citizens are content.** With 9 more cities, a content citizen would become unhappy.

### Aviation, size 2, monarchy

All cities start with 4 content citizens. Once you have more than 10 cities, a citizen becomes unhappy. Afterwards, for every 9 additional cities, a content citizen becomes unhappy. If there are no more content citizens, an unhappy citizen becomes angry instead.

You have 11 cities, resulting in a maximum of 3 content citizens. In this city of size 2, **all citizens are content.** With 18 more cities, a content citizen would become unhappy.

### DIFF
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -24,6 +24,7 @@
 // common
 #include "calendar.h"
 #include "citizens.h"
+#include "city.h"
 #include "clientutils.h"
 #include "combat.h"
 #include "culture.h"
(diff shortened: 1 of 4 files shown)


client/text.cpp
```diff
@@ -1400,128 +1401,256 @@ QString text_happiness_wonders(const struct city *pcity)
...
+        str += QString(PL_("Afterwards, for every %1 additional city, a "
+                           "content citizen becomes unhappy.",
+                           "Afterwards, for every %1 additional cities, a "
+                           "content citizen becomes unhappy.",
+                           step))
+                   .arg(step);
+
+        if (game.info.angrycitizen) {
+          str += QStringLiteral(" ");
+          str += _("If there are no more content content citizens, an "
```
jwrober, line 1475: Content is there twice

jwrober (changes requested):

client/text.cpp
```diff
@@ -1400,128 +1401,256 @@ QString text_happiness_wonders(const struct city *pcity)
...
+    // TRANS: Pluralized in "%2 angry citizens"
+    return QString(PL_("You have %1 cities, resulting in a maximum of "
+                       "%2 angry citizen.",
+                       "You have %1 cities, resulting in a maximum of "
+                       "%2 angry citizens.",
+                       max_content))
+        .arg(cities)
+        .arg(-max_content);
+  }
+}
```
daavko, line 1510: Does this account for a player having only one city? When that happens, it should be "You have %1 city, ..."

lmoureaux: Unfortunately I don't know of any library allowing pluralization twice in the same sentence :(. Maybe we can rephrase it to avoid "%1 cities" altogether?

daavko: It could be as two separate sentences, so you pluralize each of them separately

lmoureaux: Unfortunately not. Usual pluralization issue :/

jwrober (approved):
I compiled this to see it in game. It looks great so I am in approval.

client/text.cpp
lmoureaux: Proper fix not possible before #1212

PR #1209, opened 2022-08-07
Expand the city dialog text for building happiness (Make_Content)

lmoureaux:
On top of (and after) #1206 and #1208.

Give the total number of citizens that can be made content and details about
active effects. This should help towards #269.

### Big city

Infrastructure can have an effect on citizen happiness. In this city, it can make up to **5 citizens** content.

The following contributions are active:

* 3 citizens from Cathedral+not Michelangelo's Chapel
* 1 citizen from Theology+Cathedral+not Michelangelo's Chapel
* -1 citizens from Communism+Cathedral+not Michelangelo's Chapel
* 1 citizen from Temple
* 1 citizen from Mysticism+Temple

### Smaller city

Infrastructure can have an effect on citizen happiness. In this city, it can make up to **2 citizens** content.

The following contributions are active:

* 1 citizen from Temple
* 1 citizen from Mysticism+Temple

### Sad city

Infrastructure can have an effect on citizen happiness. This city doesn't receive any such bonus.

PR #1208, opened 2022-08-07
Expand the city dialog text for luxury spendings

lmoureaux:
Same as (and on top of) #1206 for luxury. The rules are much simpler and so the code also is.

Improve the city dialog tooltip text that explains how many citizens are
made happier by luxury. This provides a lot more information than the previous
"Luxury: %d total", which will hopefully help towards https://github.com/longturn/freeciv21/issues/269.

### Sad, sad city

For each 2 luxury produced in a city, a citizen becomes happier. This city generates no luxury.

### Normal sad city

For each 2 luxury produced in a city, a citizen becomes happier. This city generates a total of 1 luxury, not enough to have an effect.

### Getting happier

For each 2 luxury produced in a city, a citizen becomes happier. This city generates a total of 5 luxury.

2 citizens are made happier using 4 luxury. 1 luxury is not used.

### Happiest

For each 2 luxury produced in a city, a citizen becomes happier. This city generates a total of 8 luxury.

All available luxury is used to make 3 citizens happier.

PR #1210, opened 2022-08-07
Expand the city dialog text for nationality

lmoureaux:
Follow-up of #1206, #1208, and #1209 for citizen nationality.

Explain the rules and give the result of the calculation. The rule sentence is
not fully correct but it's easier to understand than the actual mechanism.
Hopefully helps with #269.

## Example texts

All examples are with nationality enabled.

### Example 1

The presence of enemy citizens can create additional unhappiness. However, it is not the case in this city.

### Example 2

For every 2 citizens of an enemy nation, one citizen becomes unhappy.

There is **no enemy citizen** in this city.

### Example 3

For every 2 citizens of an enemy nation, one citizen becomes unhappy.

There are 5 enemy citizens in this city, resulting in **2 additional unhappy.**

### DIFF
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -1326,7 +1326,6 @@ QString text_happiness_buildings(const struct city *pcity)
 {
   if (const auto effects = get_effects(EFT_MAKE_CONTENT);
       effect_list_size(effects) == 0) {
-    effect_list_destroy(effects);
     return QString(); // Disabled in the ruleset.
   }
 
@@ -1396,7 +1395,6 @@ QString text_happiness_nationality(const struct city *pcity)
   if (auto effects = get_effects(EFT_ENEMY_CITIZEN_UNHAPPY_PCT);
       !game.info.citizen_nationality || effect_list_size(effects) == 0) {
     // Disabled in the ruleset
-    effect_list_destroy(effects);
     return QString();
   }
 
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -28,6 +28,7 @@
 #include "clientutils.h"
 #include "combat.h"
 #include "culture.h"
+#include "effects.h"
 #include "fc_types.h" // LINE_BREAK
 #include "game.h"
 #include "government.h"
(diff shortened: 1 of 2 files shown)


client/text.cpp
```diff
@@ -1317,63 +1319,134 @@ const QString act_sel_action_tool_tip(const struct action *paction,
...
+      !game.info.citizen_nationality || effect_list_size(effects) == 0) {
+    // Disabled in the ruleset
+    effect_list_destroy(effects);
+    return QString();
+  }
 
-  str = _("Nationality: ") + qendl();
+  auto str = QStringLiteral("<p>");
+  str +=
+      _("The presence of enemy citizens can create additional unhappiness.");
```
jwrober, line 1405: Is the enemy citizens or conquered citizens? or maybe its conquered enemy citizens?

client/text.cpp
```diff
@@ -1400,128 +1473,256 @@ QString text_happiness_wonders(const struct city *pcity)
...
+                           basis))
+                   .arg(basis);
+      } else if (base_content == 1) {
+        // TRANS: Comes after "All cities start with 1 content citizen."
+        //        "it" is the citizen.
+        str +=
+            QString(
+                PL_("Once you have more than %1 city, it becomes unhappy.",
+                    "Once you have more than %1 cities, it becomes unhappy.",
+                    basis))
```
jwrober, line 1514: change "it" to "a citizen"

jwrober (changes requested):

client/text.cpp
lmoureaux: #1206

client/text.cpp
lmoureaux: Citizens of a nation you're at war with, however they came (not sure if city trading changes nationality, for instance)

client/text.cpp
```diff
@@ -1388,42 +1389,64 @@ QString text_happiness_buildings(const struct city *pcity)
...
+    auto unhappy = enemies * pct / 100;
+    // TRANS: Pluralized on "enemy citizen"
+    str +=
+        QString(PL_("There is %1 enemy citizen in this city, resulting in "
+                    "<b>%2 additional unhappy.</b>",
+                    "There are %1 enemy citizens in this city, resulting "
+                    "in <b>%2 additional unhappy.</b>",
+                    enemies))
+            .arg(enemies)
+            .arg(unhappy);
```
daavko, line 1446: This would be ideal as "..., resulting in %2 additional unhappy citizen." with pluralization. Could this be split on the comma and then pluralized separately?

lmoureaux: Splitting would assume that the preferred order is the same in all languages, which I doubt is the case... might be the lesser evil though.

lmoureaux: Decided to split this one that the comma. Also a candidate for #1212 though

jwrober (approved):
Last commit fixed my issue. Approving.

PR #1213, opened 2022-08-07
Expand the city dialog text for military happiness

lmoureaux:
Improve the description of martial law by listing the number of units applying
it and the total number of citizens made happier as a result.
Greatly expand the description of military unhappiness by explaining the three
effects that come into play and providing numbers similiar to the above.

Should help with #269.

## Example text

Up to 2 military units in the city may impose martial law. Each of them makes 2 unhappy citizens content.

1 military unit in this city imposes the martial law. <b>1 citizen</b> is made happier as a result.

Military units in the field may cause unhappiness. The unhappiness caused by military units is reduced by 3 with respect to its normal value.

This city supports 3 agressive military units, resulting in <b>1 additional unhappy citizen.</b>

### DIFF
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -1727,41 +1727,138 @@ QString text_happiness_cities(const struct city *pcity)
 }
 
 /**
-   Describing units that affect happiness.
+ * Describe units that affect happiness.
  */
-const QString text_happiness_units(const struct city *pcity)
+QString text_happiness_units(const struct city *pcity)
 {
-  int mlmax = get_city_bonus(pcity, EFT_MARTIAL_LAW_MAX);
-  int uhcfac = get_city_bonus(pcity, EFT_UNHAPPY_FACTOR);
-  QString str;
+  auto str = QString();
 
-  if (mlmax > 0) {
-    int mleach = get_city_bonus(pcity, EFT_MARTIAL_LAW_EACH);
-    if (mlmax == 100) {
-      str = QStringLiteral("%1").arg(_("Unlimited martial law in effect."))
-            + qendl();
+  /*
+   * First part: martial law
+   */
+  int martial_law_max = get_city_bonus(pcity, EFT_MARTIAL_LAW_MAX);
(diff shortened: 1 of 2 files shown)


client/text.cpp
```diff
@@ -1731,41 +1731,132 @@ QString text_happiness_cities(const struct city *pcity)
...
+  }
+
+  str += _("Military units in the field may cause unhappiness.");
+
+  // Special rule: ignore the first N military unhappy
+  int made_content = get_city_bonus(pcity, EFT_MAKE_CONTENT_MIL);
+  if (made_content > 0) {
+    str += QStringLiteral(" ");
+    str += QString(_("The unhappiness caused by military units is reduced "
+                     "by %1 with respect to its normal value."))
```
daavko, line 1814: This feels clunky. Perhaps something like "With respect to normal, %1 fewer citizens will be unhappy about military units."?

client/text.cpp
```diff
@@ -1731,41 +1731,132 @@ QString text_happiness_cities(const struct city *pcity)
...
+  }
+
+  // Special rule: ignore N military unhappy per unit
+  // Let's hope ruleset authors don't combine them too often, the sentences
+  // are very similar.
+  int made_content_each = get_city_bonus(pcity, EFT_MAKE_CONTENT_MIL_PER);
+  if (made_content_each > 0) {
+    str += QStringLiteral(" ");
+    str += QString(_("The unhappiness caused by each military unit is "
+                     "reduced by %1 with respect to its normal value."))
```
daavko, line 1825: This also feels clunky. Perhaps something like "With respect to normal, %1 fewer citizens will be unhappy per each military unit."?

client/text.cpp
lmoureaux: Better, thanks!

client/text.cpp
lmoureaux: Approved!

jwrober (approved):
Approved

PR #1215, opened 2022-08-08
Don't use nonexistent tileset option names

### LINKED ISSUE #525: Assertion failures in client/options.cpp
LTEx24Avi
```
[critical] freeciv21-client (../client/options.cpp:2900) - Assertion NULL == psoption->help_text failed
[critical] freeciv21-client (../client/options.cpp:2900) - Please report this message at https://github.com/longturn/freeciv21/issues
```

lmoureaux:
tileset_custom_load() was still using some old tileset option names from
Freeciv 2.6. Use only Freeciv21 names.

Closes #525.

### DIFF
--- a/client/menu.cpp
+++ b/client/menu.cpp
@@ -2687,8 +2687,7 @@ void mr_menu::tileset_custom_load()
   const struct option *poption;
   QStringList sl;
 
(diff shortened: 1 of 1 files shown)


jwrober (approved):
The assertion failure can no longer be reproduced.

PR #1228, opened 2022-08-10
Optimize Doxygen docs some

lmoureaux:
This PR:

* Turns Doxygen to use C++ mode as we're writing more and more C++ code as opposed to C.
* Adds the files produced by Doxygen to `.gitignore`
* Improves the documentation of the newly introduced `renderer` class

To run Doxygen, use `doxygen doc/freeciv.doxygen`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ build.ninja
 .vs
 # Project file
 *.kdev4
+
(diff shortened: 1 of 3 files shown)


jwrober (approved):
After installing doxygen and running the command I was able to generate the html output and view it in a browser.

PR #1229, opened 2022-08-10
Remove the two mapview iterators

lmoureaux:
This PR removes two very ugly iterators that were used to iterate on tiles in the map view, and replaces them with a class with the same functionality. The class isn't a proper C++ iterator but it's also not needed. It follows the old code closely but (at least in my opinion) already looks better.


client/mapview_geometry.cpp
```diff
@@ -0,0 +1,189 @@
...
+      return next();
+    }
+
+    if (m_xi % 2 == 0 && m_yi % 2 == 0) {
+      if ((m_xi + m_yi) % 4 == 0) {
+        // Tile
+        m_type = item_type::tile;
+        m_tile = map_pos_to_tile(&(wld.map), si / 4 - 1, di / 4);
+      } else {
+        // Corner
```
pranavtifr, line 111: Take this chance to split this function into smaller functions ? I think some seperation here is good, like isometric and non isometric,  infact I'd split it into 4, iso corner, iso edge, non iso corner, non iso edge.

lmoureaux: Hmmm... it's quite a sensitive function and I ported it without really understanding how it works, basically just changing variable names. Splitting it would be even more risky.

pranavtifr: Alright, I'm merging and creating an issue for this.

PR #1246, opened 2022-08-11
Show currently playing soundtrack in the client

### LINKED ISSUE #376: Display Music Soundtrack being played in the client
A nice feature of the client would be to give the player a small popup that fades in and out (similar to the new turn window at top of the screen) that would provide some basic information about the music being played

- Mood (peace or combat)
- Style or Period (early game = European, Classical, Tropical, Asian, Babylonian, or Celtic -- mid-late game = Renaissance Age, Industrial Age, Electric Age, Modern Age, Post-Modern Age)
- Artist (would require a musicset spec change to do this)
- Sound track filename

pranavtifr:
Fixes #376

### DIFF
--- a/client/hudwidget.cpp
+++ b/client/hudwidget.cpp
@@ -259,24 +259,52 @@ hud_text::hud_text(const QString &s, int time_secs, QWidget *parent)
 void hud_text::show_me()
 {
   show();
-  center_me();
+  position_me(5);
 }
 
 /**
-   Moves to top center parent widget and sets size new size
+   Moves to position in parent widget and sets size new size
  */
-void hud_text::center_me()
+void hud_text::position_me(int corner)
 {
+  /*
+   *  position pattern
+   *
+   *  |1 ... 2|
+   *  |...5...|
+   *  |.......|
+   *  |3 ... 4|
+   */
   int w;
-  QPoint p;
+  QPoint parent_pos = QPoint(parentWidget()->pos());
+  QPoint offset_pos;
 
   w = width();
   if (!bound_rect.isEmpty()) {
     setFixedSize(bound_rect.width(), bound_rect.height());
   }
-  p = QPoint((parentWidget()->width() - w) / 2,
-             parentWidget()->height() / 20);
-  move(p);
+  switch (corner) {
+  case 1:
+    move(parent_pos);
+    break;
+  case 2:
+    offset_pos = QPoint(w, 0);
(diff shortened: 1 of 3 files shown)


client/hudwidget.cpp
```diff
@@ -259,24 +259,52 @@ hud_text::hud_text(const QString &s, int time_secs, QWidget *parent)
...
+void hud_text::position_me(int corner)
 {
+  /*
+   *  position pattern
+   *
+   *  |1 ... 2|
+   *  |...5...|
+   *  |.......|
+   *  |3 ... 4|
+   */
```
lmoureaux, line 277: Except for the center, this looks a lot like [`Qt::Corner`](https://doc.qt.io/qt-5/qt.html#Corner-enum). It's in any case better to use an enum than magic numbers.

client/hudwidget.cpp
```diff
@@ -1630,6 +1629,15 @@ QString popup_terrain_info(struct tile *ptile)
   return ret;
 }
 
+/**
+   Show music info
```
lmoureaux, line 1633: New functions use this style:
```
/**
 * Comment
 */
```
The old style is there only because I didn't dare mass-replacing it with regex.

client/music.cpp
```diff
@@ -62,6 +63,7 @@ void start_style_music()
       if (!tag.isEmpty()) {
         log_debug("Play %s", qUtf8Printable(tag));
         audio_play_music(tag, nullptr, MU_INGAME);
+        show_music_info(tag);
```
lmoureaux, line 66: What's the content of `tag`? Do we want some more text, like "Now playing..."?

pranavtifr: Ya, we do. Tag is just a unique identifier currently. If it works, I'll tweak it to show something nice.

lmoureaux:
There are two factors that conspire to ensure that this PR doesn't work:

* This code deletes the widget when the first new turn HUD is shown (i.e. immediately): https://github.com/longturn/freeciv21/blob/7be9ab587bba5e7989a425bcf4fe4d4ea90441fb/client/hudwidget.cpp#L1620-L1624
* There is a positioning issue that makes the widget start at `y = -30`.

After fixing (or in my case, working around) these issues, the widget is visible as intended:
image

pranavtifr:
I'm closing this for now. I'll redo this better, after #1251

PR #1255, opened 2022-08-12
Update translations

lmoureaux:


jwrober (approved):
I was able to load the en_GB translation just fine.

Issue #1253, opened 2022-08-12
Horizontal borders are not being draw on hex tiles


jwrober:
**Describe the bug**
Horizontal borders are not being draw on hex tiles

**To Reproduce**
1. Open client
2. Load or start a game that uses hex tiles
3. See error in screen shot below

**Expected behavior**
All border lines are drawn

**Screenshots**
![image](https://user-images.githubusercontent.com/23037938/184434220-7887bd9f-4054-4221-a397-c02be5ddb109.png)

**Platform and version (please complete the following information):**
 - OS: Linux
 - Freeciv21 version: Master at a recent hash -- 4634cee5c8813b314c3e99866f159d58d2748753
 - Ruleset/Longturn game (if applicable): N/A

**Additional context**
We've made a number of changes to the underlying map drawing code, seem to have introduced a bug

lmoureaux:
This is 4634cee5c8813b314c3e99866f159d58d2748753

PR #1258, opened 2022-08-13
Fix horizontal edges not showing in iso-hex grids

lmoureaux:
In the initial implementation of gui_rect_iterator, it was assumed that each
iteration point corresponds to a single type of object in the geometry. This
was without counting on the laziness of Freeciv's hex implementation, which
simply collapsed one of the vertical edges into a corner. In hex mode, some
iteration points need to be corners and edges at the same time.

This PR also adds more documentation to the code.

Closes https://github.com/longturn/freeciv21/issues/1253.


jwrober (approved):
This patch fixes the issue.

Issue #1237, opened 2022-08-10
Multiplayer menu is a mess


jwrober:
**Is your feature request related to a problem? Please describe.**
Right now we have this disjointed mutliplayer menu that has a mismash of items that some have to do with actual multipayer games and some that don't or are for a very specific ruleset (warciv). 

**Describe the solution you'd like**
There are a few options.

1. (the easiest) is to simply hide the menu by default and then allow it to be shown with a check box on the interface tab of the local options dialog box. Then don't do anything with the menu, we just sort of make it go away

2. Evaluate what items on the menu should move to another menu. Such as the "delayed go to" options might be good to move to the unit menu as well as the "default action..." menu items.

3. Look more closely at the rest and see what we can do about them. The items specifically for trade route planning are for the WarCiv rules, maybe we do a menu break with a title or something to call that out, or do something more with trade planning in general for the rules that support trade routes (which is pretty much all of the shipped rules)

**Describe alternatives you've considered**
See above, at least 3 different ways to handle this menu

**Additional context**
None

PR #1261, opened 2022-08-13
Add two separators to the multiplayer menu

lmoureaux:
This improves item grouping and thus usability.

See #1237.


jwrober (approved):
That looks better. We still need to figure out what all to do with some of those menu entries. This however at least gives a break between the ones for warciv.

PR #1267, opened 2022-08-14
Initialize city_dialog::pcity

lmoureaux:
I encountered a class of crashes where city_dialog::pcity was garbage. Make
sure that it's always nullptr while the city screen is hidden.

## Testing

Zoom in and out many times while also constantly switching player. `master` crashes occasionally, this patch prevents the crash.


jwrober (approved):
I can't recreate the bug with this patch

PR #1272, opened 2022-08-14
Add commands to control the timeout

### LINKED ISSUE #1151: Port Longturn patches - commands
This is a breakout issue from #503 to track status better.

**Short summary of patches**
New server commands.

**Patches to port**
Commits listed chronologically, beginning from the oldest one.

- longturn/freeciv@12c64874d8 - Add `syncturn` command
- longturn/freeciv@41cc12b401 - Add `/showtimeout` and `/fixtimeout`, fix `/syncturn` to work right

lmoureaux:
Add three new timeout-related server commands:

- timeoutshow displays the current timeout
- timeoutadd adds time to the current turn
- timeoutset sets the remaining time for the current turn

They are named timeoutsomething for three reasons:

- timeoutincrease already exists
- it's easier to find them if they're together in /help commands
- sub-commands like "timeout show" are not feasible with the current
  implementation

This is heavily inspired from an earlier patch used by LT with 2.6,
41cc12b401fe4a9db8d22b78039561710fff59e3. The syntax is however more flexible,
allowing for hours or minutes to be passed directly without doing any
calculation.

Closes #1151.

### DIFF
--- a/server/commands.cpp
+++ b/server/commands.cpp
@@ -423,6 +423,32 @@ static struct command commands[] = {
      N_("If there is none, become the game organizer with increased "
         "permissions."),
      nullptr, nullptr, CMD_ECHO_ADMINS, VCF_NONE, 50},
+    {"timeoutshow", ALLOW_INFO, SYN_ORIG_("timeoutshow"),
+     N_("Show timeout settings and timers."),
+     N_("Shows information about the timeout for the current turn, for "
+        "instance how much time is left."),
+     nullptr, CMD_ECHO_ALL, VCF_NONE, 0},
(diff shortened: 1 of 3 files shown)


lmoureaux:
New commit:

> Both timeoutset and timeoutadd accept negative value for each part of their time argument (hours, minutes, seconds). I think using this feature shouln't be encouraged as it makes the time less intuitive, but I see no reason to add more code to prevent it.
>
> Still, documenting that timeoutadd can be used to reduce the timeout is important, so mention that it accepts negative values. Creative users may discover that 1:-10 is 50 seconds by themselves.

PR #1274, opened 2022-08-14
Remove comments headers with too many stars

lmoureaux:
As a rule, we keep our function header comments simple:
```
/**
 * Some comment
 */
```
The two stars are for Doxygen to recognize a docstring.

There were still parts of the code with older, longer headers with many stars:
```
/***************************
```
or
```
/**********************/ /**
```
The first variant confuses Doxygen, and the second is just ugly (tm). I used
sed and a few regexes to edit all comments using the old variants. In some
cases, the comments were describing the whole file; in such cases I added \file
for Doxygen to pick them up.

### DIFF
--- a/ai/default/aihand.cpp
+++ b/ai/default/aihand.cpp
@@ -58,7 +58,7 @@
 
 #include "aihand.h"
 
-/*****************************************************************************
+/**
   A man builds a city
   With banks and cathedrals
   A man melts the sand so he can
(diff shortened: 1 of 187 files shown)


ai/default/aihand.cpp
```diff
@@ -58,7 +58,7 @@
 
 #include "aihand.h"
 
-/*****************************************************************************
+/**
   A man builds a city
   With banks and cathedrals
   A man melts the sand so he can
```
pranavtifr, line 64: Always cracks me up lol.

lmoureaux: first time I see it :joy:

Issue #1273, opened 2022-08-14
Change default fonts to more sane values


jwrober:
**Is your feature request related to a problem? Please describe.**
With a fresh install or if you delete the `freeciv-client-rc-3.0` file, the default fonts are a bit ridiculous. Here are the defaults on my Ubuntu KDE Linux laptop:

* `gui_qt_font_chatline="Liberation Mono,20,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_city_names="Liberation Sans,22,-1,5,75,0,0,0,0,0"`
* `gui_qt_font_city_productions="Ubuntu,20,-1,5,75,0,0,0,0,0"`
* `gui_qt_font_default="Liberation Sans,22,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_help_text="Liberation Sans,20,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_notify_label="Liberation Mono,20,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_reqtree_text="Ubuntu,22,-1,5,75,0,0,0,0,0"`

and produces screen shots like these:
![image](https://user-images.githubusercontent.com/23037938/184557669-8fa4e733-b84f-42a7-991b-b820db49a427.png)
![image](https://user-images.githubusercontent.com/23037938/184557674-979cf685-0ee2-4e32-9cbe-b1698c051e03.png)
![image](https://user-images.githubusercontent.com/23037938/184557679-6b9b6de0-0269-40e7-a5da-06a83ac8aa9d.png)
![image](https://user-images.githubusercontent.com/23037938/184557684-2abd4cba-5475-427c-a80e-e19be5b065b5.png)
![image](https://user-images.githubusercontent.com/23037938/184557687-5b78bd38-fb9e-464f-ad62-97fa0dd64081.png)

These fonts are overpowering and way too big.

**Describe the solution you'd like**
I would like to suggest a default set of fonts like this list:

* `gui_qt_font_help_label="Sans Serif,12,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_chatline="Monospace,12,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_city_names="Sans Serif,14,-1,5,50,0,0,0,0,0"`
* `gui_qt_font_city_productions="Sans Serif,14,-1,5,75,1,0,0,0,0"`
* `gui_qt_font_default="Sans Serif,12,-1,5,50,0,0,0,0,0"`
[...]

PR #1281, opened 2022-08-15
Don't default to huge fonts of HiDpi systems

lmoureaux:
We have enabled Qt's transparent handling of HiDpi, and thus the manual scaling
that was done in configure_fonts() was useless. Worse, on HiDpi diplays, it
would choose larger fonts to make the UI bigger, which Qt also does internally.
This resulted in huge fonts on such systems.

Simply delete the scaling code. Use the opportunity to tidy up a bit.

See #1273.

### DIFF
--- a/client/fonts.cpp
+++ b/client/fonts.cpp
@@ -129,20 +129,11 @@ void fcFont::setFont(const QString &name, const QFont &qf)
  */
 void configure_fonts()
 {
-  int max, default_size;
   QStringList sl;
   QString font_name;
-  const QList<QScreen *> screens = QGuiApplication::screens();
-  const QScreen *screen = screens.at(0);
-  qreal logical_dpi = screen->logicalDotsPerInchX();
-  qreal physical_dpi = screen->physicalDotsPerInchX();
-  qreal screen_size = screen->geometry().width() / physical_dpi + 5;
-  qreal scale = (physical_dpi * screen_size / (logical_dpi * 27))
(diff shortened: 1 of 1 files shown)


jwrober:
So this is better. Any way to make the font closer to 12 to 14pt, vs 16? Also I don't think help label has a default. Why is it size 9?

image

What about using more generic OS non-specific font names like `monospace` and `sans serif`. I see them available in Linux and Windows. If we are to ship font files, I suggest the Liberation set from LibreOffice. They are GNU v3 I think.

jwrober (changes requested):
See comment

jwrober:
> What about using more generic OS non-specific font names like `monospace` and `sans serif`. I see them available in Linux and Windows.

I stand corrected, these are not in Windows (at least not in a fresh install of Win10) so if we want more control, might want to think about shipping some font files. At least per wikipedia, Liberation fonts are OSS - https://en.wikipedia.org/wiki/Liberation_fonts

lmoureaux:
I'm not pretending to fix all font issues in this PR, just the super large fonts that appeared in some instances. I consider this a bug fix, even if related to https://github.com/longturn/freeciv21/issues/1273, since the defaults weren't applied correctly. Changing the defaults should be another PR IMHO. To bundle fonts, we'd need to change the Windows and Mac installers, which is an even wider scope.

jwrober (approved):

PR #1282, opened 2022-08-15
Bump top bar icons to 22px

### LINKED ISSUE #1254: Increase the size of the icons on the buttons on the top bar
We have some really great icons on the buttons on the top bar. I would like to see them larger to make it more pronounced. This is a sample of what they look like now:

image

Change the icons so they fill most (~80%) of the height of the button. I think the font size does not need to change.

None

**Additional context**
None

lmoureaux:
Closes #1254.

### DIFF
--- a/client/top_bar.cpp
+++ b/client/top_bar.cpp
@@ -205,6 +205,7 @@ top_bar_widget::top_bar_widget(const QString &label, const QString &pg,
   setToolButtonStyle(Qt::ToolButtonTextBesideIcon);
   setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
   setContextMenuPolicy(Qt::CustomContextMenu);
+  setIconSize(QSize(22, 22));
 
   timer = new QTimer;
   timer->setSingleShot(false);


jwrober:
What is so magical about 22px? Where do we set the height of the top bar? At 22px, the icons are still quite small. I was looking for the icons to be much larger - e.g. make them much more prominent and fill a large portion of the height of the top bar. Even with the gui as small as I can get it on my laptop, I still think the icons could be bigger

image

jwrober (changes requested):
See comment

lmoureaux:
The height of the top bar is a function of the font size. 22px is a about the height of two lines of text at a size of 12px. Making icons taller also implies making them wider, without changing the horizontal space between the icon and the text, so this starts to look strange if you make them very large. 22px looked "right" on my display.

jwrober (approved):
I played around with the value on my local. 30px seems like a good number too, but I'm not going to quibble over 8px. This is still better than it was.

lmoureaux:
Let's revisit after #1287

lmoureaux:
So what do you think with the new colors?

jwrober:
> So what do you think with the new colors?

I still think they could be bigger. I do still like 30px. I also see something weird with the nations view button. The icon is missing or oddly small for some of the themes

image
image

lmoureaux:
> I also see something weird with the nations view button. The icon is missing or oddly small for some of the themes

It's a fixed-size png, not a scalable icon

PR #1283, opened 2022-08-15
Adding LTT ruleset to Freeciv 21

wieder-fi:
The initial version. The server settings probably need some work. There are some longturn.net 2.6 specific patches on the .serv file and it's not known if all of those are needed or if they have been ported.

### DIFF
--- a/data/LTT.serv
+++ b/data/LTT.serv
@@ -1,0 +1,183 @@
+rulesetdir LTT
+
+# map
+set mapsize="PLAYER"
+set gameseed "0"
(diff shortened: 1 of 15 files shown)


lmoureaux:
Thank you! I think we want to keep LTT either in `games` or (better) in its own repository. Assigned @pranavtifr because he already has a significant part of it.

lmoureaux:
[LTT-LTX](https://github.com/longturn/LTT-LTX) is the new home of LTT and LTX.

PR #1286, opened 2022-08-15
Add support for stable releases

### LINKED ISSUE #1252: AutoRevision needs to support stable releases
When we tag a release, a process call AutoRevision runs to update the Freeciv21 code with the tagged version automatically. Right now we have been tagging unstable Alpha and Beta releases and are close to having a stable release.

The general rule of a stable release is the tag does not have a dash (`-`) in with a `patch` release text provided.  Today our releases are like this: `v3.0-beta.3` and AutoRevision parses that out a `v3.0.base10(git hash)-beta.3`. We need AutoRev to recognize that if the tag string ends at `.0` and does not have a dash (`-`) with subsequent text that this is a stable release. When that is detected then AutoRevision only gives us a version of `vX.0` with not `base10(git hash)` or anything else.

It was discussed to potentially leave the `base10(git hash)` as part of any version, however the developers decided for stable releases we don't want that there, only for unstable point releases.

**Additional context**
We are near a full v3.0!

jwrober:
Closes #1252


lmoureaux (changes requested):
I get the following in `build/utility/fc_version.h`:
```cpp
#ifndef VERSION_STRING
#define VERSION_STRING "3.0.0.0-Stable"
#endif
```
This should be just `3.0`.

jwrober:
That is not possible. The build fails if not all 5 variables are populated with something. When we only tag to a major and minor version, I set static values to populate something in the variables to get the build to complete. 

If there is a way to get around this limitation in the code somewhere, I could attempt to change it or you could push it to my branch to include here.

lmoureaux:
> If there is a way to get around this limitation in the code somewhere, I could attempt to change it or you could push it to my branch to include here.

Branch updated, please review my changes :sweat_smile:

jwrober:
> Branch updated, please review my changes

You broke it, lol. Need to update CMakeLists.txt is the root.

lmoureaux:
> You broke it, lol. Need to update CMakeLists.txt is the root.

Done

jwrober:
I ran a subsequent test of all of the PR via the eval_pr script. I tagged my local with a stable tag and was able to build the whole thing and get the string @lmoureaux wanted. I think we can merge this, just need to remove the "changes requested" on the review.

lmoureaux (approved):

PR #1303, opened 2022-08-19
Docs: add list of game interface elements that we want to capitalize …

jwrober:
…to style guide

### DIFF
--- a/docs/Contributing/style-guide.rst
+++ b/docs/Contributing/style-guide.rst
@@ -1,3 +1,7 @@
+..
+    SPDX-License-Identifier: GPL-3.0-or-later
+    SPDX-FileCopyrightText: 2022 James Robertson <jwrober@gmail.com>
+
 Documentation Style Guide
 *************************
 
@@ -159,8 +163,10 @@ Capitalization
     Below is a more exhaustive list of game specific terms that we want to capitalize. The list can be used
     in a find/replace exercise:
 
-    :strong:`Game Interace`:
-      * Advance (as in Technology Advance)
+    :strong:`Game Interface`:
+
+    .. hlist::
+
       * Airbase
       * Base
       * Border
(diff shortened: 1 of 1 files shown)


daavko (commented):
I don't think we should be listing prefixes here. Things like `Special (also covers Specialist)`, `Entertain(er)`, `Prod(uction)`. IMO this should be just a plain list of terms that are capitalized, without explicitly listed forms of the word.

jwrober:
> I don't think we should be listing prefixes here. Things like `Special (also covers Specialist)`, `Entertain(er)`, `Prod(uction)`. IMO this should be just a plain list of terms that are capitalized, without explicitly listed forms of the word.

I did the prefixes with notes to aid another person in using find/replace. That is the method I used while updating `helpdata.txt` and the `tutorial.sav` scenario with the new capitalization rules. Do you have another suggestion?

daavko:
I'd suggest making this just a list of in-game terms that are capitalized, organized by category, since it's just a style guide. There could, of course, be a "useful list for find/replace for those terms", but I'd put that on a separate page. Or a collapsible section under the main list, if that can be done.

docs/Contributing/style-guide.rst
```diff
@@ -156,6 +156,113 @@ Capitalization
     This page provides a good overview of Title Case capitalization rules for US English:
     https://www.grammarly.com/blog/capitalization-rules/
 
+    Below is a more exhaustive list of game specific terms that we want to capitalize. The list can be used
+    in a find/replace exercise:
+
+    :strong:`Game Interace`:
```
lmoureaux, line 162: Interface

docs/Contributing/style-guide.rst
```diff
@@ -156,6 +156,113 @@ Capitalization
...
+      * Advance (as in Technology Advance)
+      * Airbase
+      * Base
+      * Border
+      * Bulbs
+      * Buoy
+      * Button (a named button)
+      * Chat
+      * Cit(y or ies)
```
lmoureaux, line 171: City/Cities

docs/Contributing/style-guide.rst
```diff
@@ -156,6 +156,113 @@ Capitalization
...
+      * Dialog (a named dialog)
+      * Embass(y or ies)
```
lmoureaux, line 173: Embassy/Embassies

docs/Contributing/style-guide.rst
```diff
@@ -156,6 +156,113 @@ Capitalization
...
+      * Fort(ress)
```
lmoureaux, line 174: Fort and Fortress

docs/Contributing/style-guide.rst
```diff
@@ -156,6 +156,113 @@ Capitalization
...
+      * Improvement
+      * Map
+      * Menu (a named menu)
+      * Messages
+      * Minimap
+      * Small Great Wonder (only with Wonders)
+      * Tax(es) Rate(s)
```
lmoureaux, line 181: "Taxes Rates", really? Maybe this should be two items

lmoureaux (commented):
I agree with Daavko that it would be more clear to skip the prefix thing entirely and just list the variants when it makes sense. I was actually confused the first time and it's probably going to confuse non-programmers even more (we're used to this kind of reasoning from regexes).

Some of these terms, like Prod, may sometimes be abbreviated. In such a case the abbreviation could be listed next to it.

I think this would be better rendered as one or more [hlists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=hlists#directive-hlist) with subtitles inbetween (HTML gives us 6 heading levels, let's not be shy on using them).

PR #1304, opened 2022-08-21
Add a new documentation page about Freeciv21 on servers

lmoureaux:
The intent of this page is to help people set up their own server. It should be an introduction with pointers to the relevant parts of the upcoming server manual.

### DIFF
--- a/docs/General/on-the-server.rst
+++ b/docs/General/on-the-server.rst
@@ -5,23 +5,22 @@
 Freeciv21 on the Server
 ***********************
 
-In order to play Freeciv21 on the network, one machine needs to act as the *server*: the state of
+In order to play Freeciv21 over the network, one machine needs to act as the *server*: the state of
 the game lives on the server and players can connect to play. There are many ways to operate such a
-server; this page gives an introduction to the topic.
+server. This page gives an introduction to the topic.
 
 In order to run a server, you need a computer that other players can reach over the network. If
-you're planning to play at home, this can usually be any machine connected to your WiFi or local
+you are planning to play at home, this can usually be any machine connected to your WiFi or local
 cabled network. Otherwise, the easiest is to rent a small server from a hosting provider. We
 strongly recommend that you choose a Linux-based server, as that is what we have experience with.
 You will need the ability to run your own programs on the server, so SSH access is a must. Apart
-from that, Freeciv21 is quite light on resources so you will hardly hit the limits of even the
+from that, Freeciv21 is quite light on resources, so you will hardly hit the limits of even the
 cheapest options.
 
 Whether you choose to use your own machine or to rent one, the basic principle of operating a
 server is the same: you need to run a program called ``freeciv21-server`` for as long as the game
 will last. This program will wait for players to connect and handle their moves in the exact same
-way as in a single player game. In fact, Freeciv21 always uses a server, even when there is only
-one player!
+way as in a single player game. In fact, Freeciv21 uses an internal server for single player games!
 
 Installation
 ============
@@ -34,23 +33,23 @@ for packaging reasons, but will not be used.
 Starting the Server
 ===================
 
-Once Freeciv21 is installed, it's immediately ready to run a simple server, which one can do by
+Once Freeciv21 is installed, it is immediately ready to run a simple server, which one can do by
 running the following command:
 
 .. code-block:: sh
 
-    freeciv21-server
+    $ freeciv21-server
 
-This starts a server listening on the port traditionally used for Freeciv21, 5556. You can provide
+This command will start a server on the default port for Freeciv21, which is 5556. You can provide
 a custom port number by passing it to the ``-p`` argument:
 
 .. code-block:: sh
 
-    freeciv21-server -p 5000
+    $ freeciv21-server -p 5000
 
-A server started with this command can be reached by pointing the game to port 5000 of your domain
-name. We suggest to start the server from within a terminal multiplexer such as ``tmux`` or
-``screen``, which will let it run independently.
+A server started with this command can be reached by pointing the game to port 5000 of your of your
+machine's IP address or hostname. We suggest to start the server from within a terminal multiplexer
+such as ``tmux`` or ``screen``, which will let it run independently.
 
 User Permissions
 ================
@@ -83,8 +82,8 @@ the ``cmdlevel`` command::
 
     cmdlevel info new
 
-This command would grant access level ``info`` to any newly connecting player. A few more options
-are available; please refer to the documentation of ``cmdlevel`` for more information.
+This command grants access level ``info`` to any newly connecting player. A few more options are
+available; please see ``help cmdlevel`` for more information.
 
 .. note::
     The ``take`` and ``observe`` commands require access level ``info`` only. Their use can be
--- a/docs/General/on-the-server.rst
+++ b/docs/General/on-the-server.rst
@@ -18,3 +18,11 @@ server is the same: you need to run a program called ``freeciv21-server`` for as
 will last. This program will wait for players to connect and handle their moves in the exact same
 way as in a single player game. In fact, Freeciv21 always uses a server, even when there is only
 one player!
+
+Installation
+============
+
+Installing Freeciv21 on a server is done the normal way, as documented in :doc:`/General/install`.
+Because the official packages come with the complete game, they will pull a number of dependencies
+that are normally not used on servers (for instance, a display server). This software is installed
+for packaging reasons, but will not be used.
--- a/docs/General/on-the-server.rst
+++ b/docs/General/on-the-server.rst
@@ -26,3 +26,24 @@ Installing Freeciv21 on a server is done the normal way, as documented in :doc:`
 Because the official packages come with the complete game, they will pull a number of dependencies
 that are normally not used on servers (for instance, a display server). This software is installed
 for packaging reasons, but will not be used.
+
+Starting the Server
+===================
+
+Once Freeciv21 is installed, it's immediately ready to run a simple server, which one can do by
+running the following command:
+
+.. code-block:: sh
+
+    freeciv21-server
+
+This starts a server listening on the port traditionally used for Freeciv21, 5556. You can provide
+a custom port number by passing it to the ``-p`` argument:
+
+.. code-block:: sh
+
+    freeciv21-server -p 5000
+
+A server started with this command can be reached by pointing the game to port 5000 of your domain
+name. We suggest to start the server from within a terminal multiplexer such as ``tmux`` or
+``screen``, which will let it run independently.
--- a/docs/General/on-the-server.rst
+++ b/docs/General/on-the-server.rst
@@ -51,3 +51,42 @@ a custom port number by passing it to the ``-p`` argument:
 A server started with this command can be reached by pointing the game to port 5000 of your domain
 name. We suggest to start the server from within a terminal multiplexer such as ``tmux`` or
 ``screen``, which will let it run independently.
+
+User Permissions
+================
+
+Freeciv21 supports several access levels for players connected to a server, restricting which
+commands they are allowed to run:
+
+``none``
+    The user may not issue any command.
+
+``info``
+    The user may use informational commands only.
+
+``basic``
+    The user may use informational commands as well as commands affecting the game. Commands
+    affecting the game start a vote if more than one user is connected.
+
+``ctrl``
+    Same as ``basic``, but the vote is bypassed for commands affecting the game.
+
+``admin``
+    May use any command, except for ``quit``, ``rfcstyle``, and ``write``. This includes
+    potentially destructive commands such as ``save`` and ``fcdb`` --- use with care.
+
+``hack``
+    May use all commands without restriction.
+
+By default, users connected to your server have access level ``basic``. This can be changed using
+the ``cmdlevel`` command::
+
+    cmdlevel info new
+
+This command would grant access level ``info`` to any newly connecting player. A few more options
+are available; please refer to the documentation of ``cmdlevel`` for more information.
+
+.. note::
+    The ``take`` and ``observe`` commands require access level ``info`` only. Their use can be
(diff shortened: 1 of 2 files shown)


docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
+..
+    SPDX-License-Identifier: GPL-3.0-or-later
+    SPDX-FileCopyrightText: 2022 louis94 <m_louis30@yahoo.com>
+
+Freeciv21 on the Server
+***********************
+
+In order to play Freeciv21 on the network, one machine needs to act as the *server*: the state of
```
daavko, line 8: "over the network", not "on the network"

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+strongly recommend that you choose a Linux-based server, as that is what we have experience with.
+You will need the ability to run your own programs on the server, so SSH access is a must. Apart
+from that, Freeciv21 is quite light on resources so you will hardly hit the limits of even the
+cheapest options.
+
+Whether you choose to use your own machine or to rent one, the basic principle of operating a
+server is the same: you need to run a program called ``freeciv21-server`` for as long as the game
+will last. This program will wait for players to connect and handle their moves in the exact same
+way as in a single player game. In fact, Freeciv21 always uses a server, even when there is only
+one player!
```
daavko, line 24: I'd rephrase this as "In fact, Freeciv21 uses an internal server for single player games!", as it's referencing this. "Even when there is only one player" might be confused for a single player being connected to a regular server. Then again, this duplicates the "single player game" term, but I think it's fine here.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+
+``hack``
+    May use all commands without restriction.
+
+By default, users connected to your server have access level ``basic``. This can be changed using
+the ``cmdlevel`` command::
+
+    cmdlevel info new
+
+This command would grant access level ``info`` to any newly connecting player. A few more options
```
daavko, line 86: I'd suggest "will" rather than "would" here, as it's guaranteed to happen.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+are available; please refer to the documentation of ``cmdlevel`` for more information.
```
daavko, line 87: "documentation for", not "documentation of"

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Freeciv21 on the Server
+***********************
+
+In order to play Freeciv21 on the network, one machine needs to act as the *server*: the state of
+the game lives on the server and players can connect to play. There are many ways to operate such a
+server; this page gives an introduction to the topic.
+
+In order to run a server, you need a computer that other players can reach over the network. If
+you're planning to play at home, this can usually be any machine connected to your WiFi or local
```
daavko, line 13: "you are", as per style guide

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Installing Freeciv21 on a server is done the normal way, as documented in :doc:`/General/install`.
+Because the official packages come with the complete game, they will pull a number of dependencies
+that are normally not used on servers (for instance, a display server). This software is installed
+for packaging reasons, but will not be used.
+
+Starting the Server
+===================
+
+Once Freeciv21 is installed, it's immediately ready to run a simple server, which one can do by
```
daavko, line 37: "it is", as per style guide

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+
+This starts a server listening on the port traditionally used for Freeciv21, 5556. You can provide
+a custom port number by passing it to the ``-p`` argument:
+
+.. code-block:: sh
+
+    freeciv21-server -p 5000
+
+A server started with this command can be reached by pointing the game to port 5000 of your domain
+name. We suggest to start the server from within a terminal multiplexer such as ``tmux`` or
```
daavko, line 52: I'm not sure how strongly I feel about this, but I'd use "of your machine's IP address or hostname" instead.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+    The ``take`` and ``observe`` commands require access level ``info`` only. Their use can be
+    restricted using the ``allowtake`` server option or, in more advanced setups, using the
+    ``user_take`` :doc:`fcdb </Coding/fcdb>` hook.
+
+Authentication
+==============
+
+By default, Freeciv21 servers are open to anyone without any sort of authentication. It is
+possible to make Freeciv21 ask for a password. There is also a built-in registration procedure
+with which one can create an account and set a password.
```
daavko, line 99: If there's a built-in registration procedure, it should be either documented here, or linked to from here.

daavko (changes requested):

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+Freeciv21 on the Server
+***********************
+
+In order to play Freeciv21 on the network, one machine needs to act as the *server*: the state of
+the game lives on the server and players can connect to play. There are many ways to operate such a
+server; this page gives an introduction to the topic.
+
+In order to run a server, you need a computer that other players can reach over the network. If
+you're planning to play at home, this can usually be any machine connected to your WiFi or local
+cabled network. Otherwise, the easiest is to rent a small server from a hosting provider. We
```
jwrober, line 14: Instead of "we", I recommend "the Longturn Community". It is technically the same thing, but gives more definition of who "we" are.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+strongly recommend that you choose a Linux-based server, as that is what we have experience with.
+You will need the ability to run your own programs on the server, so SSH access is a must. Apart
```
jwrober, line 16: SSH is a program, so wrap in two backticks or use :file: directive.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+from that, Freeciv21 is quite light on resources so you will hardly hit the limits of even the
```
jwrober, line 17: Need a comma after "resources" and before "so"

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+cheapest options.
+
+Whether you choose to use your own machine or to rent one, the basic principle of operating a
+server is the same: you need to run a program called ``freeciv21-server`` for as long as the game
```
jwrober, line 21: Either change the colon to a semi-colon or break up into two sentences.

docs/General/on-the-server.rst
jwrober: So this interesting. On one hand with the client manual we are thinking of moving away from the explicit understanding of the "game" being client/server in nature, but then here we go and make it explicit. The FAQ does this as well. We might need to discuss how we want to handle this. I am personally fine with a consistent message that fc21 follows a client/server model. We can explain what that means for users who may not be super-technical and still accomplish our goals with documentation.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+will last. This program will wait for players to connect and handle their moves in the exact same
+way as in a single player game. In fact, Freeciv21 always uses a server, even when there is only
+one player!
+
+Installation
+============
+
+Installing Freeciv21 on a server is done the normal way, as documented in :doc:`/General/install`.
```
jwrober, line 29: I have always used relative paths for :doc: or other location based directives such as :image: or :figure:

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Starting the Server
+===================
+
+Once Freeciv21 is installed, it's immediately ready to run a simple server, which one can do by
+running the following command:
+
+.. code-block:: sh
+
+    freeciv21-server
```
jwrober, line 42: I have made a personal habit of giving a non-elevated prompt of "`$`" in front of these types of commands. This is to denote that you don't have to have an elevated level of access to run `freeciv21-server`.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
...
+
+This starts a server listening on the port traditionally used for Freeciv21, 5556. You can provide
```
jwrober, line 44: suggestion: This command will start a server on the default port for Freeciv21, which is 5556.

docs/General/on-the-server.rst
jwrober: I agree

docs/General/on-the-server.rst
jwrober: Where do we find such documentation?

docs/General/on-the-server.rst
jwrober: It is not documented anywhere "yet". My personal expectation was to include in the server manual that has not been written yet. Not sure how we get around the ambiguity here.

jwrober (commented):
I'm not going to block this being merged. However, I do think that this page may cause some confusion for readers and we could do more cross-doc links to other pages as time progresses. My thinking with a forth-coming server manual was to include some game "recipes" to help potential server admins use the docs as a guide.

docs/General/on-the-server.rst
lmoureaux: I'm assuming we'll have a page describing all server commands that will be linked. In the meantime, consider that this is referring to `/help cmdlevel`.

jwrober: I like that as an interim solution

docs/General/on-the-server.rst
lmoureaux: I think in this case there is no way around mentioning the client/server, but I still don't use the word *client*. I wrote this paragraph as if the reader had never heard of the server before, so they (sight) can understand it here since it's usually not mentioned.

The FAQ could contain an entry *What's the "server"?* or similar. I agree it would be good to discuss the structure of the upcoming server manual before it gets written :D

docs/General/on-the-server.rst
lmoureaux: Should I change it? This imho is more stable if we decide to move the page somewhere else

jwrober: The one use case I have for "yes" is in the client manual I call an image from the `dist` directory that is technically outside of the `docs` "area" per se. So my concern would be that using absolute paths vs relative paths may get us in trouble down the road if we change directions. That being said, if we do move to a different platform for our documentation, that implies a lot of refactoring to make it work on new platform. I personally prefer consistency, but won't block due to differences of opinion.

lmoureaux:
> My thinking with a forth-coming server manual was to include some game "recipes" to help potential server admins use the docs as a guide.

I followed a similar course of thought. This was prompted by kreis' recent questions about setting up a public server*, and some of the content clearly originates from that discussion (the discussion about dependencies). I agree that this prompts for more pages, in particular a page documenting server command line options and another for game commands (which shouldn't mention the server, since they can be used from the client as well). The fcdb page also needs to be improved and taken out of `Coding`.

At the same time I wanted to keep the PR small to allow for easier review, which is why some content feels "missing".

*Hence the title "Freeciv21 *on the* Server" and not "Server Manual". This page is higher level than purely technical docs listing all available options, and omits all but the most relevant content.

docs/General/on-the-server.rst
lmoureaux: It's a function in fcdb, but the documentation we have for that is rather poor. It's intended to be linked in the last paragraph.

docs/General/on-the-server.rst
```diff
@@ -0,0 +1,109 @@
+..
+    SPDX-License-Identifier: GPL-3.0-or-later
+    SPDX-FileCopyrightText: 2022 louis94 <m_louis30@yahoo.com>
+
+Freeciv21 on the Server
+***********************
+
+In order to play Freeciv21 on the network, one machine needs to act as the *server*: the state of
+the game lives on the server and players can connect to play. There are many ways to operate such a
+server; this page gives an introduction to the topic.
```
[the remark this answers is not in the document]
lmoureaux: I like to use the `;` as a weaker full stop but doesn't matter

docs/General/on-the-server.rst
lmoureaux: SSH stands for *Secure Shell* and is not usually set in monospace in its uppercase form. It's also not a `:file:` because the file is called `ssh` (lowercase).

docs/General/on-the-server.rst
lmoureaux: Why? This that you need to run `freeciv21-server` is the continuation of "the basic principle is the same". I think using a full stop or a semicolon would lose this nuance.

docs/General/on-the-server.rst
lmoureaux: I added a note to the first paragraph: "This page gives an introduction to the topic, guided by the experience gathered by running the [Longturn](https://longturn.net) website."

docs/General/on-the-server.rst
jwrober: I like the idea of another question in the FAQ and we def want to discuss more about change in name of the current client manual as part of upcoming server manual.

docs/General/on-the-server.rst
jwrober: Agreed. OK.

docs/General/on-the-server.rst
jwrober: OK

docs/General/on-the-server.rst
jwrober: Yea that makes sense.

docs/General/on-the-server.rst
jwrober: or maybe just a comma instead? IDK, the colon feels wrong, but I can't put my finger on it directly. I am also ok to leave as is.

jwrober (approved):

PR #1305, opened 2022-08-21
List small wonders in the client, but hide them by default

### LINKED ISSUE #318: Everyone can see everyone's small wonders
A lua script is enough. This shouldn't be sent unless maybe on embassies.

> This game currently has a bug/hidden dev backdoor. 
>
> You can actually find all the small wonders any player in the world has using lua.
> `player:has_wonder()` function receives small wonders too while it is supposed to get only great wonders.
> It was reported by Ignatus and would probably br fixed soon upstream.

lmoureaux:
The client has always known which small wonders any player had built (#318). This adds code to show this in the client. For backward compatibility, the patch also hides small wonders by default, keeping only great wonders. This can be controlled using a new effect `Wonder_Visible`. All shipped rulesets use the old way.

Closes #318.

### DIFF
--- a/ai/default/daieffects.cpp
+++ b/ai/default/daieffects.cpp
@@ -574,6 +574,7 @@ adv_want dai_effect_value(struct player *pplayer, struct government *gov,
   case EFT_STEALINGS_IGNORE:
   case EFT_MAPS_STOLEN_PCT:
   case EFT_UNIT_SHIELD_VALUE_PCT:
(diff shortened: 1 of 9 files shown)


lmoureaux:
Figured I'd make some more changes here, hold on

lmoureaux:
updated

jwrober (approved):
Looks great in my local test of a saved game.

PR #1307, opened 2022-08-22
Switch to the network page when autoconnecting to an auth-enabled server

### LINKED ISSUE #540: Autoconnecting to a auth-enabled server doesn't show the password prompt
Try:
```
$ freeciv21-client -a fc21://louis94@longturn.net:5050
```
This should allow to connect to LTEx24Avi.

What actually happens is that the server asks for the password, but the client fails to recognize it. Hence it doesn't switch to `PAGE_NETWORK` and doesn't enable the password field.

Passing the password on the command line works.

lmoureaux:
Closes #540.

### DIFF
--- a/client/connectdlg.cpp
+++ b/client/connectdlg.cpp
@@ -37,6 +37,7 @@ void close_connection_dialog()
 void handle_authentication_req(enum authentication_type type,
                                const char *message)
 {
+  set_client_page(PAGE_NETWORK);
   qobject_cast<page_network *>(king()->pages[PAGE_NETWORK])
       ->handle_authentication_req(type, message);
 }


jwrober:
@lmoureaux How would you suggest I test this?

lmoureaux:
Start games, stop them, connect to other games, run LAN scans, start a game and change your mind for a scenario, oh no you wanted to load a save instead, ...

jwrober:
Actually, I will follow the directions in PR #1304 when it is published to setup an auth enabled server and test against that.

jwrober (approved):
Interesting how one line can fix something. I was able to setup an auth enabled server and recreate the bug and then it went away with this patch.

Issue #1179, opened 2022-08-01
Define, document and implement a naming convention


lmoureaux:
Now that we are converting more and more stuff to classes, we need to think a bit about the naming convention. We wanted to use CamelCase to fit with Qt's convention which we use everywhere. I'm approving this, but is the naming convention finalized. If yes, we need to document it in contributor's guidelines.

_Originally posted by @pranavtifr in https://github.com/longturn/freeciv21/pull/1171#discussion_r934194868_

lmoureaux:
`clang-tidy` can do it:
https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html
https://sarcasm.github.io/notes/dev/clang-tidy.html

PR #1312, opened 2022-08-23
Restore goto into the unknown

### LINKED ISSUE #890: Restore goto into the unknown
**Describe the bug**

This feature was removed in #824 and does nothing. Restore it (there's an option in the config dialog).

lmoureaux:
Add a flag to the path finder class to control whether it should include
unknown tiles in the search. Bind it to the corresponding gui option when
creating the finders.

This restores the ability to produce paths going through unknown parts of the
map, which is useful in some circumstances.

Closes #890.

### DIFF
--- a/common/path_finder.h
+++ b/common/path_finder.h
@@ -71,6 +71,8 @@ private:
     const detail::vertex
         initial_vertex; ///< The starting point in the search graph.
 
+    bool unknown_tiles_allowed = false;
+
     // Storage for Dijkstra's algorithm.
     // In most cases, a single vertex will be stored for a given tile. There
     // are situations, however, where more vertices are needed. This is for
@@ -101,6 +103,7 @@ private:
     void attempt_action_move(detail::vertex &source);
 
     bool run_search(const destination &destination);
+    void reset();
   };
 
 public:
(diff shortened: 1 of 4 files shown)


common/path_finder.h
```diff
@@ -110,6 +113,14 @@ class path_finder {
 
   inline path_finder &operator=(path_finder &&other);
 
+  /// Returns whether goto into the unkown is allowed.
+  bool are_unknown_tiles_allowed() const
```
pranavtifr, line 117: Just a small nitpick/comment/suggestion should this be called get_unknown_tiles_allowed() to be consistent with the getter-setter idiom in cpp ? I'm also ok if we consistently name all setters with a question like this. Then it needs to go into coding style and the lot, getter-setter is more standard.

pranavtifr, line 117: Just a small nitpick/comment/suggestion should this be called get_unknown_tiles_allowed() to be consistent with the getter-setter idiom in cpp ? I'm also ok if we consistently name all setters with a question like this. Then it needs to go into coding style and the lot, getter-setter is more standard.

pranavtifr:
Apart from that, looks ok and works ok.

common/path_finder.h
lmoureaux: I'm using the Qt naming convention here: boolean getters start with `is` and others don't have a specific prefix. Someone should push #1179 forward...

PR #1320, opened 2022-08-25
Remove libclient and rewrite the client CMakeLists.txt

lmoureaux:
Since the help text generation was moved to `libcommon` in https://github.com/longturn/freeciv21/pull/926, civmanual no longer needs libclient. The distinction between libclient and the client itself is thus moot.

The CMakeLists.txt for the client/ subdir had been written before the client
code was consolidated into client/ (we still had client/gui-qt and a few
others). It also had the requirement of a separate libclient, which is no
longer needed.

Rewrite the file to create a single target (apart from source generation) with
all sources added when needed. Use the opportunity to group include directories
and linked libraries together.


lmoureaux:
Of course wasm had to fail :cry:

jwrober (approved):
Looks good to me. Reduced the number of build steps as well.

Issue #916, opened 2022-02-18
Allow opening the client by clicking on fc21:// links


lmoureaux:
**Is your feature request related to a problem? Please describe.**

Many applications can be opened by clicking a link in a Web browser. This is, for instance, the case of Zoom, Teams, and similar communication clients. It would be great to allow the same for Freeciv21. A button could be added to the Longturn website, that would open the client and autofill all connection settings, which would relieve us from the recurring question "how do I connect?".

**Describe the solution you'd like**

On Linux, URL handlers are set in `.desktop` files [[1](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html), [2](https://unix.stackexchange.com/questions/497146/create-a-custom-url-protocol-handler)]. Thus we would need to add `%u` to `Exec` and add the following key:
```
MimeType=x-scheme-handler/fc21
```

For Windows, one needs to edit the Registry as explained in Ref. [[3](https://docs.microsoft.com/en-us/windows/win32/search/-search-3x-wds-ph-install-registration#installing-and-registering-a-protocol-handler)]. Mac OS X is left as an exercise to the implementer.

**Describe alternatives you've considered**

Doing nothing, but how lame this would be!

jwrober:
I found out how to do it on macOS - https://stackoverflow.com/questions/471581/how-to-map-a-custom-protocol-to-an-application-on-the-mac. Now just need to find a Mac user to help test.

PR #1335, opened 2022-08-28
Support fc21:// links on Linux

jwrober:
Part 1 of 3 in #916

PR #1339, opened 2022-08-28
Support fc21:// links on Windows

jwrober:
Part 2 of 3 in #916


jwrober:
To test this, you need to create a package, run a local server and use an html file similar to what is in #1335

jwrober:
I also fixed a delete registry issue I found while testing

lmoureaux (approved):

PR #1343, opened 2022-08-29
Update detailed combat info widget

### LINKED ISSUE #1219: Enhance Win/Loss to combat pop-up
When a battle between two nation's units occurs, the client will display something like this:

image

A visually impaired person, such as one that is red/green color blind will not see that the winner of the combat has a green shading in the box and the loser will have a red shading in the box. It is also not clear that the negative number value in the upper right is for hit points.

In the lower right corner of the box and a white W or L to denote win or loss and then add "HP" in front of the value in the upper right to make what happened more clear.

None

**Additional context**
None

jwrober:
Closes #1219

### DIFF
--- a/client/hudwidget.cpp
+++ b/client/hudwidget.cpp
@@ -1794,8 +1794,8 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   QFont f = fcFont::instance()->getFont(fonts::default_font);
   QString ahploss, dhploss;
 
-  // TRANS: HP - Hit Points
   if (att_hp_loss > 0) {
+    // TRANS: HP - Hit Points
     ahploss = QString(_("-%1 HP")).arg(att_hp_loss);
   } else {
     ahploss = QString(_("%1 HP")).arg(att_hp_loss);
--- a/client/hudwidget.cpp
+++ b/client/hudwidget.cpp
@@ -1682,8 +1682,10 @@ hud_unit_combat::hud_unit_combat(int attacker_unit_id, int defender_unit_id,
   def_hp_loss = defender->hp - def_hp;
   if (defender_hp <= 0) {
     center_tile = attacker->tile;
+    att_win = true;
   } else {
     center_tile = defender->tile;
+    def_win = true;
   }
   init_images();
 }
@@ -1792,15 +1794,16 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   QFont f = fcFont::instance()->getFont(fonts::default_font);
   QString ahploss, dhploss;
 
+  // TRANS: HP is Hip Point
   if (att_hp_loss > 0) {
-    ahploss = "-" + QString::number(att_hp_loss);
+    ahploss = "-" + QString::number(att_hp_loss) + " HP";
   } else {
-    ahploss = QStringLiteral("0");
+    ahploss = QStringLiteral("0") + " HP";
   }
   if (def_hp_loss > 0) {
-    dhploss = "-" + QString::number(def_hp_loss);
+    dhploss = "-" + QString::number(def_hp_loss) + " HP";
   } else {
-    dhploss = QStringLiteral("0");
+    dhploss = QStringLiteral("0") + " HP";
   }
   f.setBold(true);
 
@@ -1834,17 +1837,24 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   p.setFont(f);
   p.setPen(QColor(Qt::white));
   if (def_veteran) {
-    p.drawText(right,
-               Qt::AlignHCenter | Qt::AlignJustify | Qt::AlignAbsolute,
-               QStringLiteral("*"));
+    p.drawText(right, Qt::AlignBottom | Qt::AlignRight | Qt::AlignAbsolute,
+               QStringLiteral("*Vet*"));
   }
   if (att_veteran) {
-    p.drawText(left, Qt::AlignHCenter | Qt::AlignJustify | Qt::AlignAbsolute,
-               QStringLiteral("*"));
+    p.drawText(left, Qt::AlignBottom | Qt::AlignRight | Qt::AlignAbsolute,
+               QStringLiteral("*Vet*"));
   }
   p.drawText(left, Qt::AlignHorizontal_Mask, ahploss);
   p.drawImage(right, dimg);
   p.drawText(right, Qt::AlignHorizontal_Mask, dhploss);
+  if (def_win) {
+    p.drawText(right, Qt::AlignBottom | Qt::AlignLeft | Qt::AlignAbsolute,
+               QStringLiteral("Win"));
+  }
+  if (att_win) {
+    p.drawText(left, Qt::AlignBottom | Qt::AlignLeft | Qt::AlignAbsolute,
+               QStringLiteral("Win"));
+  }
   p.end();
 }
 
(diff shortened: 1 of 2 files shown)


client/hudwidget.cpp
```diff
@@ -1792,15 +1794,16 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   QFont f = fcFont::instance()->getFont(fonts::default_font);
   QString ahploss, dhploss;
 
+  // TRANS: HP is Hip Point
   if (att_hp_loss > 0) {
-    ahploss = "-" + QString::number(att_hp_loss);
+    ahploss = "-" + QString::number(att_hp_loss) + " HP";
```
lmoureaux, line 1799: It's always better to translate strings in full, in case the order must change (consider currencies, $10 is 10€). The underscore function is the translation:
```suggestion
    ahploss = QString(_("-%1 HP")).arg(att_hp_loss);
```
Same below...

client/hudwidget.cpp
```diff
@@ -1834,17 +1837,24 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   p.setFont(f);
   p.setPen(QColor(Qt::white));
   if (def_veteran) {
-    p.drawText(right,
-               Qt::AlignHCenter | Qt::AlignJustify | Qt::AlignAbsolute,
-               QStringLiteral("*"));
+    p.drawText(right, Qt::AlignBottom | Qt::AlignRight | Qt::AlignAbsolute,
+               QStringLiteral("*Vet*"));
```
lmoureaux, line 1841:
```suggestion
               _("*Vet*"));
```

client/hudwidget.cpp
```diff
@@ -1834,17 +1837,24 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
...
   }
   if (att_veteran) {
-    p.drawText(left, Qt::AlignHCenter | Qt::AlignJustify | Qt::AlignAbsolute,
-               QStringLiteral("*"));
+    p.drawText(left, Qt::AlignBottom | Qt::AlignRight | Qt::AlignAbsolute,
+               QStringLiteral("*Vet*"));
```
lmoureaux, line 1845:
```suggestion
               _("*Vet*"));
```

client/hudwidget.cpp
```diff
@@ -1834,17 +1837,24 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
...
   p.drawImage(right, dimg);
   p.drawText(right, Qt::AlignHorizontal_Mask, dhploss);
+  if (def_win) {
+    p.drawText(right, Qt::AlignBottom | Qt::AlignLeft | Qt::AlignAbsolute,
+               QStringLiteral("Win"));
+  }
+  if (att_win) {
+    p.drawText(left, Qt::AlignBottom | Qt::AlignLeft | Qt::AlignAbsolute,
+               QStringLiteral("Win"));
+  }
```
lmoureaux, line 1857: Can both `def_win` and `att_win` be true simultaneously? Maybe with `Combat_Rounds` or bombarding?

lmoureaux (changes requested):
Some comments without running the code, mainly around translateability.

client/hudwidget.cpp
jwrober: I wasn't sure so I went with a variable for both and added them to the `if (defender_hp <= 0)` block

client/hudwidget.cpp
```diff
@@ -1793,14 +1795,15 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   QString ahploss, dhploss;
 
   if (att_hp_loss > 0) {
-    ahploss = "-" + QString::number(att_hp_loss);
+    // TRANS: HP - Hit Points
+    ahploss = QString(_("-%1 HP")).arg(att_hp_loss);
   } else {
-    ahploss = QStringLiteral("0");
+    ahploss = QString(_("%1 HP")).arg(att_hp_loss);
```
lmoureaux, line 1801: I had a hard time understanding what the two branches of the `if` do. The `true` branch displays the HP loss if it is positive and the second branch is for no HP loss. Assuming that `att_hp_loss` is never negative, we could simplify this beast to:
```cpp
// TRANS: Unit lost %1 hit points in combat. Keep short.
ahploss = QString(_("-%1 HP")).arg(att_hp_loss);
```
And similar for `dhploss`.

client/hudwidget.cpp
```diff
@@ -1834,17 +1837,24 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
   p.setFont(f);
   p.setPen(QColor(Qt::white));
   if (def_veteran) {
-    p.drawText(right,
-               Qt::AlignHCenter | Qt::AlignJustify | Qt::AlignAbsolute,
-               QStringLiteral("*"));
+    p.drawText(right, Qt::AlignBottom | Qt::AlignRight | Qt::AlignAbsolute,
+               _("*Vet*"));
```
lmoureaux, line 1841: What are the stars for? Maybe change the color to something goldy and leave a hint for translators?

client/hudwidget.cpp
```diff
@@ -1834,17 +1837,24 @@ void hud_unit_combat::paintEvent(QPaintEvent *event)
...
-               QStringLiteral("*"));
+    p.drawText(left, Qt::AlignBottom | Qt::AlignRight | Qt::AlignAbsolute,
+               _("*Vet*"));
   }
   p.drawText(left, Qt::AlignHorizontal_Mask, ahploss);
   p.drawImage(right, dimg);
   p.drawText(right, Qt::AlignHorizontal_Mask, dhploss);
+  if (def_win) {
+    p.drawText(right, Qt::AlignBottom | Qt::AlignLeft | Qt::AlignAbsolute,
+               _("Win"));
```
lmoureaux, line 1852: I suggest `Winner` here and/or leave a note for translators (`Win`/`Gagne` would make no sense, we really want `Victor`/`Vainqueur` in French).

lmoureaux (commented):
With the basic things out of the way it was easier to review the logic. I'm convinced that the code works. Some things could be improved or simplified here and there but I won't oppose merging as is.

client/hudwidget.cpp
lmoureaux: Or maybe `Wins`...

client/hudwidget.cpp
jwrober: I mostly just kept what was there and only modified to suit the issue. I wasn't looking to improve/refactor anything. I can look at this, however.

client/hudwidget.cpp
jwrober: As you can see, the original `drawText()` simply showed a star for a vet, which wasn't clear. I can look at an alternate color or simply change to remove the stars.

client/hudwidget.cpp
jwrober: I'll change to "Winner" and add a //TRANS

client/hudwidget.cpp
jwrober: I think we need both sides. 
```  
  if (att_hp_loss > 0) {
    // TRANS: HP - Hit Points
    ahploss = QString(_("-%1 HP")).arg(att_hp_loss);
  } else {
    ahploss = QString(_("%1 HP")).arg(att_hp_loss);
  }
```
If HP loss is greater than 0 (meaning the unit took a hit and lost some HP), then we show a negative number in the text. We then assume in the `else` statement that there was no loss to HP and give a "positive" number that will essentially say "0 HP". I think this is more about the text setup. We don't want to show "-0 HP".

lmoureaux: I kind of assumed that `-0 HP` was fine (would be useful to distinguish lost HP from remaining HP). Otherwise it can be `ahploss = QString(_("%1 HP")).arg(-att_hp_loss)`

lmoureaux (approved):

PR #1345, opened 2022-08-31
Fix killing citizens when can't disband units

### LINKED ISSUE #1332: Assertion pcity->surplus[O_SHIELD]
While playing, I got this:

`Assertion pcity->surplus[O_SHIELD] >= 0 failed`
`Please report this message at https://github.com/longturn/freeciv21/issues`

Savegame and ruleset are attached

(this is actually RAR but Github rejected it so I changed the extension)
[SIM30-fc21.zip](https://github.com/longturn/freeciv21/files/9437462/SIM30-fc21.zip)

(this is a genuine ZIP)
[sim-test.zip](https://github.com/longturn/freeciv21/files/9437465/sim-test.zip)

lmoureaux:
When a city has negative shields and doesn't manage to disband enough offending
units, it starts killing citizens instead, one per free unit upkeep. Except
this wasn't working: the units were always counted as if disbanding them had
succeeded. Then, when killing citizens the shields weren't given back to the
city and the negative shields were causing an assertion in the coinage
code.

Closes #1332.

### DIFF
--- a/server/cityturn.cpp
+++ b/server/cityturn.cpp
@@ -2222,7 +2222,6 @@ static bool city_distribute_surplus_shields(struct player *pplayer,
     unit_list_iterate_safe(pcity->units_supported, punit)
     {
       /* Should we look at punit->upkeep[O_SHIELD] here, instead of the
-      && pcity->surplus[O_SHIELD] < 0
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Took me a bit to recreate the problem. Was not able to after this patch.

PR #1353, opened 2022-09-04
Consolidate all network protocol docs on a single page

lmoureaux:
This PR mostly moves text around, but I also reviewed the new page and deleted the obsolete information I could spot.


daavko (approved):
LGTM, no issues found

Issue #1360, opened 2022-09-07
The path finding code believes that disembarking always uses up full MP


lmoureaux:
**Describe the bug**

Some units (e.g. airplanes) can unload from transports without losing full MP. The path finding code thinks they lose them and displays turn steps incorrectly.

**To Reproduce**
1. Load a fighter on an aircraft carrier (classic ruleset), make sure the fighter has MP left
2. Select the fighter
3. Do `g`
4. Move the mouse around and look at the first goto step (always next to the carrier)

**Expected behavior**

TC marks reflect reality.

**Screenshots**

This Bomber has 4 MP left, there shouldn't be a TC marker next to the Carrier
![image](https://user-images.githubusercontent.com/22327575/188986578-dfe6a969-0342-4849-8d8e-c05b5a745da9.png)

**Platform and version (please complete the following information):**
 - OS: Irrelevant
 - Freeciv21 version: 3.0-beta4+
 - Ruleset/Longturn game (if applicable): `classic`

Issue #1153, opened 2022-07-17
Port Longturn patches - game mechanics fixes


daavko:
This is a breakout issue from #503 to track status better.

This is only for game mechanics fixes that don't have a new server option or a ruleset option associated with them.

**Short summary of patches**
Exploit fixes and bugfixes, see commit messages listed below.

**Patches to port**
Commits listed chronologically, beginning from the oldest one.

- longturn/freeciv@1945cd8f3c - Remove exploits of bouncing units due to terrain change and stack conflicts: * Bouncing does not reset UWT; * Bouncing on a native terrain from non-native (TF applied) zeroes mp with `slow_invasions` set.
- longturn/freeciv@0e98d6216d - Cleanup: Invert `city_build_stuff()` conditional in `update_city_activity()`
- longturn/freeciv@0112a87d5b - Fix buildings being sold unexpectedly with `gold_upkeep_style Mixed`.
- longturn/freeciv@5533fbeaa3 - Fix a crash when a city is destroyed during TC
- longturn/freeciv@61e044c30f - Fix a crash when auto-upgrading unique units

lmoureaux:
I'm working on https://github.com/longturn/freeciv/commit/1945cd8f3c

lmoureaux:
https://github.com/longturn/freeciv/commit/0e98d6216d is irrelevant, https://github.com/longturn/freeciv/commit/0112a87d5b already applied, https://github.com/longturn/freeciv/commit/5533fbeaa3 irrelevant, https://github.com/longturn/freeciv/commit/61e044c30f in #1380

PR #1361, opened 2022-09-07
Port 1945cd8f3ccea261dc8c6ff769e42543aca32a31

lmoureaux:
This has grown quite big...

The original commit https://github.com/longturn/freeciv/commit/1945cd8f3c solved two issues:

* UWT could be bypassed by forced moves on the server side (e.g. terraforming).
* Bounced units when breaking an alliance allowed disembarking without losing MP (related to a now-removed `slow_invasions` server setting)

I think the first issue is better fixed elsewhere, so I have a very simple commit for that.

The second issue looks totally different in Fc21 because unit moves are now controlled by action enablers. After a few failed attempts, I chose to run the fully-fledged path finding algorithm to legal paths for the unit and pick one at random.

### Testing

Here are a few cases where unit "bouncing" happens:
* Upon breaking alliance, units in transports and cities get bounced out
* When civil war is triggered, allied units are kicked out of the civil war nation
* When bribing or exchanging a city supporting a gameloss unit, the King flees away
* When capturing or bribing a unit in a stack, the bribed unit may need to leave the stack
* Terrain changes (terraforming, GW)

### See also

Part of #1153
#1360 is a bug in the path finding code indirectly affecting this

### DIFF
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -1465,7 +1465,10 @@ void request_units_return()
   for (const auto unit : get_units_in_focus()) {
     // Find a path to the closest city
     auto finder = freeciv::path_finder(unit);
-    finder.set_unknown_tiles_allowed(gui_options.goto_into_unknown);
+    if (!gui_options.goto_into_unknown) {
+      finder.set_constraint(
+          std::make_unique<freeciv::tile_known_constraint>(client_player()));
+    }
     if (auto path = finder.find_path(
             freeciv::allied_city_destination(unit->owner))) {
       auto steps = path->steps();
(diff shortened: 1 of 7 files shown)


jwrober (approved):
I was able to do the following use cases:
* Upon breaking alliance, units in transports and cities get bounced out
* When bribing a city supporting a gameloss unit, the King flees away (I bribed, did not test exchange)
* When capturing or bribing a unit in a stack, the bribed unit may need to leave the stack (I bribed, did not capture)
* Terrain changes (terraforming, GW) - still works as always used to.

Issue #1292, opened 2022-08-17
Bundle our own fonts


jwrober:
**Is your feature request related to a problem? Please describe.**
Being a cross-platform application we rely on the available set of fonts across many OS ecosystems. To better control the look at feel of the text in fc21, I think we should look at bundling our own TTF fonts and reference those in the options dialog.

**Describe the solution you'd like**
We need to find an open source licensed font set that supports a wide array of languages and character sets. The Liberation Fonts first written by Red Hat -- https://github.com/liberationfonts/liberation-fonts -- is a potential candidate. They use an open source license and support a wide array of languages. Another option is the GNU FreeFont set -- https://www.gnu.org/software/freefont/ -- that is licensed GPL same as fc21 and also supports a huge array of languages.

This enhancement will require a change to the packages as well to ensure the font files are included.

**Describe alternatives you've considered**
None

**Additional context**
This issue came up while working on a font bug fix in PR #1281

pranavtifr:
Should we really bundle it for linux too ? Cant we add it as a dependency, like `ttf-liberation` or something ?

jwrober:
> Should we really bundle it for linux too ? Cant we add it as a dependency, like `ttf-liberation` or something ?

Yea, we could look at that if/when we come around to doing this PR

lmoureaux:
Linux packaging is up to the packager -- if there is already a package, we should of course use it.

jwrober:
I agree, for the Linux packages we should add dependency. For Windows we will need to add the MSYS2 package for the font set we go with and then include the files into the package. I think for the macOS package, we can go with what we would do with Linux, just add another brew command to the the package.

lmoureaux:
When fonts aren't installed system-wide: https://doc.qt.io/qt-6/qfontdatabase.html#addApplicationFont

PR #1373, opened 2022-09-14
Bundle the freefont package in dist

jwrober:
Part 1 of 2 of #1292. Once this is merged, then I can update the varying packages. Linux will use a package dep, while Windows and macOS will need the files copied to the installer.


lmoureaux:
What do the fonts look like in context?

jwrober:
Good question. With these settings as an example:
image

You get...
image
image
image
image
image

jwrober:
I forgot research tree
image

lmoureaux:
FreeMono feels out-of-place. Are the others bold?

jwrober:
> FreeMono feels out-of-place. Are the others bold?

No, all are "regular". This is what bold mono looks like
image

jwrober:
After a discussion on Discord. We are going to go with a different font set.

PR #1374, opened 2022-09-15
Bundle Linux Libertine Fonts

### LINKED ISSUE #1292: Bundle our own fonts

### LINKED ISSUE #1341: Move macos/info.plist to dist/info.plist
We use `macos/info.plist` as part of the macOS packaging process. This file really should be in `dist/` along with the other distribution files.

Move the file, remove the `macos` directory and update the `build.yaml` file to capture the change.

None

**Additional context**
I found this as part of working on #916

jwrober:
Instead of adding the files directly to the repo, we use a `cmake` feature to go fetch the files at install and package time. We also set the dependency for the `.deb` package. Will need @pranavtifr to update the AUR package to match.

Closes #1292 
Closes #1341


CMakeLists.txt
```diff
@@ -342,5 +342,31 @@ if(UNIX AND NOT APPLE)
   endif(FREECIV_ENABLE_RULEDIT)
 endif(UNIX AND NOT APPLE)
 
+# We grab the Libertine Font package online for non-Unix (e.g. Win32, MSYS2, and macOS)
+#  For Linux, we add a dependency to the package.
+if(NOT UNIX)
```
lmoureaux, line 347: `UNIX` is true on Mac as well. I suggest making this a configure-time option.
Also, this should only be downloaded when we build the client (thinking of the LT server where we have a restrictive proxy).

CMakeLists.txt
```diff
@@ -342,5 +342,31 @@ if(UNIX AND NOT APPLE)
...
+  message(STATUS "Downloading Linux Libertine Font Package")
+
+  include(ExternalProject)
+  ExternalProject_Add(Libertine
+    PREFIX ${CMAKE_BINARY_DIR}
+    LOG_DIR ${CMAKE_BINARY_DIR}
+    DOWNLOAD_NO_PROGRESS TRUE
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+    TIMEOUT 30
```
lmoureaux, line 356: 30s for 3MB assumes that the user has a bandwidth of 100kB/s. Let's increase this a bit (5-10 minutes to be more on the safe side?).

CMakeLists.txt
lmoureaux, line 347: Maybe move to a separate file in `cmake/` to avoid cluttering the main CMakeLists too much.

CMakeLists.txt
```diff
@@ -342,5 +342,31 @@ if(UNIX AND NOT APPLE)
...
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+    INSTALL_COMMAND ""
+    URL https://sourceforge.net/projects/linuxlibertine/files/linuxlibertine/5.3.0/LinLibertineOTF_5.3.0_2012_07_02.tgz
```
lmoureaux, line 360: Can we add `URL_HASH` to ensure integrity?

CMakeLists.txt
```diff
@@ -342,5 +342,31 @@ if(UNIX AND NOT APPLE)
...
+  )
+
+  install(
+    DIRECTORY ${CMAKE_BINARY_DIR}/src/Libertine
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/fonts
```
lmoureaux, line 365: This installs to `/usr/share/fonts` on Linux and could overwrite some system files. Better put them in our private data dir, `SHARE/freeciv21/fonts`.

lmoureaux (changes requested):

pranavtifr:
For future reference, this needs to be added as a dependency
https://archlinux.org/packages/community/any/ttf-linux-libertine/

CMakeLists.txt
jwrober: That is the intention (to get picked up on all but Linux/Unix). I'll add another condition so it only runs if you are building the client.

CMakeLists.txt
jwrober: Yea looking at the documentation, I need to bump this up some to handle slower connections.

CMakeLists.txt
jwrober: Maybe move all the install stuff at the bottom to another file?

CMakeLists.txt
jwrober: The project does not offer a hash file. I would have included if one was available.

CMakeLists.txt
lmoureaux: What if I'm on Slitaz and I don't have the `ttf-libertine` package though? I'm for downloading it by default and allowing distributors to use the system version

jwrober:
> For future reference, this needs to be added as a dependency https://archlinux.org/packages/community/any/ttf-linux-libertine/

Is there an otf version of the package? The Debian one is otf instead of ttf.

CMakeLists.txt
jwrober: LoL, now that is an edge case since we don't technically support that distro. I'll change it.

lmoureaux:
> > For future reference, this needs to be added as a dependency https://archlinux.org/packages/community/any/ttf-linux-libertine/
> 
> Is there an otf version of the package? The Debian one is otf instead of ttf.

The Arch package has both ttf and otf, click on "View the file list for ttf-linux-libertine" to confirm. ttf- is just a prefix for font packages in Arch (though otf- does exist).

lmoureaux (approved):

Issue #1367, opened 2022-09-13
Documentation Reorg / Cleanup - Meta Issue


jwrober:
There are a few things that I would like to see us do to improve the layout of the documentation (`docs`) as well as how we interface/integrate with the in-game help. I reviewed the documentation for "The Battle of Wesnoth" and "Warzone 2100" for a view in what other popular OSS games are doing.

A view of current status:
* `General/*`: This directory has a miss-mash of content. FAQ, Install & Compile, Windows Install, Modpack Installer usage, server introduction. Sections should be on topic.
* `General/Manuals/client-manual`: Client specific manual. Other games that are technically client/server obfuscate the difference between the two and keep the conversation higher level for base content. Wesnoth differentiates between regular (e.g. casual) players vs advanced players.
* `Playing/*`: We have good docs here that are topic specific. The How To Play page has content that is very close, but also different in many ways as the "Strategy and Tactics" page in the in-game help. The topic of the Citizen Governor is on topic and easily linked to from other pages.
* `Modding/*`: On topic, well laid out.
* `Contributing/*`: Also on topic with some very recent improvements/changes.
* `Coding/*`: On topic as well. Similar to `Contributing`, has some very recent changes with some pending issues.

A potential future state:
* If a reader would read from "top to bottom", we should start with an "about fc21" page, then lead into "getting fc21" and eventually "compiling" it as a last resort. See how Wesnoth does it at https://wiki.wesnoth.org/Play. We sort of have a beginning of this on the main repo `README.md` file, but even that is incomplete.
* We need to break out and in some cases consolidate duplicate content. The FAQ, How to Play, Client Manual and "Strategy and Tactics" in-game help all have duplicate text. This should be consolidated as much as possible. The FAQ has a lot of super advanced stuff in it that should be broken out.
* We need to have "casual user" and "advanced user" documents. The casual user section should obfuscate the client/server nature of the game. You just "play". The client manual should be broken apart into casual and advanced concepts with cross-links where necessary. The advanced user section would get into the more technical aspects of the game, especially the server and hosting multiplayer games. The `on-the-server.rst` file is better in advanced user for example. 
* The current Client Manual should be broken apart and turned into a Game Manual. When talking about the two "options" dialog boxes: the current "local options" dialog should become the "interface options" and the "server options" should become the "game options".

So how do we break this up into small-ish PRs to aid reviews?

- [x] I think we need to change `General` to `Getting` and move things around. 
[...]

PR #1376, opened 2022-09-16
Documentation Update

### LINKED ISSUE #1330: INSTALL is out of date
In the root of the source tree we have a plain text `INSTALL` file. It points to an out of date URL for the RTD version as well as the text is out of date.

**Do you have suggestions?**
Use the RTD `install.rst` to update `INSTALL`

jwrober:
As part of the meta issue #1367, changing `General` to `Getting`, added an `about.rst` file, updated `install.rst` and moved `faq.rst` to `Playing`. Update files as needed for the move.  Also update `INSTALL` to point to RTD.

Closes #1330


docs/Getting/install.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Windows
+=======
+
+For more information on using the Windows Installer package, you can read about it at :doc:`windows-install`.
+
+Debian Linux
+============
+
+To install the Debian package, you will use the ``apt`` command with elevated privileges like this:
```
daavko, line 25:
```suggestion
To install the Debian package, use the ``apt`` command with elevated privileges like this:

```

daavko (commented):
Only small nitpicks from me (things that probably got missed previously, looking at the history). Other than that, LGTM

PR #1377, opened 2022-09-16
Make clang-tidy unhappy

lmoureaux:
It seems I uploaded clang-tidy results ages ago and for that reason we don't see the new ones. Trigger a new warning to test the integration.


lmoureaux:
# Success!!
image

PR #1381, opened 2022-09-17
Load Linux Libertine Font

### LINKED ISSUE #1273: Change default fonts to more sane values

jwrober:
Right now it assumes there is no `rc` file and then loads the font if not found in the font database from the filesystem and sets the defaults. I think there might need to be another check, but not sure where from.

Closes #1273

### DIFF
--- a/client/fonts.cpp
+++ b/client/fonts.cpp
@@ -172,11 +172,11 @@ void configure_fonts()
   }
 
   /* Sans Serif List */
-  sl << QStringLiteral("Linux Biolinum O")
-     << QStringLiteral("Liberation Sans") << QStringLiteral("Droid Sans")
-     << QStringLiteral("Ubuntu") << QStringLiteral("Noto Sans")
-     << QStringLiteral("DejaVu Sans") << QStringLiteral("Luxi Sans")
-     << QStringLiteral("Lucida Sans") << QStringLiteral("Arial");
+  sl << QStringLiteral("Linux Biolinum") << QStringLiteral("Liberation Sans")
+     << QStringLiteral("Droid Sans") << QStringLiteral("Ubuntu")
+     << QStringLiteral("Noto Sans") << QStringLiteral("DejaVu Sans")
+     << QStringLiteral("Luxi Sans") << QStringLiteral("Lucida Sans")
+     << QStringLiteral("Arial");
 
   font_name = configure_font(fonts::default_font, sl, max);
   if (!font_name.isEmpty()) {
@@ -202,7 +202,7 @@ void configure_fonts()
   sl.clear();
 
   /* Monospace List */
-  sl << QStringLiteral("Linux Libertine Mono O") << QStringLiteral("Cousine")
+  sl << QStringLiteral("Linux Libertine Mono") << QStringLiteral("Cousine")
      << QStringLiteral("Liberation Mono")
      << QStringLiteral("Source Code Pro")
      << QStringLiteral("Source Code Pro [ADBO]")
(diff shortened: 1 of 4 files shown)


jwrober:
ok found another use case - you have an existing `rc` file, but the font isn't loaded and needs to be.

client/fonts.cpp
```diff
@@ -125,73 +127,117 @@ void fcFont::setFont(const QString &name, const QFont &qf)
...
+ */
+bool isFontInstalled(const QString &font_name)
+{
+  QFontDatabase database;
+
+  if (database.families().contains(font_name)) {
+    return true;
+  } else {
+    return false;
+  }
```
lmoureaux, line 140:
```suggestion
  return database.families().contains(font_name);
```

client/fonts.cpp
```diff
@@ -125,73 +127,117 @@ void fcFont::setFont(const QString &name, const QFont &qf)
...
+}
+
+/**
+   Loads the fonts into the font database.
```
lmoureaux, line 144:
```suggestion
 * Loads the fonts into the font database.
```

client/fonts.cpp
```diff
@@ -125,73 +127,117 @@ void fcFont::setFont(const QString &name, const QFont &qf)
...
+ */
+void load_fonts()
+{
+  QFileInfoList il;
+
+  il = find_files_in_path(get_data_dirs(), QStringLiteral("fonts"), false);
```
lmoureaux, line 150:
```suggestion
  const auto il = find_files_in_path(get_data_dirs(), QStringLiteral("fonts"), false);
```

client/options.cpp
```diff
@@ -1942,44 +1945,46 @@ static struct client_option client_options[] = {
                    COC_FONT, 0, -100, 100, allfont_changed_callback),
     GEN_FONT_OPTION(gui_qt_font_default, "default_font", N_("Default font"),
                     N_("This is default font"), COC_FONT,
-                    "Sans Serif,10,-1,5,75,0,0,0,0,0",
+                    "Linux Biolinum O,12,-1,5,50,0,0,0,0,0",
```
lmoureaux, line 1948: What happens if we have the TTF and not the OTF? I would hope for a gentle fallback/substitution...

lmoureaux (commented):
Sorry for the single comments earlier, wrong button :eject_button:

jwrober:
well darn, i was hoping to cheat a little and just accept the changes, but clang-format is complaining.

client/options.cpp
lmoureaux: Then let's remove `*.ttf` from the list of files we can load

lmoureaux: Actually on my system it will fallback if you don't pass the O:
```
$ fc-match "Linux Biolinum"
LinBiolinum_R.otf: "Linux Biolinum O" "Regular"
```

PR #1385, opened 2022-09-18
Consistent use of "Longturn" instead of "LongTurn"

### LINKED ISSUE #1364: Change to a consistent use of "Longturn"
There is inconsistent usage of our organization in docs, website, etc. We have used both "LongTurn" and "Longturn".

**Do you have suggestions?**
The organization name is properly "Longturn", so any references to "LongTurn" should be changed to "Longturn".

jwrober:
Quick sed really. On top of #1376 

Closes #1364


daavko (approved):
LGTM

PR #1386, opened 2022-09-18
Improve the buttons on various dialog boxes

### LINKED ISSUE #1110: Replace "Ok" and "Cancel" in dialog buttons with something that better describes what will happen
Current dialogs (didn't check all of them, but I assume they share this behavior) have two buttons: "Ok" and "Cancel" (or variations, such as "Yes" and "No"). Those don't actually describe what will happen when the button is clicked.

As an example, let's take the "Quit game" dialog. It's current form:
freeciv21-client_Aa8WlybG0l
I'd replace "Ok" with "Quit", and "Cancel" with "Keep playing". Same as this with all other dialogs for actions with two outcomes. As another example, the "disband unit" dialog. Replacing "Yes" with "Disband" and "No" with "Cancel" makes it a little clearer IMO.

Leaving it as it is, which would probably be fine, but IMO titling dialog buttons better will make the game more polished.

**Additional context**
This was inspired by what Google's Material Design guidelines recommend for alert dialogs: [link to relevant section](https://material.io/components/dialogs#alert-dialog). Apple has [similar guidelines](https://developer.apple.com/design/human-interface-guidelines/components/presentation/alerts/#buttons).

### LINKED ISSUE #1370: "Leaving a local game will end it!"
**Describe the bug**

When doing Game > Leave in a single-player game, we get a warning saying that "Leaving a local game will end it!". This text sounds like something very annoying will happen. In practice, the server is terminated (that's what the text is trying to say), but the game is saved on exit.

**To Reproduce**
1. Start a local game
2. Click on "Leave Game" in the "Game" menu
3. See the offending box

**Expected behavior**

We always save the game, or maybe we add a box offering to do it instead of emitting a very loud warning.

**Screenshots**

image

 - OS: any and all
 - Freeciv21 version: 3.0-beta4

jwrober:
Closes #1110 
Closes #1370

### DIFF
--- a/client/menu.cpp
+++ b/client/menu.cpp
@@ -2900,7 +2900,7 @@ void mr_menu::back_to_menu()
   if (is_server_running()) {
     ask = new hud_message_box(king()->central_wdg);
     ask->set_text_title(
-        _("Do you want to leave the game?\n\n Leaving a single-player game "
+        _("Do you want to leave the game?\n\nLeaving a single-player game "
           "will end it. Be sure to save first."),
         QStringLiteral("Leave Game"));
     ask->setStandardButtons(QMessageBox::No | QMessageBox::Yes);
--- a/client/menu.cpp
+++ b/client/menu.cpp
@@ -1,5 +1,5 @@
 /*
- Copyright (c) 1996-2020 Freeciv21 and Freeciv contributors. This file is
+ Copyright (c) 1996-2022 Freeciv21 and Freeciv contributors. This file is
  part of Freeciv21. Freeciv21 is free software: you can redistribute it
  and/or modify it under the terms of the GNU  General Public License  as
  published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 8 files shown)


client/menu.cpp
```diff
@@ -2899,10 +2899,14 @@ void mr_menu::back_to_menu()
 
   if (is_server_running()) {
     ask = new hud_message_box(king()->central_wdg);
-    ask->set_text_title(_("Leaving a local game will end it!"),
-                        QStringLiteral("Leave game"));
-    ask->setStandardButtons(QMessageBox::Cancel | QMessageBox::Ok);
-    ask->setDefaultButton(QMessageBox::Cancel);
+    ask->set_text_title(
+        _("Do you want to leave the game?\n\n Leaving a single-player game "
```
lmoureaux, line 2903: I think the space after \n will result in a line slightly off center

lmoureaux (approved):
Only one very minor comment

client/menu.cpp
jwrober: This is what it looks like on my system
image

lmoureaux: Second line: 25px on the left, 20px on the right

jwrober: Yep, it sure does. I'll push another commit

PR #1389, opened 2022-09-26
Add sha256 hash files to the source archives

jwrober:
No related issue. Added since we are doing the same for the macOS `.dmg`, figured it good to offer for all of our assets at release.


jwrober:
image

PR #1390, opened 2022-09-26
Move General/Manuals to Manuals

jwrober:
Part of the re-org work in #1367. While in some of the files added license and cleaned up a few sphinx error/warnings


daavko (approved):
LGTM

PR #1392, opened 2022-09-27
Documentation links fixes

jwrober:
Branch name would suggest game manual, but Discord conversation derailed it. While here, found out how to use `:ref:` role properly. Fixed cross link in how to play and document the usage in style guide. No referenced issue.


daavko (commented):
LGTM, also a question:
Are we using `:numref:` anywhere else that would make more sense as a `:ref:`? A quick search suggests there's a lot of `:numref:`s, but I have no idea whether they should be replaced with `:ref:`s or not

Apologies for the delay, completely missed that I have a pending review

jwrober:
I don't think there are other improper uses of `:numref:`. Thanks for the review @daavko.

PR #1403, opened 2022-10-05
Docs: Add Qt Creator to Linux dev env

### LINKED ISSUE #1400: Document how to edit ui files

We should document that .ui files can be edited with Qt Designer and how to install it.

**Do you have suggestions?**

In the dev tips section of the docs?
> Qt user interface files (`.ui`) can be edited with [Qt Designer](https://doc.qt.io/qt-6/qtdesigner-manual.html), which comes bundled with the [Qt Creator](https://www.qt.io/product/development-tools) IDE.

I'm unsure what the VS integration provides.

jwrober:
Closes #1400


daavko (approved):
LGTM, nothing particular pops out as wrong

PR #1404, opened 2022-10-05
Change Tax Rate to National Budget

Koffy25:
Change Tax Rate to National Budget


client/top_bar.cpp
```diff
@@ -81,7 +81,7 @@ QSize tax_rates_widget::sizeHint() const
 }
 
 /**
- * Renders the tax rates widget
+ * Renders the national budget widget
  */
 void tax_rates_widget::paintEvent(QPaintEvent *event)
```
jwrober, line 86: should this be `national_budget_widget`?

client/top_bar.h
```diff
@@ -28,7 +28,7 @@ void top_bar_left_click_science();
 void top_bar_show_map();
 
 /**
- * Top bar widget for tax rates.
+ * Top bar widget for national budget.
  */
 class tax_rates_widget : public QToolButton {
```
jwrober, line 33: same `national_budget_widget` class?

common/networking/packets.def
```diff
@@ -488,7 +488,7 @@ PACKET_GAME_INFO = 16; sc, is-info
   UINT16 history_interest_pml;
   /* size limit for cities before they can celebrate */
   UINT8 celebratesize;
-  BOOL changable_tax;
+  BOOL changable_budget;
```
jwrober, line 491: @lmoureaux does this break the network protocol?

docs/Manuals/client-manual.rst
```diff
@@ -1389,24 +1387,24 @@ you what was sold where.
 .. note:: You can only sell one city improvement at at time per turn, so you might not be able to do all the
   things you want every turn.
 
-Tax Rates View
+National Budget View
 ^^^^^^^^^^^^^^
```
jwrober, line 1391: you need the same number of `^` symbols as there are characters in the heading.

jwrober (changes requested):

common/networking/packets.def
lmoureaux: No, names are meaningless for the protocol (only the data layout is meaningful)

Koffy25:
Fixes for Change Tax Rate to National Budget

PR #1408, opened 2022-10-06
Add server commands to chat auto completion word list

### LINKED ISSUE #535: Add commands to the chat autocompletion word list
It would be nice to have autocompletion of commands in the chat line. This can be achieved by simply adding them to the word list:

https://github.com/longturn/freeciv21/blob/80d2c1546a2e3b7398daa7dd677b2ef54f6d1d8c/client/gui-qt/chatline.cpp#L54

Make sure to have the `/` so they don't randomly appear in the text.

jwrober:
Closes #535

### DIFF
--- a/client/chatline.cpp
+++ b/client/chatline.cpp
@@ -1,5 +1,5 @@
-/*
-   //           Copyright (c) 1996-2020 Freeciv21 and Freeciv contributors.
+/**
+   //           Copyright (c) 1996-2022 Freeciv21 and Freeciv contributors.
  _oo\ This file is  part of Freeciv21. Freeciv21 is free software: you can
 (__/ \  _  _   redistribute it and/or modify it under the terms of the GNU
    \  \/ \/ \    General Public License  as published by the Free Software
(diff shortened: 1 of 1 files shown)


client/chatline.cpp
```diff
@@ -385,7 +577,7 @@ void chat_widget::set_chat_visible(bool visible)
 }
 
 /**
- * Shows the chat and ensures the chat line has focus
+   Shows the chat and ensures the chat line has focus
```
lmoureaux, line 580: The * was correct :D
```suggestion
 * Shows the chat and ensures the chat line has focus
```

lmoureaux (commented):
How much of the list is shared between the two branches of the `if`? Maybe we can store the commands in a static `QStringList`?

client/chatline.cpp
jwrober: Interesting. How so? It was the only header like it in the file

jwrober:
> How much of the list is shared between the two branches of the `if`? Maybe we can store the commands in a static `QStringList`?

Actually a lot of it is the same. Let me look into a static list defined first.

client/chatline.cpp
lmoureaux: https://longturn.readthedocs.io/en/latest/Coding/guidelines.html#documentation-comments

jwrober: Not consistent then. most of the function comments in the file use this convention

```
/**
   some comment
*/
```

lmoureaux: Yes, most functions are still like this because I didn't dare mass-replacing the old style.

PR #1409, opened 2022-10-06
Break out production buy gold and turns as columns in city view

### LINKED ISSUE #1052: Cities: Provide an option to break out production and gold into two columns
On the Cities page (F4) there is currently a column that combines the number of turns to produce an item with the amount of gold it would take to rush-buy the item. The problem is often it would be good to be able to simply sort by amount of gold required only, which is not available at this time.

Would like to see another option to break those into two columns, in the same way we do that for food, prod, trade, etc. Keep the existing column as other players may prefer it, but then also allow for two separate columns.

Right now I personally keep those in an external sheet when playing games by hand. Would prefer to be in game and not have to use anything external.

**Additional context**
None

jwrober:
Also updated the function headers to match coding guidelines

Closes #1052

### DIFF
--- a/client/cityrepdata.cpp
+++ b/client/cityrepdata.cpp
@@ -37,10 +37,10 @@
 #include "cityrepdata.h"
 
 /**
- * cr_entry = return an entry (one column for one city) for the city report
- * These return ptrs to filled in static strings.
- * Note the returned string may not be exactly the right length; that
- * is handled later.
+   cr_entry = return an entry (one column for one city) for the city report
+   These return ptrs to filled in static strings.
+   Note the returned string may not be exactly the right length; that
+   is handled later.
  */
 static const char *cr_entry_cityname(const struct city *pcity,
                                      const void *data)
@@ -52,7 +52,7 @@ static const char *cr_entry_cityname(const struct city *pcity,
 }
 
 /**
- * Translated name of nation who owns this city.
+   Translated name of nation who owns this city.
  */
 static const char *cr_entry_nation(const struct city *pcity,
                                    const void *data)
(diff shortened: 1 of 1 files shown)


client/cityrepdata.cpp
```diff
@@ -884,21 +940,22 @@ void init_city_report_game_data()
...
+   The following several functions allow intelligent sorting city report
+   fields by column.  This doesn't necessarily do the right thing, but
+   it's better than sorting alphabetically.
+
+   The GUI gives us two values to compare (as strings).  We try to split
+   them into an array of numeric and string fields, then we compare
+   lexicographically.  Two numeric fields are compared in the obvious
+   way, two character fields are compared alphabetically.  Arbitrarily, a
+   numeric field is sorted before a character field (for "justification"
+   note that numbers are before letters in the ASCII table).
```
lmoureaux, line 952:
```suggestion
  The following several functions allow intelligent sorting city report
  fields by column.  This doesn't necessarily do the right thing, but
  it's better than sorting alphabetically.
  The GUI gives us two values to compare (as strings).  We try to split
  them into an array of numeric and string fields, then we compare
  lexicographically.  Two numeric fields are compared in the obvious
  way, two character fields are compared alphabetically.  Arbitrarily, a
  numeric field is sorted before a character field (for "justification"
  note that numbers are before letters in the ASCII table).
```

lmoureaux (changes requested):
The code with `bufone` and `buf` looks a bit overcomplicated but I guess it fits the style of the file. We'll refactor if/when we turn this to QString

One last suggestion to avoid breaking the blame. Just apply it and stack merge.

jwrober:
> The code with `bufone` and `buf` looks a bit overcomplicated but I guess it fits the style of the file. We'll refactor if/when we turn this to QString
> 
I simply copied the function that gathered both gold and turns and then pasted it. So yea, it does use the same style.  

I'll push another commit to fix that block and we can do a squash merge when the CI completes.

lmoureaux (approved):

Issue #941, opened 2022-03-16
Organize the storage of media files in the source tree


jwrober:
**Is your feature request related to a problem? Please describe.**
Right now we have images stored in multiple places in the source tree. Some of the images are actually duplicates. This request is to organize as many of the images and other media files (e.g. icons, music, sound) that is common to a single directory structure under `/data`.

**Describe the solution you'd like**
We have common/shared media in the following directories:
- `/data`
- `/data/buildings`
- `/data/flags`
- `/data/graphics`
- `/data/graphics/buildings`
- `/data/graphics/cimpletoon`
- `/data/graphics/logo+splash`
- `/data/graphics/tech`
- `/data/graphics/wonders`
- `/data/icons/*`
- `/data/misc`
- `/data/themes/gui-qt/icons`
- `/data/wonders`

**Describe alternatives you've considered**
Consolidate all to a `/media` structure, such as:

- `/data/media/sound`
- `/data/media/buildings`
- `/data/media/flags`
- `/data/media/icons`
- `/data/media/music`
- `/data/media/techs`
- `/data/media/tilesets`
- `/data/media/other`
- `/data/media/wonders`

Should move `/data/graphics/cimpletoon` to `/data/cimpletoon` so the files are in the tileset folder where they belong.

**Additional context**
None, just trying to organize things better.

lmoureaux:
Random thoughts:
* Files used by tilesets should keep the same filesystem layout to avoid breaking compat. What we could do is split `data` into tileset and ruleset folders.
* I feel like `media` is redundant with `data`? Most of these files have an associated `.tilespec` somewhere in the tree.

jwrober:
`/data/misc` is the only listed directory that has tileset stuff in it. I like the idea of moving the packaged rulesets and tilesets to a couple sub-dirs. Even if we reduce the number of shipped items of each time, organization is always good.

I went with `media` because I was thinking of including sound, but based on your comment maybe not combine and just go with `images` or keep `graphics` and just re-org and consolidate inside there. The stdsounds and stdmusic musicsets would just have their files self contained like we do for tilesets.

Taking another look, I can't tell what `/data/graphics/cimpletoon` is used for. The tileset does not reference it.

lmoureaux:
These are "source files" for the graphics -- in theory you can use Blender and redo all the renders. Everything in `graphics` is like this.

PR #1412, opened 2022-10-11
Tileset image organization phase 1

jwrober:
* Create a new `./data/tilesets` directory to consolidate tileset sprites into a common area.
* Move the individual unit images out of `docs`.
* Move buildings, wonders and flags.

Part of #941


jwrober:
I wonder if I need to worry about/fix those codeFactor issues...

PR #1416, opened 2022-10-13
Create small versions of the standard buildings and wonders

jwrober:
Standardize on a set of sprites for buildings and wonders

Part of #941


data/hex2t.tilespec
```diff
@@ -151,6 +151,7 @@ files =
   "misc/governments.spec",
   "misc/specialists.spec",
   "misc/buildings.spec",
+  "misc/wonders.spec",
```
lmoureaux, line 154: Won't this break tilesets relying on `buildings.spec` containing wonders?

jwrober: Technically yes. I handled all the ones we have in our repo. I would think that SnF can update Aviation to fit this model without much trouble. Same with other tileset editors.

Can a spec file reference another spec file? e.g I add ref to wonders.spec in buildings.spec?

PR #1472, opened 2022-11-13
Support Libertinus as a Libertine alternative

### LINKED ISSUE #1428: Linux Libertine Font is going out of support on AUR
**Describe the bug**
Linux Libertine Font is going out of support on AUR. They are asking for users to switch to Linux Libertinus Fonts instead. 

**To Reproduce**
No real steps. See note from AUR.

```
Note: development of Libertine seems to have halted, despite outstanding bugs.                                                                                                                                                                                                
Users are advised to upgrade to the de facto successor,                                                                                                                                                                                                                       
[community]/libertinus-font, and to update any documents and configuration                                                                                                                                                                                                    
if necessary.                                                                                                                                                                                                                                                                 
For compatibility's sake, Linux Libertine will remain in the repos until 2022.                                                                                                                                                                                                
Please refer to https://wikipedia.org/wiki/Linux_Libertine#Derivative_works                                                                                                                                                                                                   
for more information. 
```

**Expected behavior**
Maintain support of the AUR repo for our bundled font. Issue right now is there is no Debian package that installs Linux Libertinus fonts. However we could simply fall back to the packaged version and not worry about a system dependency.

**Screenshots**
N/A

 - OS: Arch Linux
 - Freeciv21 version: 3.0-beta.5

**Additional context**
Came up as part of the release of Beta.5

lmoureaux:
Support both the Linux Libertine and Libertinus font families. Give priority to the latter because the Libertine project is dead.

Closes #1428.

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -74,6 +74,7 @@
 #include "connectdlg_common.h" // client_kill_server()
 #include "control.h"
 #include "editor.h"
+#include "fonts.h"
 #include "global_worklist.h"
 #include "governor.h"
 #include "mapview_common.h"
@@ -532,6 +533,7 @@ int client_main(int argc, char *argv[])
   init_themes();
 
   options_init();
+  configure_fonts();
   options_load();
 
   script_client_init();
--- a/client/fonts.cpp
+++ b/client/fonts.cpp
@@ -19,6 +19,10 @@
 #include "gui_main.h"
 #include "options.h"
 
+static void configure_font(const QString &font_name, const QStringList &sl,
+                           QFont::StyleHint hint, int size,
+                           bool bold = false);
+
 /**
    Font provider constructor
  */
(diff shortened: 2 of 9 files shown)


jwrober:
I don't think Libertinus is installed in most distros (certainly not on Windows and probably not on Mac). We might need to go fetch it from their repo and include in the distribution like we do for Libertine.

lmoureaux:
This is about using Libertinus instead of Libertine if it is available, not shipping it by default. We might consider doing that, especially with the Libertine Sourceforge mirror failing so often in the CI, but the point here is mainly to get rid of the AUR warning.

jwrober:
I guess my thinking here was to move to Libertinus and away from Libertine for the varying issues (old, no longer active, sf download issues, AUR deprecation, etc) would be a better approach. As I stated before, Libertinus is not installed anywhere and there are no native packages for it, so best to just bundle it instead of Libertine.

lmoureaux:
I agree we should do that in the long run, but it would require more changes to our packaging. This PR makes using Libertinus possible at all, keeping the rest for later.

jwrober:
I need to install Libertinus to test this. Will work on it soon.

lmoureaux:
That's why I asked @psampathkumar to review :P

jwrober:
I'm an equal opportunity reviewer, LoL

jwrober (approved):

psampathkumar (approved):

PR #1474, opened 2022-11-13
Small tweaks to default settings

lmoureaux:
* **Fullscreen by default:** We've been talking about this for ages but I couldn't find a relevant issue.
* **Show terrain on the Minimap:** Why was this even disabled by default?


jwrober (approved):
Seems reasonable to me.

PR #1479, opened 2022-11-13
Remove fc_client::quit()

lmoureaux:
The only thing it was doing was calling QApplication::quit(). Make that explicit instead.


jwrober (approved):
LTGM

PR #1483, opened 2022-11-13
Add support for scroll events on the map

lmoureaux:
Support the following gestures:

* Mouse wheel/touchpad scroll gesture: scroll the map
* Shift+wheel/touchpad: scroll the map with x and y inverted
* Ctrl+wheel/touchpad: zoom

This is consistent with the gestures widely used in browsers and other software.


jwrober (approved):
Works for me

PR #1496, opened 2022-11-17
Fix saving map images

### LINKED ISSUE #1486: "Save map image" saves a black image
**Describe the bug**
As the title says

**To Reproduce**
1. Game > Save map as image
2. Open the saved image
3. It's fully black

**Expected behavior**
The whole visible map is saved as an image.

 - OS: Linux
 - Freeciv21 version: 3.0-beta5
 - Ruleset/Longturn game: Reproduced with LT75 and in a local game

lmoureaux:
It's big because rendering is no longer synchronous, so I had to switch to an async callback.

Closes #1486.

### DIFF
--- a/client/listener.h
+++ b/client/listener.h
@@ -148,7 +148,7 @@ template <class _type_> void listener<_type_>::listen()
 ***************************************************************************/
 template <class _type_> listener<_type_>::~listener()
 {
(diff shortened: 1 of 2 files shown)


jwrober (approved):
Works for me.

PR #1500, opened 2022-11-19
Install theme.conf

lmoureaux:
Colors are incorrect (use the system palette) without this file.

Reported by Mu on Discord.


jwrober (approved):
Interesting find. LTGM.

PR #1505, opened 2022-11-19
Bump jidicula/clang-format-action to 4.9.0

### LINKED ISSUE #1503: CI failing because clang-format-11 is no longer in Debian sid
**Describe the bug**

Our CI started failing because `clang-format-11` has been removed from Debian sid, which is used by the action. We could use the opportunity to upgrade to version 14 present in Ubuntu 22.04 and current Arch Linux.

**Additional context**

Loosely related to #115, #1478

lmoureaux:
4.4.0 started failing today for clang-format-11. Since 4.5.0, the action supports use even older versions so 11 should be fine.

Closes #1503.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -414,7 +414,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Run clang-format style check for C/C++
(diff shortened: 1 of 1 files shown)


jwrober (approved):
The action completing tells us all we need to know.

Issue #1517, opened 2022-11-23
Various feedback and bug reports from LT75 participants


daavko:
Some feedback from LT75 participants, grouped into an issue so we don't forget about it. No particular order, but I added checkboxes so things can potentially be tracked as completed as necessary. Also provided links to Discord where appropriate in case more context is necessary

- [x] Tech tree should look more like 2.6 GTK when using the Classic theme
  - https://discord.com/channels/378908274113904641/912500712833974322/1042818119481180200
- [x] Online status for players that's visible even if the player has clicked "End Turn"
- [x] State of "Allies Only" button in chat isn't saved between sessions, so you have to remember to click it every time
  - https://discord.com/channels/378908274113904641/912500712833974322/1044700678729240696
- [x] Inside the research button, there could be bulbs/turn and turns needed until the tech is reached. Seems like there's enough space
  - https://discord.com/channels/378908274113904641/1010150995843625040/1042352147024580628
- [x] Clicking "End Turn" doesn't provide enough visual feedback that it has been clicked this turn already
  - https://discord.com/channels/378908274113904641/1010150995843625040/1044708610510434455
- [x] Font for "Notify Label" should be monospace by default (it was monospace before, now it's sans-serif)
  - https://discord.com/channels/378908274113904641/1010150995843625040/1044711318311161979
  - https://github.com/longturn/freeciv21/blob/master/client/fonts.cpp#L179 (where it's set)
  - https://github.com/longturn/freeciv21/commit/20bdecc8e761ed897e95c1efc97f4530df43438f (commit where it was changed)
- [x] Tooltip for "Turn Done" that shows TC date and time in local time and UTC, and time left until TC
- [x] `(network exception)` is shown in the `/list connections` command output next to player names due to `closing_reason` in `connection` objects not being cleared when attaching a connecting player to an existing `connection` object
  - https://discord.com/channels/378908274113904641/912500712833974322/1044278231852908554
  - https://discord.com/channels/378908274113904641/912500712833974322/1044284621535334461
- [ ] It either shouldn't be necessary to `/take` your nation when first connecting to a game, or it should be clearly documented that you need to do so. It causes unnecessary confusion
  - https://discord.com/channels/378908274113904641/1010150995843625040/1042063542406369300

lmoureaux:
> Clicking "End Turn" doesn't provide enough visual feedback that it has been clicked this turn already 

That's a theme issue, it's quite clear with NightStalker.

jwrober:
[...]

PR #1519, opened 2022-11-26
Clear closing reason for connection object when a client connects

daavko:
The closing reason would otherwise persist from the previous client disconnecting and would appear when listing connected clients via the `/list connections` command.

Part of #1517


jwrober (approved):
LGTM

PR #1523, opened 2022-11-26
Improvements to the classic theme

lmoureaux:
This PR polishes Classic a bit and fixes the tech tree colors issue from #1517:

* Use the pattern background for the bottom-right widget around the minimap
* Use the pattern background for the tech tree
* Use tileset colors for the tech tree (restoring Gtk behavior)
* Override tileset colors for the tech tree in NightStalker (because they look terrible on the dark blue background)

This gets rid of a mir3xism in the code to make everything work.

I didn't check themes other than Classic and Nightstalker. They'll use colors from the tileset as the default.


jwrober (approved):
This looks really good. I tested on the other themes as well and they look "the same" effectively so no breakage I see. Nice to see the removal of magic numbers in the refactor!

PR #1532, opened 2022-11-27
Libertinus Font

### LINKED ISSUE #1484: Bundle Libertinus instead of Libertine Fonts
Linux Libertine has not been updated since 2012 and is deprecated in the AUR for Arch (see #1428).

Replace the downloaded and bundled Libertine font with the supported Libertinus Font

We could offer both, but that seems excessive. One really good font is good enough.

**Additional context**
Came up as part of discussing PR #1472

jwrober:
Switch to Libertinus font
Fix notify label to monospace

Closes #1484 
Part of #1517

### DIFF
--- a/client/fonts.cpp
+++ b/client/fonts.cpp
@@ -175,8 +175,6 @@ void configure_fonts()
      << QStringLiteral("Linux Biolinum");
 
   configure_font(fonts::default_font, sl, QFont::SansSerif, default_size);
-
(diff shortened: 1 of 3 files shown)


psampathkumar:
The new configure flag needs to be documented.

jwrober:
It's in #1533

psampathkumar (approved):

PR #1534, opened 2022-11-27
Add a CMakePreset with the flags used in major distros and enable in …

### LINKED ISSUE #172: Consider enabling LTO by default in release builds
According to quick tests it can save up to 30% of TC processing duration.

### LINKED ISSUE #1530: Test more compiler warning flags

OpenSUSE enables `-Walloc-size-larger-than` by default (using GCC 12). Our code triggers a few of these warnings.

Test various warning flags as used by distros in the CI. Consider enabling them by default for all builds.

**Additional context**

Similar problem in Arch with `-Werror=format-security` breaking the build.

psampathkumar:
…actions

Fixes #1530

This enables linker flags too so fixes #172

Also fix a issue which pops up from enabling -Werror=format-security and have more secure format strings.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -78,6 +78,37 @@ jobs:
             build/Linux-x86_64/freeciv21_*_amd64.deb
             build/Linux-x86_64/freeciv21_*_amd64.deb.sha256
 
+  flagstest:
+    name: "Additional compile flags in Ubuntu"
+    needs: clang-format
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


lmoureaux (changes requested):
The workflow is not valid. .github/workflows/build.yaml (Line: 82, Col: 5): Required property is missing: runs-on

jwrober (approved):
Other than the issue lmoureaux mentioned. It looks ok to me. Geez it spits out a billion warnings.

lmoureaux (approved):

PR #1539, opened 2022-11-28
Add open diplomacy button with base code to nations tab

### LINKED ISSUE #1420: Don't want to have to right-click Nations to see open diplomacy sessions
If you have an open diplomacy session with another player and leave the screen to do something, you have to right-click on the nations view button to come back to the diplomacy screen. None of the other buttons on the top view require the use of a right-click to enable an action. This makes  playing the game harder that it should.

Add a button to the top menu bar of the nations view instead. The "top menu bar of the nations view" is where you see "meet", "cancel treaty", "withdraw vision" and "toggle ai mode". I would add a "view open meetings" or something like that. The button would be greyed out if no meetings are open and clickable if there are.

None

**Additional context**
Came up as part of the review of PR #1419

jwrober:
Also includes some text formatting in the nations view.

Closes #1420


lmoureaux (commented):
Right-clicking on top bar button after this change:

* Map: Shows the map when not visible
* Units: Recenters the map (!)
* Cities: Shows the map
* Nations: Does nothing
* Techs: Quick tech selection
* Gold: Shows the map
* National Budget: Does nothing
* Indicator sprites: Does nothing
* Messages: Shows the map

Would sorting out the inconsistencies be hard? I'm in favor of no right click action except for the quick tech selection.

jwrober:
> Would sorting out the inconsistencies be hard? I'm in favor of no right click action except for the quick tech selection.

Let me see what I can figure out. Seems logical to get some consistency.

jwrober:
ok I think this might be harder that it looks right away. I think we should open another issue to clean up/standardize right-click behavior on the top bar. Thoughts?

client/plrdlg.cpp
```diff
@@ -750,14 +757,27 @@ void plr_report::call_meeting()
...
       &client.conn, player_number(other_player), CLAUSE_CEASEFIRE);
 }
 
+/**
+   Slot for diplomacy
+ */
+void plr_report::plr_diplomacy()
+{
+  int i;
+  i = queen()->gimmeIndexOf(QStringLiteral("DDI"));
```
lmoureaux, line 774: Remark for the future: `int i = queen()->gimmeIndexOf(...);` in a single line and simplify if clauses when possible.

lmoureaux (approved):

PR #1542, opened 2022-11-30
Enable resizable minimap panel

### LINKED ISSUE #1489: Make the minimap resizeable

It may be too big for some people or on some systems. Also mentioned in #1358 

As title says

Big/normal/small UI modes

jwrober:
Closes #1489

### DIFF
--- a/client/minimap_panel.h
+++ b/client/minimap_panel.h
@@ -22,7 +22,6 @@ class QPainter;
 class QPushButton;
 class QPixmap;
 class QResizeEvent;
-class chatwdg;
 
 class map_view;
 
--- a/client/minimap_panel.h
+++ b/client/minimap_panel.h
@@ -4,9 +4,25 @@
  *
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
+#pragma once
 
 #include "ui_minimap_panel.h"
+
+// qt-client is one true king
 #include "widgetdecorations.h"
+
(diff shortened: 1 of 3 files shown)


client/minimap_panel.h
```diff
@@ -4,17 +4,32 @@
...
+class QItemSelection;
+class QListWidget;
+class QMouseEvent;
+class QObject;
+class QPaintEvent;
+class QPainter;
+class QPushButton;
+class QPixmap;
+class QResizeEvent;
+
```
lmoureaux, line 25: Why do we need those?

lmoureaux (commented):
Code only, I need to test if it remembers the size after closing and reopening (that code will always surprise me)

client/minimap_panel.h
jwrober: I don't think it remembers the size. I copied over from the messages widget, so might not need all of them.

jwrober: I pushed another commit getting rid of that extra stuff that turns out I didn't need. I can confirm that the geometry is not saved between turns, same as messages widget. I am open to adding that to this PR for minimap_panel. Could use some direction on how to do that.

lmoureaux (approved):
I think we can got without remembering the size for now. What we have for the chat is broken anyway, and it would make sense to use a common solution.

PR #1545, opened 2022-12-01
Revert cruise missile graphic

lmoureaux:
Version in the repo was from FCW. We don't want graphics from there; also document what to check when merging artwork.


jwrober (approved):
The new image is the same as `atorpedo.png`, but that is fine.

PR #1546, opened 2022-12-02
Try to fix Github Actions not running in repo

lmoureaux:
I think the action from the fork cancels the one from the main repo (if actions are enabled in the fork, as is the case with mine).


lmoureaux:
Looks like they're running and I didn't start them manually.

jwrober (approved):
LGTM

PR #1547, opened 2022-12-03
Initialize options after creating the QApplication

lmoureaux:
QFont can only be instantiated once we have the QApplication. Font options were instantiated statically instead, leading to an immediate crash on Mac and, likely, strange behaviour with HiDPI on Windows reported by some users.

Modify the code to initialize options dynamically on startup, after creating the QApplication. For this I had to turn client_options to a pointer, necessitating many changes from client_options.x to client_options->x. The important changes are in options.{h,cpp} and client_main.cpp: static initializers were turned into functions and options_init is called a bit earlier.

Closes #1538.
Closes #1549.

**Test plan**

* Stress test options (both client and server).
* Ask @wieder-fi to test the Mac binary

PR #1551, opened 2022-12-03
Fix unit movement animations

### LINKED ISSUE #1516: Unit movement setting has no effect
**Describe the bug**
Any change to the unit movement setting has no effect

**To Reproduce**
1. Move a unit and notice how fast it moves.
2. Max out the interface options -> graphics tab -> unit movement animation time to 2000 ms
3. Move a unit and notice that it still moves at the same rate as before and is not "slow"

**Expected behavior**
The setting actually does what it says it will.

**Screenshots**
N/A

 - OS: Linux, but assume its for all platforms
 - Freeciv21 version: master branch as of 11/22/22

**Additional context**
None

lmoureaux:
Confusion between seconds and milliseconds resulted in the animation never working. Fix the inconsistency.

Closes #1516.

Will conflict with #1547


jwrober (approved):
This patch fixes the bug

PR #1552, opened 2022-12-04
Fix unit bouncing

### LINKED ISSUE #1515: Transport and engineers disbanded after terraforming swamp
**Describe the bug**
I just had a Transport and a couple Engineers disbanded after terraforming shallow ocean to swamp.

**To Reproduce**
1. Put as many engineers as you can on a boat
2. Have them terraform shallow to swamp
3. Wait until they're done
4. They're gone!

**Expected behavior**
The transport gets bounced to a neighboring tile.

 - OS: Linux
 - Freeciv21 version: 883f4de67b1476984bbceac57bc6986ce0eb4e3d
 - Ruleset: `classic`

lmoureaux:
* Unset any existing unit activity so the path finding code works
* Use real distance instead of Manhattan, so we can move to any adjacent tile with distance = 1

Closes #1515.

[Load and do Shift+Enter.](https://github.com/longturn/freeciv21/files/10147587/transform.sav.gz)

### DIFF
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -1298,7 +1298,7 @@ bool bounce_path_constraint::is_allowed(
 {
   return step.order.order != ORDER_FULL_MP
          && step.order.order != ORDER_ACTION_MOVE && step.turns < 1
-         && map_distance(m_start, step.location) <= m_distance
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Works for me.

PR #1558, opened 2022-12-04
Remove the animation from prod and research progress bars

lmoureaux:
An animation only makes sense if the bar is dynamic and progress is expected without user interaction. There is no reason whatsoever to animate progress bars that will only update during TC.


jwrober (approved):
I personally never had an issue with the animated progress bar. That being said, this patch does remove them.

PR #1563, opened 2022-12-05
Ensure that all options start at their default

lmoureaux:
Before loading options, set them all to their default values. This way, if anything goes wrong while loading, we fall back to the default. This is especially relevant for options like fonts whose default is determined at runtime.

**Test plan**

* Do `cmake --build build --target install` at least once to have the fonts in the install folder
* `rm ~/.local/share/freeciv21/freeciv-client-rc-3.0` or the Windows equivalent
* Start fc21
* Check that Libertinus is in use

May only work properly on top of #1547

### DIFF
--- a/client/options.cpp
+++ b/client/options.cpp
@@ -4359,6 +4359,11 @@ void options_load()
   const char *const prefix = "client";
   const char *str;
 
+  // Ensure all options start with their default value
+  for (auto &option : client_options) {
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Works great. Does not need 1547

daavko (approved):
Tested under Win10, works as advertised

PR #1565, opened 2022-12-07
Fix a client crash after reseting options

lmoureaux:
The last shortcut, SC_PILLAGE, was never added back to the map of shortcuts after clearing it for a reset, causing a crash when selecting any pillage-capable unit.

### DIFF
--- a/client/shortcuts.cpp
+++ b/client/shortcuts.cpp
@@ -336,7 +336,7 @@ void fc_shortcuts::init_default(bool read)
     suc = this->read();
   }
   if (!suc) {
-    for (int i = 0; i < SC_LAST_SC - 1; i++) {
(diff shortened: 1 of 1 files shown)


lmoureaux:
Testing:
* Load [freeciv-T0264-Y01776-auto.sav.gz](https://github.com/longturn/freeciv21/files/10179739/freeciv-T0264-Y01776-auto.sav.gz)
* Take `louis`
* Reset options
* Change production in `Měnĭskŭ` (top right of the map, next to the forest, doesn't matter what you change the production to)
* Close city dialog with Esc
* The engineer in Chelm is selected
* Hit W
* Crash (or not)

Make sure you can reproduce with the current version before claiming it's fixed...

jwrober (approved):

PR #1574, opened 2022-12-09
Update tool tip on allies/private button on chat

### LINKED ISSUE #1160: Allies only option not styled properly when checked
**Describe the bug**
The popup for the allies only checkbox always shows, "allies only". It is confusing given it isnt a check but has a icon change currently. 

**To Reproduce**
1. Open a game. 
2. Hover over allies only checkbox in chat widget

**Expected behavior**
The popup should say "broadcast globally" (or something along those lines), when allies only is enabled, and "allies only" when allies only is disabled. So that the user understands what that option goes. 

**Screenshots**
image
image

- master

jwrober:
closes #1160

### DIFF
--- a/client/chatline.cpp
+++ b/client/chatline.cpp
@@ -425,6 +425,11 @@ chat_widget::chat_widget(QWidget *parent)
         priv ? QLatin1String("public") : QLatin1String("private");
     cb->setIcon(fcIcons::instance()->getIcon(icon_name));
     gui_options->gui_qt_allied_chat_only = priv;
+    if (cb->isChecked()) {
+      cb->setToolTip(_("Allies only"));
+    } else {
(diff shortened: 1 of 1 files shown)


lmoureaux (approved):
Not sure I like how the first conditional uses `priv` and the second uses `cb->isChecked()`, but they should be identical for all practical matters.

jwrober:
I also noticed that the calls to `getIcon()` does not include `QStringLiteral` like we have in other areas of the code. Wonder if I should push a commit to change that.

lmoureaux:
> I also noticed that the calls to `getIcon()` does not include `QStringLiteral` like we have in other areas of the code. Wonder if I should push a commit to change that.

We have `QLatin1String` which should be equivalent

lmoureaux (approved):

PR #1580, opened 2022-12-10
Fix /timeoutshow when loaded from save

### LINKED ISSUE #1579: /timeoutshow reports wrong results when loaded from save
**Describe the bug**

When loading a game from a save, the timeout is restored. `/timeoutshow` doesn't understand this and reports wrong results.

**To Reproduce**
1. Start a game
2. Set timeout to a few minutes
3. Wait for some time
4. Save
5. Exit
6. Reload
7. Do `/timeoutshow` & compare to client display
8. See inconsistent results
9. Wait for the end of the timeout and conclude that the client was right

**Expected behavior**
`/timeoutshow` is correct

 - OS: all
 - Freeciv21 version: 3.0-beta7
 - Ruleset/Longturn game (if applicable): `LT75`

lmoureaux:
/timeoutshow was reporting incorrect results because it didn't take game.server.additional_phase_seconds into account.

Closes #1579.

**Test:** see #1579 repro steps

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.repository }}-${{ github.head_ref || github.run_id }}-{{ github.action }}
+  group: ${{ github.repository }}-${{ github.head_ref || github.run_id }}-build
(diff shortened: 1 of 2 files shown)


jwrober (approved):
Works for me

PR #1587, opened 2022-12-11
Don't disable the CMA before changing its parameters

### LINKED ISSUE #1586: Citizen Governor does not always update on screen when switching presets
**Describe the bug**
Occasionally when switching between governor presets on the Governor screen, the governor status switches to disabled, which seems to indicate that it failed, but when i close the city window and re-open it, it has switched to the new presets.

**To Reproduce**
This bug doesn't always occur when changing presets but I do see it quite frequently.  I've recorded a 30-second video of it occurring on a local game which shows that it does not occur at all in one city, but happens reliably with anpther.

**Expected behavior**
That the governor will reliably update immediately when switching, and only drop to "disabled" when it can't.

MAC OS 12.6
Freecic 21 Beta 7
LT75, local game under both LT75 and default rules

lmoureaux:
It's not needed and was causing visible UI glitches.

Closes #1586.

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -1632,12 +1632,8 @@ void city_dialog::cma_double_clicked(int row, int column)
     return;
   }
   param = cmafec_preset_get_parameter(row);
-  if (cma_is_city_under_agent(pcity, nullptr)) {
-    cma_release_city(pcity);
-  }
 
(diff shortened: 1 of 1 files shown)


jwrober (approved):
This is a much needed improvement.

PR #1592, opened 2022-12-11
Convert CRLF line endings to LF

daavko:
Ran the entire repo through `dos2unix`, this is what it spit out


lmoureaux:
Exclude `dependencies/sol2` to match upstream better. I'm worried about a few Windows-specific files (`freeciv21-server.cmd`, NSIS template). Otherwise looking good!

daavko:
Excluded `dependencies/sol2`. Also excluded `freeciv21-server.cmd`, I missed that one when looking through the list. Batch files don't work with LF endings.

lmoureaux:
Asking jwrober for the NSIS, but if someone can download the Windows installer and show that it works I'm good as well

jwrober (approved):
We need to alter the `.gitattributes` file so we don't have to deal with this all the time, but the PR is fine. NSIS installs with no issue.

PR #1595, opened 2022-12-12
Add detailed info about tech progress to top bar

lmoureaux:
Requested by horror-vacui on Discord.

See #1517.

image

### DIFF
--- a/client/sciencedlg.cpp
+++ b/client/sciencedlg.cpp
@@ -536,7 +536,6 @@ void real_science_report_dialog_update(void *unused)
 {
   Q_UNUSED(unused)
   int i;
-  int percent;
   science_report *sci_rep;
   bool blk = false;
   QWidget *w;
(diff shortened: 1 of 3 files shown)


psampathkumar:
Looking good.!

jwrober:
I would have asked for a little more whitespace between values in my review `28 (+35) / 120`. It looks squished to me as is.

lmoureaux:
Spacing as given by LaTeX:
image
Graying out the `(+35)` might help in the future.

lmoureaux:
Typographers have put a lot of work in making LaTeX look good, that's why I'm using it as a reference here. Notice for instance how the / and + aren't the usual / and + signs.

psampathkumar:
I think it looks good as is, especially for serif fonts, latex is usually very good reference. Not because we are writing for latex programmers, but latex was specifically designed for math/programmerish people, who are don't have to manually worry about such things. (or are bad at such things)

PR #1608, opened 2022-12-17
Close the city screen when closing reports

### LINKED ISSUE #1606: Dialog Weirdness
**Describe the bug**
Opening and closing the Nations, Cities, Economy or Research tabs when viewing city production causes the message, chat and minimap windows to re-open, in addition to the city windows.

**To Reproduce**
- Start a new local game using default rules.
- Build one city and click on it to open up the city production windows.
- Observe that the Chat, Messages and Minimap all disappear, to be replaced by three city-specific dialogs.
- Click one one of the Nations, Cities, Economy or Research tabs twice to open and close those windows.
- Observe that the Chat, Messages and Minimap are now visible, in addition to the three city dialogs.

**Expected behavior**
The Chat, Messages and Minimap should not be visible at the same time as the city dialogs.

 - OS: macOS Monterey 12.6
 - Freeciv21 version: 3.0.836246.7-beta

lmoureaux:
Otherwise we may get the minimap and messages overlapping with the city screen widgets. Closing the city screen is maybe not the best solution, but it does work around the inconsistency.

Closes #1606.

### DIFF
--- a/client/cityrep.cpp
+++ b/client/cityrep.cpp
@@ -30,6 +30,7 @@
 #include "mapview.h"
 #include "page_game.h"
 #include "qtg_cxxside.h"
+#include "top_bar.h"
 
 /**
    Overriden compare for sorting items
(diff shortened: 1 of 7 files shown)


jwrober:
The issue is still present for research (e.g. have messages and chat expanded, open a city, click research and click it again). Also national budget and messages can be opened to overlap with the city dialog. Not sure if we want that or not.

lmoureaux:
Good catch, one `setCurrentIndex(0)` had slipped through!

I think opening budget or messages isn't annoying since it's a direct result of user interaction.

jwrober (approved):

PR #1609, opened 2022-12-17
Prevent the autocenter option from autochanging

### LINKED ISSUE #1316: Auto center units option does not do anything
**Describe the bug**
If you have the local option "Auto center on units" checked, the client does not actually auto center on the selected unit.

**To Reproduce**
1. Open a saved game with multiple units available needing moves
2. Ensure the option is checked in your settings
3. Pick a unit and move it or click done on the control bar
4. Notice that another unit has been selected, but the map didn't move

**Expected behavior**
I expect the map to move to center the unit on the screen with this option enabled

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.0-beta3

**Additional context**
None

lmoureaux:
The "auto center on unit" option was being changed without user interaction because of badly engineered code that was temporarily changing its value and failing to restore it. Make the code safer by introducing a new, purely internal, flag to pause the auto centering so we don't need to touch the user-visible setting.

Closes #1316.

### DIFF
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -413,6 +413,15 @@ static struct tile *find_a_focus_unit_tile_to_center_on()
   }
 }
 
+namespace {
+bool auto_center_enabled = true;
+}
+
+/**
+ * Allows disabling automatic map recentering
+ */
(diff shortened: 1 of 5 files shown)


jwrober (approved):
LGTM

PR #1614, opened 2022-12-19
Remove python3 from macOS CI

jwrober:
Remove python3 from macOS CI since its included in the image and is causing a conflict.


lmoureaux:
I hope the image Python isn't configured in a weird way that would cause the installer to fail like on Windows...

PR #1620, opened 2022-12-21
Fix loading audio plugin and theme options

lmoureaux:
The list of available options wasn't initialized properly before loading.

**Test plan:**
Set Classic theme, check that it's picked up after restart. Check other options that use a combo box.


jwrober (approved):
This fixes the regression we talked about on Discord.

PR #1621, opened 2022-12-21
Fix inconsistent display of city happiness

### LINKED ISSUE #1616: "Happy" is not a state in city view or city dialog
Right now the city dialog and city view give the following city citizen "states": Celebrating, Disorder, Peace. In the city dialog we have "Content" as well. The LT community plays rulesets that need "Happy" for a turn to get to "Celebrating", but that isn't visible anywhere in the game.

In both the city view and the city dialog we should be able to see when a city goes to "Happy" as a player manipulates the citizens to entertainers to help get a city to "Celebrate". 

**Additional context**
Came up as part of a Discord conversation

lmoureaux:
In the city dialog, we had `Celebrating/Happy/Peace/Disorder`. The city list was missing `Happy`. Add it.

This is big because I mass-changed `const char *` to `QString` to allow code sharing between the city dialog and city list (which makes more sense than two separate functions). It's also safer.
The first commit contains the boilerplate, the second commit has the functional changes.

Closes #1616.

### DIFF
--- a/client/cityrep.cpp
+++ b/client/cityrep.cpp
@@ -138,17 +138,14 @@ bool city_item::setData(int column, const QVariant &value, int role)
  */
 QVariant city_item::data(int column, int role) const
 {
-  struct city_report_spec spec;
-
   if (role == Qt::UserRole && column == 0) {
     return QVariant::fromValue((void *) i_city);
   }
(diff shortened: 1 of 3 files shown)


jwrober (approved):
LGTM

PR #1626, opened 2022-12-22
Client Manual Update

### LINKED ISSUE #1624: Update client manual
We've made some changes to the client such as button embellishments on top bar, minimap, city dialog, etc. Also the right click menu on the city dialog production tab for worklists and options isn't documented.

**Do you have suggestions?**
As part of release of stable 3.0, the manual should be as accurate as possible.

jwrober:
Closes #1624 

- [x] update the city dialog images and text to bring current
- [x] include more details about the right-click menu on prod tab (e.g. for worklists and city options)


docs/Manuals/client-manual.rst
```diff
@@ -339,7 +350,7 @@ The Interface Options dialog is broken down into the following tabs:
 * :guilabel:`Map Image`: Freeciv21 can save summary images of the map every turn. This tab allows you to
   configure how you want to do that. Refer to `Game Menu`_ for more information.
 * :guilabel:`Font`: There is a collection of font styles used by the client. This tab allows you to tailor
-  them to your own favorites.
+  them to your own favorites. We ship with the Liberinus font set and it is the default.
```
daavko, line 353:
```suggestion
  them to your own favorites. We ship with the Libertinus font set and it is the default.
```

daavko (commented):
Found one typo, other than that looks good

PR #1628, opened 2022-12-23
Clarification on multiple Tech_parasite effects

Corbeau-1:
"Worse" is a matter of perspective. Having two effects could also be "better", depending what your goal is. Also, the explanation is unclear, it doesn't indicate how exactly this changes.


jwrober (approved):

jwrober:
Need to fix the CI in a subsequent PR. Adding this change to PR #1626.

PR #1632, opened 2022-12-26
Specify which action a sanity warning is emitted for

lmoureaux:
When emitting warnings about action enablers, include the action name in the message (unfortunately, we currently don't store the enabler section name). This gives some context to messages that would previously look like:

  This action enabler is never used by any unit.

**Test plan:**
Load the latest LTX (from [LTT-LTX](https://github.com/longturn/LTT-LTX)) with and without this change.


jwrober (approved):
LGTM

PR #1633, opened 2022-12-26
Fix a class of crashes in the city list

lmoureaux:
Modifying a city can (somehow) change the selection, upon which selected_cities is modified. Modifying a QList invalidates iterators, thus we were triggering undefined behavior when doing this in loops. Copy the list before doing anything with it, which avoids any invalidation (since we work on the copy).

Also remove a duplicated `nullptr` check (items in the list shouldn't be `nullptr` anyway).

**Test plan:**
* Select all cities
* Switch to a governor that works for some cities but not all
* Observe crash without patch (YMMV), no crash with patch

### DIFF
--- a/client/cityrep.cpp
+++ b/client/cityrep.cpp
@@ -424,7 +424,9 @@ void city_widget::clear_worlist()
   struct worklist empty;
   worklist_init(&empty);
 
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Works for me. I was able to recreate the crash and see it go away with the patch.

PR #1634, opened 2022-12-26
Fix Guangzhou Port Facility graphic

lmoureaux:


daavko:
Since you're already touching the file, could you also fix the wrong graphic used for `b.ecclesiastical_palace` a couple lines up? It currently uses the `-small` variant, it should use the regular one. Probably a copy/paste error

jwrober (approved):
I am seeing some other typos and CnP errors (per daavko's post). I'll make another PR to fix those.

PR #1635, opened 2022-12-26
Fix some wonder sprite issues

jwrober:


daavko (approved):
LGTM, works as advertised

PR #1636, opened 2022-12-27
Take team_pooled_research into account when looping over research

lmoureaux:
The number of valid research objects is the number of teams when team_pooled_research is enabled, and the number of players when it is disabled. The code was always assuming the first case, which created bugs in team games without pooled research.

Fixes 22c0c4b47ed3ed239358933af15818c7b4873c49.

Observed in the LT76Team test game.

**Test plan:**
* Check that research is working normally in single player
* Check that research is working normally in a team game with shared research (`/set team_pooled_r en`)
* Check that research is working normally in LT76Team (which will have this patch in a few minutes)

### DIFF
--- a/common/research.cpp
+++ b/common/research.cpp
@@ -120,6 +120,18 @@ struct research *research_get(const struct player *pplayer)
 }
 
 /**
+ * Checks whether the research object is valid in the current game.
+ */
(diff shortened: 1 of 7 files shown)


jwrober (approved):
This is working for me on all three scenarios.

PR #1637, opened 2022-12-27
Remove paths-ignore

jwrober:
We require build CI to complete as it is our repo protection mechanism. We had paths-ignore to not run full CI on documentation only updates (docs/**), however that doesn't really work. You have to go through a collection of hoops to actually get it to satisfy the check. Removing as that is easier and keeps our protection mechanism in place.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,11 +1,5 @@
 name: Build
 on:
-  push:
-    paths-ignore:
-      - 'docs/**'
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
   schedule:
     - cron: "0 5 * * MON"
   workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -1,11 +1,6 @@
 name: Build
 on:
-  push:
```
lmoureaux, line 3: Why remove `push`?

jwrober: I figured it best to just get rid of all of it, even on a push

lmoureaux: So it won't run when pushed on master. Will it run when pushing to the branch of a PR?

jwrober: All runners run all the time, except for concurrency.

lmoureaux: ok I'll take your word for it :D

PR #1638, opened 2022-12-27
Check cache fc_version.h at configure time

### LINKED ISSUE #1564: fc_version.h not updated automatically
**Describe the bug**

fc_version.h isn't updated automatically when fc_version.h.in is modified or the version changes, potentially leading to inconsistent builds.

**To Reproduce**
1. Build fc21
3. Modify fc_version.h.in
5. Rebuild (do **not** do a clean build)
7. Nothing is rebuilt

**Expected behavior**
fc_version.h gets updated and many files get rebuilt.

 - OS: all
 - Freeciv21 version: 3.0-beta.6

**Additional context**
This caused issues updating the server at longturn.net.

jwrober:
For devs who do not clean their cache directory often, we do a check for the generated `fc_version.h` and if the patch version is different then we delete the cache file. Closes #1564

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ include(AutoRevision)
 list(LENGTH FC21_REV_TAG_LIST FC21_REV_TAG_LIST_LENGTH)
 
 # When we tag a stable release we only get 2 of the 4 components populated
-#  So we manually set some of the variables we need for the full version.
+#   So we manually set some of the variables we need for the full version.
 if(${FC21_REV_TAG_LIST_LENGTH} EQUAL 2)
   list(GET FC21_REV_TAG_LIST 0 FC21_MAJOR_VERSION)
   list(GET FC21_REV_TAG_LIST 1 FC21_MINOR_VERSION)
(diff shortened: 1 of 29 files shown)


lmoureaux:
I'm afraid that this will trigger a full rebuild after every local commit, which is definitely too much, but maybe I misremember what we put in the version string?

jwrober:
> I'm afraid that this will trigger a full rebuild after every local commit, which is definitely too much, but maybe I misremember what we put in the version string?

I didn't test that scenario. Let me see what happens.

jwrober:
It works while you are working on a bunch of local changes. Multiple commits are fine in a branch while you are working. You might run into an issue if you switch to a different branch that has a diff hash from master branch, that will force a "mostly" recompile, but not a total recompile. The code is essentially getting the current hash turned into a integer from the last pull, so based on what that is in the branch you are working on would cause a recompile. I think this better than ignoring it altogether like we did in the past. One recompile with a fresh `git pull` is not horrible.

lmoureaux:
> You might run into an issue if you switch to a different branch that has a diff hash from master branch, that will force a "mostly" recompile, but not a total recompile. 

I switch between branches all the time, so this will slow me down significantly :( I think we keep this out for now and revisit after 3.0 is out, moving the info to a file that gets compiled exactly once (and fast).

jwrober:
Yea I was thinking something about that too. I wonder if it would be better to check the tag and do the removal of the cached file when a new tag is done. That would be a more rare event and allow branch switching.

lmoureaux (changes requested):
Just to prevent immediate merge

lmoureaux (approved):
Fixed my concerns and rebased for the new CI checks

PR #1640, opened 2022-12-28
Fix infinite recursion in research_is_valid

lmoureaux:
Spotted when loading the tutorial or the Polish game


jwrober (approved):
I wasn't able to get into an infinite loop, but this does look good to me.

PR #1641, opened 2022-12-28
Enable combat messages by default

### LINKED ISSUE #1627: Combat info
Combat info does not show in messages even if activated and player is online.

lmoureaux:
Users were confused that they weren't shown.

Closes #1627.

### DIFF
--- a/client/options.cpp
+++ b/client/options.cpp
@@ -3386,8 +3386,6 @@ static void message_options_init()
   int none[] = {E_IMP_BUY,
                 E_IMP_SOLD,
                 E_UNIT_BUY,
-                E_UNIT_LOST_ATT,
-                E_UNIT_WIN_ATT,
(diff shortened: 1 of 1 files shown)


jwrober (changes requested):
So we have an option in the view menu to enable/disable it and also have an option in the game -> messages dialog? Why do we have it in two places. If it's disabled in the view menu, should be disabled in messages and vice versa. No wonder it "got lost" and we thought it was broken.

lmoureaux:
The option in the view menu controls the widget that appears on top of the map. Maybe we need to reword that one, or remove it from the menu and hide it in the settings?

jwrober:
That is then confusing (even tripped me up it seems). I'll open another issue for that.

jwrober (approved):

PR #1653, opened 2022-12-28
The value of idle stays at 0 until it turns 3 #1643 FIX

### LINKED ISSUE #1643: The value of idle stays at 0 until it turns 3
**Describe the bug**
The idle column in nations view stays at zero until 3 turns have passed and then it updates in the view.

**To Reproduce**
1. Play a game with more than one player and at least one of them not assigned an AI (e.g. an idler)
2. Enable the column in the nations view
3. Notice it starts at zero
4. Play a couple turns and eventually notice that after 3 turns it changes

**Expected behavior**
The counter should change after the first idle turn, not wait until 3

**Screenshots**
n/a

 - OS: All, tested on Linux
 - Freeciv21 version: 3.0-beta.7+

**Additional context**
Discord convo - https://discord.com/channels/378908274113904641/1049684182848262155/1049684182848262155

@psampathkumar states: "This is a good first issue for someone wanting to get into development. The answer is in col_idle function in **plrdlg_common.cpp**

Sokolan:
new:
```c++
static QString col_idle(const struct player *plr)
{
  int idle;
  static char buf[100];
  //nturns_idle >= 0
  if (plr->nturns_idle == 0) {
    idle = 0;
  } else {
    idle = plr->nturns_idle - 1;
  }
  fc_snprintf(buf, sizeof(buf), "%d", idle);
  return buf;
}
```

old:
```c++
static QString col_idle(const struct player *plr)
{
  int idle;
  static char buf[100];

  if (plr->nturns_idle > 3) {
    idle = plr->nturns_idle - 1;
  } else {
    idle = 0;
  }
  fc_snprintf(buf, sizeof(buf), "%d", idle);
  return buf;
}
```

Fixes #1643

### DIFF
--- a/client/plrdlg_common.cpp
+++ b/client/plrdlg_common.cpp
@@ -267,7 +267,7 @@ static QString col_idle(const struct player *plr)
   int idle;
   static char buf[100];
 
-  if (plr->nturns_idle > 3) {
+  if (plr->nturns_idle > 0) {
     idle = plr->nturns_idle - 1;
   } else {
     idle = 0;
--- a/client/plrdlg_common.cpp
+++ b/client/plrdlg_common.cpp
@@ -266,7 +266,7 @@ static QString col_idle(const struct player *plr)
 {
   int idle;
   static char buf[100];
-  //nturns_idle >= 0
+  // nturns_idle >= 0
   if (plr->nturns_idle == 0) {
     idle = 0;
   } else {
--- a/client/plrdlg_common.cpp
+++ b/client/plrdlg_common.cpp
@@ -266,11 +266,11 @@ static QString col_idle(const struct player *plr)
 {
   int idle;
   static char buf[100];
-
-  if (plr->nturns_idle > 0) {
-    idle = plr->nturns_idle - 1;
-  } else {
+  //nturns_idle >= 0
+  if (plr->nturns_idle == 0) {
     idle = 0;
+  } else {
+    idle = plr->nturns_idle - 1;
   }
   fc_snprintf(buf, sizeof(buf), "%d", idle);
   return buf;
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Editor-based HTTP Client requests
-/httpRequests/
-# Datasource local storage ignored files
-/dataSources/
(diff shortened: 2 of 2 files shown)


client/plrdlg_common.cpp
```diff
@@ -266,11 +266,11 @@ static QString col_idle(const struct player *plr)
 {
   int idle;
   static char buf[100];
-
-  if (plr->nturns_idle > 3) {
-    idle = plr->nturns_idle - 1;
-  } else {
+  // nturns_idle >= 0
```
lmoureaux, line 269: I think this is old code? If so, I would prefer if you could remove it. If it's a comment, I don't understand what it's trying to say (and I don't think a comment is necessary, but fine to add one as well).

client/plrdlg_common.cpp
```diff
@@ -266,11 +266,11 @@ static QString col_idle(const struct player *plr)
...
+  if (plr->nturns_idle == 0) {
     idle = 0;
+  } else {
+    idle = plr->nturns_idle - 1;
   }
   fc_snprintf(buf, sizeof(buf), "%d", idle);
```
lmoureaux, line 275: Looks like the whole `if` clause could be replaced by `max(0, plr->nturns_idle - 1)`. But it's also good as it is now.

lmoureaux (changes requested):
Hi @Sokolan, thank you for your fix and congrats for passing the `clang-format` bot! :1st_place_medal: 
In case you're not familiar with Github, you can see your changes in the `Files changed` tab at the top.

I would prefer not to introduce a new folder just for IntelliJ IDEA (it will quickly fall out of date if nobody using IDEA contributes for a while). We already have some IDE-specific paths in the [main `.gitignore`](https://github.com/longturn/freeciv21/blob/master/.gitignore), maybe you can add `/.idea/` there?

I also have small comments on the code, see below :arrow_double_down:

I'm going to modify the description to include the magic sentence `Fixes #1643` to link the issue with the pull request.

jwrober:
Another contributor! Yay.  Thanks for the submission. Another note for your next pull, we prefer to have all changes come from another branch instead of `master` as documented here - https://longturn.readthedocs.io/en/latest/Contributing/pull-request.html. Not required for this patch, but would ask it for future ones.

psampathkumar:
Welcome :) 
Also, remember to add your name to the AUTHORS list if the PR gets merged :)

client/plrdlg_common.cpp
Sokolan: If you're talking about line 269:
// nturns_idle >= 0
Then it's not the old code but a commentary that the value isn't negative (as it's declared int and not unsigned int in the player struct), after checking the usages of nturns_idle and seeing that it's never being decreased but only increased or set to 0.
If you referring to something else then excuse me for misunderstanding.

lmoureaux: Aah ok, I see now that the code would behave weirdly for negative values. Sounds good then!

lmoureaux (approved):
LGTM

PR #1661, opened 2022-12-30
Don't advance unit focus when the city dialog is shown

### LINKED ISSUE #1395: Giving a command to a unit inside city view activates the unit
**Describe the bug**
When you have an active unit, open any city and give a command to any units within the city, a unit gets activated within the city view.

**To Reproduce**
1. Have a city with units in it
2. Have an active unit, doesn't have to be in the city (best tested with multiple units that haven't received a command this turn)
3. Open city view for the city
4. Give any unit a command via the right-click menu for the unit
5. Observe a unit gets activated (most of the time, it's the "next active unit in sequence", but sometimes it's the unit that received the command)

**Expected behavior**
Units shouldn't activate while in city view after receiving a command. The only exception to this should be the "Activate unit" command, which should either mark the unit as active so it can be chosen automatically when cycling through units, or it should close the city view. Best would probably be to have two separate items, "Activate unit" and "Activate unit and close city view".

 - OS: Windows 10
 - Freeciv21 version: beta 4
 - Ruleset/Longturn game (if applicable): LT73Avi, again it's very annoying

**Additional context**
No additional context

lmoureaux:
Closes #1395.

### DIFF
--- a/client/packhand.cpp
+++ b/client/packhand.cpp
@@ -76,6 +76,7 @@
 #include "attribute.h"
 #include "audio.h"
 #include "chatline_common.h"
+#include "citydlg.h"
 #include "citydlg_common.h"
 #include "client_main.h"
 #include "climap.h"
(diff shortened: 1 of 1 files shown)


jwrober:
What if I want to activate more than one unit? Right now when I right-click on a unit in the city and pick Activate, the dialog closes immediately.

lmoureaux:
* Select more than one:
   image
  Or use Ctrl+click and Shift+click.
* Then Right click and activate them.

jwrober (approved):
LoL, I had that documented in the client manual!

PR #1663, opened 2022-12-30
Fix bugs with the units actions menu in the city dialog

### LINKED ISSUE #1615: Add set home city to present units
There is a right-click menu of actions that a unit can do in the present units panel in the city dialog. It is missing "set home city" as an option. So to do that function you have to activate the unit, which closes the city dialog, press "h" to set home city and then click on the city again to open it up again if needed.

Add an option to set the home city on the right-click menu.

**Additional context**
Came up in a Discord conversation - https://discord.com/channels/378908274113904641/1054416902274949230

lmoureaux:
* "Set homecity" was hidden when right clicking without a selection and never shown again afterwards
* Initially, clicking anywhere on the widget (except on units) would show the full menu with all options enabled (but none working)

This patch fixes the above and changes the menu so items are enabled or disabled depending on available options, instead of hiding unavailable ones. This enables better visual memory of the items location.

Closes #1615.


jwrober (approved):
Looks good to me, especially with the fix in #1661

PR #1678, opened 2023-01-02
Add an effect to keep food surplus when a city grows

lmoureaux:
Tested by adding the following to `classic/effects.ruleset`:
```ini
[effect_foo]
type    = "Growth_Surplus_Pct"
value   = 100
```

Closes #510.

PR #1679, opened 2023-01-02
Always send wonder information to the player's team

### LINKED ISSUE #1656: Small wonder luxury bonus not documented in popup text
The General tab of the city production windows show various breakdowns of the stats such as corruption penalties, government bonuses, etc.  One piece of information missing from this is the luxury bonus from small wonders like the Hypogeum or Hanging Gardens.  These bonuses show up in the stats, but not in the tooltip, so there's no explanation of where the luxury bonus comes from.

Adding data to the tooltip indicating where the luxury bonus comes from would make it clear what is contributing to that bonus.

**Additional context**
This is a screenshot from a local game with the LT75 ruleset of a city with the Hypogeum and Hanging Gardens built.  The luxury value is 10, which is correct, but there's no information in the tooltip saying where this value is coming from.  Logically it should say +6 from the Hypogeum and +4 from the Hanging  Gardens.

Screenshot 2022-12-29 at 15 15 06

### LINKED ISSUE #1675: Temple of Artemis bonus is lost when selecting Citizen Governor preset
**Describe the bug**
The Temple of Artemis bonus is not applied when selecting a CV preset in LT75.

**To Reproduce**
I can only see this bug in LT75.  I tried to replicate the same issue in a local game with LT75 rules, but it did not appear.

In LT75, with the citizen governor and citizens tabs open, if I have a preset selected I can see that I'm only getting 2 content citizens from buildings in cities with a temple.  Disabling the governor does not fix this, but by manually changing the city workers, the bonus kicks in and buildings now provide 4 content citizens.  Selecting another preset disables this issue again

**Expected behavior**
With a temple and Temple of Artemis in a city, buildings should provide 4 content citizens with the governor preset enabled.

**Screenshots**
As this is an on-going multiplayer game, I will only provide screenshots privately through Discord upon request.

-  MacOS 12.6
- FC21 Beta 7

lmoureaux:
With the introduction of EFT_WONDER_VISIBLE, it became possible not to send (small) wonder information to the team owning them. This is obviously not the right thing to do, since the players can see the wonders in the cities. So always send the info.

Closes #1656.
Closes #1675.
For #1675, unfortunately the governor results will be wrong until this is implemented on the server.

Backport candidate.

### DIFF
--- a/server/plrhand.cpp
+++ b/server/plrhand.cpp
@@ -1207,7 +1207,7 @@ static void package_player_info(struct player *plr,
       const auto bonus = get_target_bonus_effects(
           nullptr, city_owner(pcity), receiver, pcity, pimprove, nullptr,
           nullptr, nullptr, nullptr, nullptr, nullptr, EFT_WONDER_VISIBLE);
-      if (bonus > 0) {
+      if (players_on_same_team(plr, receiver) || bonus > 0) {
         packet->wonders[i] = plr->wonders[i];
       } else {
         packet->wonders[i] = WONDER_NOT_BUILT;


lmoureaux:
Looks like this was included by mistake in #1678 (already merged).

PR #1680, opened 2023-01-02
Don't prompt to remove previous version on windows

jwrober:
No related issue.

Highly recommend this to backports stable


lmoureaux (approved):
Will this also uninstall previous-to-previous versions?

Backporting approved.

jwrober:
I don't recall when I made the change, but awhile ago - at @daavko's request - I stopped installing into a version specific directory and started installing on top of the old version. So this change will not get those version specific installs, it will only get the one installed to the non-version specific location.

PR #1689, opened 2023-01-04
Show unit vision in help dialog for units

### LINKED ISSUE #1644: Add unit vision to in game help
The in game help does not tell you what the vision radius is for units.

Add this ruleset feature to the game help.

**Additional context**
Discord https://discord.com/channels/378908274113904641/1052690193943121920/1052690193943121920

daavko:
Fixes #1644.

If I understand correctly, `static_cast<whatever>` is preferred over `(whatever)`, please do correct me if I'm wrong about that.

### DIFF
--- a/client/helpdlg.cpp
+++ b/client/helpdlg.cpp
@@ -886,6 +886,9 @@ void help_widget::set_topic_unit(const help_item *topic, const char *title)
                       utype_build_shield_cost_base(max_utype));
     add_info_progress(_("Firepower:"), utype->firepower, 0,
                       max_utype->firepower);
+    add_info_progress(
+        _("Vision:"), static_cast<int>(std::sqrt(utype->vision_radius_sq)),
+        0, static_cast<int>(std::sqrt(max_utype->vision_radius_sq)));
 
     // Upkeep
     upkeep = utype->upkeep[O_FOOD] + utype->upkeep[O_GOLD]


client/helpdlg.cpp
```diff
@@ -886,6 +886,12 @@ void help_widget::set_topic_unit(const help_item *topic, const char *title)
                       utype_build_shield_cost_base(max_utype));
     add_info_progress(_("Firepower:"), utype->firepower, 0,
                       max_utype->firepower);
+    add_info_progress(
+        _("Vision:"),
+        static_cast<int>(sqrt(static_cast<double>(utype->vision_radius_sq))),
+        0,
+        static_cast<int>(
+            sqrt(static_cast<double>(max_utype->vision_radius_sq))));
```
lmoureaux, line 894: Doing `std::sqrt(x)` (note the `std::` forcing usage of the C++ function, as opposed to C) is equivalent to `sqrt(static_cast<double>(x))`. If precision isn't an issue, `sqrt(x)` will also work just fine (might just cast to `float`, not sure what the rules are).

lmoureaux:
And you're right that `static_cast<>` is preferred over C-style cast `(type)` :+1:

client/helpdlg.cpp
daavko: Should I change this to `std::sqrt()` without the inner cast then?

lmoureaux: Would be more readable

lmoureaux (approved):
Noting that it will round down, but rounding up wouldn't be better

PR #1693, opened 2023-01-04
Ask QTextStream to decode text as UTF-8

### LINKED ISSUE #1692: Nation infos with special characters are broken
**Describe the bug**
As an example, let's take the Danish, with the capital city of København. The first issue is that the suggested name is "KÃ¸benhavn". This broken version also gets saved to the savefile. It then gets progressively worse on each save-load cycle. For example, after the third time of saving and loading the game, the name is now "KÃƒÆ’Ã‚Â¸benhavn". I assume this will go on until the save is unusable due to length limitations.

Even worse: Even if you manually fix the suggested name (here to København), it gets broken anyway when the game is saved and then loaded.

Looks like the "suggested name is wrong" part only affects local games, in LT76 I was able to build a city with a special character and it auto-suggested correctly. Unsure if the "save and load breaks it more" part affects running LT games, I'm unable to test that.

This seems to apply to leaders as well, leader names with special characters are broken in the pre-game nation selector.

**To Reproduce**
1. Choose the Danish in a new game
2. See that ruler names are broken
3. Build a city
4. See that the suggested name is "KÃ¸benhavn" instead of "København"
5. For bonus points, save and load the game several times (different saves, not overwriting the same one every time)
6. See that the name gets worse and worse every time you load the newest save

**Expected behavior**
I expect the name to stay as-is

**Screenshots**
Initial name suggestion
freeciv21-client_XnfBI4RfLO
Increasingly broken name
freeciv21-client_63TT6xKo5N
freeciv21-client_FNo4SfBGJu
freeciv21-client_8otZ8QI8o3

 - OS: Windows 10
 - Freeciv21 version: 3.0-rc1
 - Ruleset/Longturn game (if applicable): not applicable, probably

**Additional context**
Savefile in question
[kabenhavn.zip](https://github.com/longturn/freeciv21/files/10341190/kabenhavn.zip)

lmoureaux:
Also allow wider Unicode encodings in case a BOM is present.

Closes #1692.

### DIFF
--- a/utility/inputfile.cpp
+++ b/utility/inputfile.cpp
@@ -233,6 +233,8 @@ struct inputfile *inf_from_stream(QIODevice *stream,
   inf->filename.clear();
   inf->fp = stream;
   inf->stream = new QTextStream(stream);
+  inf->stream->setCodec("UTF-8");
+  inf->stream->setAutoDetectUnicode(true); // Allow UTF-16 and UTF-32
   inf->datafn = datafn;
 
   qCDebug(inf_category) << "opened" << inf_filename(inf) << "ok";


daavko (approved):
LGTM, works correctly

lmoureaux:
Thanks!

Issue #1097, opened 2022-07-09
Organize the client code in folders


lmoureaux:
The client code should be organized in semantic folders rather than a single flat directory.

> Take this chance to move layers to a separate folder ? I think something like, client/layers/layer_*.cpp/h is better. But I'll approve anyways, since that can be done in a different PR, where we reorganize everything within client. Apart from that, rest look good to me.

_Originally posted by @pranavtifr in https://github.com/longturn/freeciv21/pull/1083#pullrequestreview-1033535980_

jwrober:
We need some proposals, so added the question flag.

lmoureaux:
Loose categories that come to mind:

* Widgets (utility widgets like the turn done button)
* Map view (the widget and support classes)
* Rendering engine (the actual rendering engine, i.e. the things that draw the map and schedule updates)
* Tileset (the definition of the tileset)
* Music
* Actions (things the player can do)
* Help
* Theming
* Networking & server management (start/stop the server, scenarios, ...)

PR #1695, opened 2023-01-04
Start breaking the city dialog

lmoureaux:
This PR introduces a unified widget to show everything supported by a city: improvements and units. It adopts a purely vertical layout that should be easy to scroll. All items respond to double click and right click. Eventually this widget would be added to a side of the city screen with as much height as possible. I want to add the widget to master early to gather feedback, so this PR adds it to the left of the 3.0 city screen layout.

Some infrastructure work for better tileset changing support is also done in this PR. I also started using a more structured folder layout with #1097 in mind.

image

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -133,6 +133,9 @@ target_sources(
   voteinfo_bar.cpp
   widgetdecorations.cpp
   unithudselector.cpp
+
+  widgets/city/upkeep_widget.cpp
+
   # Generated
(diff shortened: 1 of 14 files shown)


jwrober:
As it goes from a compilation and usage scenario, this PR looks good. Since it also starts a big change to the city dialog, I would like to hold on approving to see if the community has comments on Discord.

jwrober (approved):
On second thought, let's just merge and go from there.

PR #1710, opened 2023-01-07
Set the connection state before starting the metaserver scan

### LINKED ISSUE #785: Minor bug: occasionally can't enter password when connecting to lt.net game
Sometimes when I try to connect to a LT game, I click "Connect", get a message "Please enter your password", the field is ungreyed and cursor appears, only for the field to be greyed again and cursor to disappear a moment later.

Happens maybe 20% of the time, retrying usually solves the problem.

Windows version 3.0.20211008.

image

lmoureaux:
Setting the connection state resets, among others, the password field. The metaserver scan runs synchronously with its own nested event loop.

If a user was fast enough to start connecting before the scan was finished, the password field would be reset when the scan finished.

Closes #785.
This is the patch proposed by @daavko.

### DIFF
--- a/client/fc_client.cpp
+++ b/client/fc_client.cpp
@@ -247,9 +247,9 @@ void fc_client::switch_page(int new_pg)
     break;
   case PAGE_NETWORK:
     qobject_cast<page_network *>(pages[PAGE_NETWORK])
-        ->update_network_lists();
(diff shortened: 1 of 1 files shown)


lmoureaux:
Backporting suggested

jwrober (approved):
Excellent idea by @daavko. Works great.

jwrober:
I agree with backport

PR #1729, opened 2023-01-13
Implement hex_corner with MATCH_SAME

lmoureaux:
I need this for a hexagonal tileset project I'm doing on the side. Vectron might also benefit from this feature.

The patch touches the drawing of most terrain, all bundled tilesets seemed to work but always good to check.

### DIFF
--- a/client/layer_terrain.cpp
+++ b/client/layer_terrain.cpp
@@ -770,6 +770,29 @@ int layer_terrain::terrain_group(const terrain *pterrain) const
   return m_terrain_info.at(terrain_index(pterrain)).group->number;
 }
 
+namespace /* anonymous */ {
+
+/**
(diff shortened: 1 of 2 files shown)


jwrober (commented):
Code seems to work fine. Docs table is a bit wonky. Wonder if we need a css change to make the page wider?

image

lmoureaux:
> Docs table is a bit wonky.

That was already the case before, but the PR does make it wider. May need to break it down into a list. Should I do it in this PR?

jwrober:
Oh if it was already like that, then it can stay. Might look to fix it at a later date. Maybe add an issue as a reminder.

lmoureaux:
Rebased to allow merging

jwrober (approved):

Issue #1674, opened 2023-01-01
Flatpak/snap packaging


lmoureaux:
**Is your feature request related to a problem? Please describe.**
Our distribution for Linux currently includes only a Debian package for the latest Ubuntu LTS. Other/older distros are left out. We aren't included in any app store.

**Describe the solution you'd like**
Publish an official snap or flatpak on the main platform.

**Describe alternatives you've considered**
Package Freeciv21 for every possible distribution. I got one running (OpenSuse RPM) and we have an PKGBUILD for Arch, but it's a major pain (especially outdated distros). It will also not bring us to the app stores.

jwrober:
So is the idea here to NOT provide a `.deb` package any longer and just do a snap for all Linux distros? I am not familiar with flatpack.

lmoureaux:
It could replace the `.deb`, but we can still support it since it's well in place by now.

jwrober:
https://phoenixnap.com/kb/flatpak-vs-snap-vs-appimage

lmoureaux:
In practice both snap and flatpak are integrated with most modern distros.

jwrober:
I looks like snap is on most of them including Arch, RHEL and Fedora. The last two we don't support since we don't issue an `RPM`. It looks like legacy freeciv has gone the flatpack route.

jwrober:
Closing via #1911, #1912, #1916, and #1918

PR #1733, opened 2023-01-15
Support building snap images

jwrober:
This is a second attempt from #1713, part of #1674 

Upgrades to use 128x128 icons for higher fidelity. Corrects some license attribution errors.

Should be backported to stable.

What works:
- fc21:// links
- Modpacks downloaded are discovered properly
- Loading the tutorial
- Connected to LT76 just fine
- Playing a local game

Known issues:
- Not related to the snap, but we don't set a user name in the options rc file at "new". Until fixed a user has to go into options and set a user name on network tab
- The top level `icon:` directive seems to override the value for the icons in the .desktop files.
- If you have saved games in `~/.local/share/freeciv21` they won't be found.


dist/snapcraft.yaml
```diff
@@ -0,0 +1,123 @@
...
+name: freeciv21
+title: Freeciv21
+summary: Freeciv21 - Freeciv for the 21st Century
+description: |
+  Freeciv21 is a free and open source empire-building strategy game inspired by the history of human civilization. The game commences in prehistory and your mission is to lead your tribe from the Stone Age to the Space Age.
+  This snap package contains everything you need to play local single-player games against the computer as well as connect to online multi-player games against human opponents.
+  Freeciv21 is based on the Qt framework and brought to you by the team at Longturn.net.
+
+# TODO: At some point, will want to integrate into cmake process and configure_file()
+version: '3.0-rc.1'
```
lmoureaux, line 28: Let's make an issue for this right away.

lmoureaux (approved):
So setting `FREECIV_DATA_PATH` was not needed in the end?

jwrober:
> So setting `FREECIV_DATA_PATH` was not needed in the end?

Correct. The native paths work fine.

dist/snapcraft.yaml
jwrober: #1735

PR #1745, opened 2023-01-15
Use SSL v3 libraries for MSYS2

jwrober:
This change will fix the CI. We need upstream MSYS2 team to fix `QSslSocket` so the modpack installer will work.

I will want to push some more commits for vcpkg here too as its related.

### DIFF
--- a/cmake/FreecivInstall.cmake
+++ b/cmake/FreecivInstall.cmake
@@ -55,15 +55,15 @@ if(MSYS OR MINGW)
   if("$ENV{MSYSTEM}" STREQUAL "MINGW32")
     install(
       FILES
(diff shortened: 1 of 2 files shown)


jwrober:
Looks like Visual Studio/vcpkg is still failing. Looks like whatever the GH action is pulling from has not been updated from recent merge that fixed the issue upstream. Let's see what happens with macOS and MSYS2.

lmoureaux:
The new packages are `Ready for Upload` on the MSYS2 side, but I don't think they've been uploaded

jwrober:
I noticed that as well. We might need to wait a few more days on this one to allow upstream time to get updated.

lmoureaux (approved):
LGTM

Issue #1498, opened 2022-11-17
Change units view from a widget to a page/table


jwrober:
**Is your feature request related to a problem? Please describe.**
Right now, when you click on the units button on the top bar you get a transparent widget
![image](https://user-images.githubusercontent.com/23037938/202522862-54f6a08f-5e43-4bf8-aa5a-33bfc31297f1.png)

The widget provides little true functionality except to tell you how many of what you have and how many are under construction.

**Describe the solution you'd like**
The legacy freeciv Gtk client has the units view as a page in a table format.
![image](https://user-images.githubusercontent.com/23037938/202523258-3392415f-e004-46a9-b3b5-89c9486cd243.png)

This provides much more functionality such as giving totals and giving options for some buttons like find nearest. We can't do that in a widget overlay, but can on a table page like we do for cities view.

**Describe alternatives you've considered**

**Additional context**
More conversation as part of the lead up to LT75

PR #1747, opened 2023-01-16
Remove transparency from a few widgets

### LINKED ISSUE #1739: Make partially transparent background for widgets optional
It's a little distracting to see the background behind widgets like Units or Demographics, especially when screenshots of such are shared in team games. It's also not consistent with Messages and Chat widgets, those are completely opaque.

Best by me would be adding a slider that controls how transparent you want the widgets to be, which would then apply  to *all* the widgets you can open (Units, Demographics, Wonders of the World, Historian reports, Messages, Chat, ...). Then people can just choose how transparent they want the widget to be. Someone might like being able to see the map.

Two other alternatives:
- A simple on-off option
  - Probably simpler than variable transparency
- Leaving it as-is
  - Please no

**Additional context**
Short conversation on Discord
https://discord.com/channels/378908274113904641/912500712833974322/1063049102675419136

lmoureaux:
Affected widgets:
* Demographics
* UWT table (under F2)
* Middle click popup
* Unit combat report
* Unit actions HUD

`info_tab` doesn't exist so I removed it from the styles.

Closes #1739.

### DIFF
--- a/data/themes/gui-qt/Classic/resource.qss
+++ b/data/themes/gui-qt/Classic/resource.qss
@@ -57,7 +57,7 @@ QMainWindow, minimap_panel {
 }
 
 *[unit_wait_table="true"] {
-  background-color: rgba(15,10,5,175);
+  background-color: rgb(15,10,5);
   border: 1px solid #685f47;
   color: white;
 }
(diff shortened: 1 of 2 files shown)


jwrober:
Looks good. What about the units widget (F2)? I see it is still transparent. I think being consistent is a good idea.

lmoureaux:
I tried and failed because of the weird nested widgets there. I also don't want to invest too much in this widget since we have #1498.

jwrober (approved):
I see. Makes sense. Will need to rebase after #1745 is complete.

PR #1748, opened 2023-01-16
Make passing a fc21:// URL to the client trigger autoconnect

### LINKED ISSUE #1743: Using fc21:// links should assume autoconnect (-a) is used
When using a fc21:// URI to load the game from a web page, you have to click connect to network game and then the connect button to get freeciv21 to fully utilize the connection. If you use fc21:// from a command line, you can add "-a" as an option and you can log in to a game with a single command or "click" of a script file.

Set things up so that when fc21:// protocol is used, the game assumes that autoconnect (-a) is also part of the command line.  This will aid us putting links on the main LT webpage. Could include logged on user to the link dynamically that way.

lmoureaux:
When invoking the client by clicking on a fc21 link, the expected behaviour is that it tries and joins the linked-to game. This is only possible if passing a URL implies -a.

Suggest backporting this to `stable`.

Closes #1743.

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -480,6 +480,7 @@ int client_main(int argc, char *argv[])
         qFatal("%s", qPrintable(url.errorString()));
       }
     }
+    auto_connect = true;
   } else if (!positional.isEmpty()) {
     qFatal(_("Too many positional arguments."));
   } else {


jwrober (approved):
Works great. I tested a URI with userid and password in it (one click all the way into a game) and also with just a userid (one click + password entry into a game).

PR #1751, opened 2023-01-18
Update hexemplio rivers and roads for better contrast

jwrober:
No related issue.


lmoureaux:
I like it! Maybe the border could be a bit more subtle, especially on top of forests (old color or 25% transparent?) and irrigation could also be darkened a slightly (though not as much)

image

PR #1757, opened 2023-01-21
Add government icons to the players dialog

lmoureaux:
Make the players dialog more attractive by decorating the table with government icons. Also change the widget to be a table instead of a tree, since none of the tree features were actually used.


jwrober (approved):
Looks great.

PR #1759, opened 2023-01-21
Don't update the chat completion list with every new message

lmoureaux:
This caused performance issues when connecting. Instead, update the list whenever the chat input gets focus (so never if it's never used).

Used the opportunity to add relevant strings for translation and use `QStringLiteral()` on the others. Also removed O(N^2) duplicate-checking code when adding players to the list.

This fixes the "crocodiles" issue when connecting to long-running games. Backport recommended.


jwrober (approved):
Wow, big improvement in connecting to an online game. I like the backport idea too.

PR #1760, opened 2023-01-21
Fix maxconnectionsperhost

### LINKED ISSUE #1754: maxconnectionsperhost treats all users as coming from the same host
**Describe the bug**
The option `maxconnectionsperhost` is supposed to limit the number of concurrent connections coming from a single host. On the LT server at least, it treats all connections as coming from the same host and thus acts as a limit to the total number of players connected. Extra players get kicked immediately without any logging.

**To Reproduce**
1. Run LT76Team on the lt.net server

**Expected behavior**
The option works as described in the first sentence.

 - OS: lt.net server
 - Ruleset/Longturn game (if applicable): LT76Team

lmoureaux:
It was doing the exact opposite of the intended behavior, allowing an infinite number of connections from the same host but disallowing any new connection once N players were connected.

Closes #1754.

Also upgrade the rejection message to a warning so it appears in the server logs.

Backport recommended.

### DIFF
--- a/server/server.cpp
+++ b/server/server.cpp
@@ -468,15 +468,16 @@ void server::accept_connections()
       {
         // Use TolerantConversion so one connections from the same address on
         // IPv4 and IPv6 are rejected as well.
-        if (socket->peerAddress().isEqual(
+        if (!socket->peerAddress().isEqual(
                 pconn->sock->peerAddress(),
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Looks good to me. Backport approved.

PR #1768, opened 2023-01-22
Docs: Update client manual battle log, middle-click and unit controls

### LINKED ISSUE #1709: Add "center unit" button to units control bar
The units control bar establishes a mouse-driven alternative to the keyboard-driven  handling of units.
If the player choses to uncheck "auto center unit" in the interface options the unit controls bar is not sufficient for this because the unit has to be centered manually pressing "c".
While centering the unit is no unit action it nevertheless is in this case indispensable.
Combining the use of the control bar with the keyboard is error prone: If the focus is not on the keyboard misclicks happen more frequently.
The question is if a "center unit" button could be added to the bar and if this could and should be done conditionally, e.g. only if the player does not use the "auto center unit" feature.

jwrober:
Closes #1709

### DIFF
--- a/docs/Manuals/client-manual.rst
+++ b/docs/Manuals/client-manual.rst
@@ -545,9 +545,9 @@ Show New Turn Information
 
 Show Detailed Combat Information
     Enables or Disables populating the :guilabel:`Battle Log` widget. When enabled you will see a widget
-    appear on the screen (typically in the upper left corner) after a combat occurs in your nation's vision.
+    appear on the screen (typically in the upper left corner) after combat occurs in your nation's vision.
     Your nation's vision is all map tiles that are visible to your nation, either natively or via shared
-    vision treaty with an ally or team mate. The figure below gives an example of 3 events.
+    vision treaty with an ally or team mate. The figure below gives an example of 3 combat events.
 
     .. _Battle Log:
     .. figure:: /_static/images/gui-elements/battle-log.png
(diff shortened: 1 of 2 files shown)


docs/Manuals/client-manual.rst
```diff
@@ -1136,9 +1138,12 @@ To move around the map canvas, you can right-click in the main map area and the
 from the center of the screen, the faster the map canvas will move per mouse click. You can also use
 two-finger gestures on your mouse/trackpad to swipe up, down, left, and right.
 
-One other feature of the :guilabel:`Map View` is the ability to middle-click on a unit and a pop-up widget
-will appear giving you some information about the unit. :numref:`Unit Information`, gives an example of a
-:unit:`Howitzer`.
+One other feature of the :guilabel:`Map View` is the ability to middle-click on a unit (via the
+:ref:`Popup Tile Info Shortcut <shortcut-popup-tile-info>`. After a middle-click a popup widget will appear
```
daavko, line 1142:
```suggestion
One other feature of the :guilabel:`Map View` is the ability to middle-click on a unit (via the
:ref:`Popup Tile Info Shortcut <shortcut-popup-tile-info>`). After a middle-click a popup widget will appear
```

docs/Manuals/client-manual.rst
```diff
@@ -1547,17 +1556,22 @@ has been selected.
...
 
-You can see that the :unit:`Workers` is selected because it has a white selection ring around its base. Looking
-at the dialog, in the header, you can see that this unit is ID # 111, has 4 3/9 Move Points (MPs), and 10 of
-10 Hit Points (HPs). From left to right you can see an image of the unit with MPs overlaid, the terrain it is
-on with infrastructure improvements shown, and then lastly the actions that this unit can take. In this
-example the actions available are: Plant to Forest/River, Build Road, Go to Tile, Sentry, Auto Worker, Wait,
-and Done.
+You can see that the :unit:`Workers` is selected because it has a white selection ring around its base.
+Looking at the dialog, in the header, you can see that this unit is ID # 111, has 4 3/9 Move Points (MPs), and
+10 of 10 Hit Points (HPs).
```
daavko, line 1561: Do we need to specify that Move Points are MPs and Hit Points are HPs?

jwrober: It's typically good form to document want an acronym is, especially when shown on an image. Don't want to assume the reader knows what they mean.

PR #1771, opened 2023-01-22
Resize the main start page at leave game

### LINKED ISSUE #1684: Resize game when leaving
The default setting is full screen mode now. When you "leave a game" the UI goes to the start screen still in full screen mode. It looms large and is obnoxious. Same is true if you are in windowed mode (non-full screen), but the window is maximized.

The interface should be pulled out of full screen mode like it is on startup and the resized back to its start size.

None

**Additional context**
None

jwrober:
Closes #1684

### DIFF
--- a/client/page_main.ui
+++ b/client/page_main.ui
@@ -10,6 +10,12 @@
     <height>397</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>821</width>
+    <height>397</height>
+   </size>
+  </property>
   <property name="windowTitle">
    <string>Form</string>
   </property>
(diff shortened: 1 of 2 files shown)


lmoureaux (changes requested):
Not working when the server quits while you're connected.

client/page_main.ui
```diff
@@ -10,6 +10,12 @@
     <height>397</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>821</width>
+    <height>397</height>
```
lmoureaux, line 16: These numbers look quite arbitrary, is there a particular reason for this choice?

jwrober: I just took the same values as the base form geometry

lmoureaux (approved):

PR #1779, opened 2023-01-25
Update message options to use a generic code

### LINKED ISSUE #1740: Events E_CITY_AQUEDUCT and E_CITY_AQ_BUILDING don't apply to just Aqueducts
The two events listed `E_CITY_AQUEDUCT` and `E_CITY_AQ_BUILDING` talk about just Aqueducts, however in the code we support any kind of building that is needed to allow growth and don't limit it to just one building. This is confusing.

Rename the event and the associated title that displays in the message options dialog box to make it more clear what the message does. A suggested title would be "City: Needs Improvement to Grow"

**Additional context**
Came up as part of review of #1726

jwrober:
Instead of hard coded ``E_CITY_AQUEDUCT``

Closes #1740


lmoureaux (changes requested):
Needs backward compatibility code in `options.cpp:message_options_load`, around line 3562.

jwrober:
Do I add the old values to the `enum event_type old_events[]`?

lmoureaux:
> Do I add the old values to the `enum event_type old_events[]`?

I don't think so, this seems to be migration code for 2.1 and earlier.

lmoureaux (approved):

PR #1789, opened 2023-01-29
Client reorg #2

jwrober:
See #1097 

In this PR I move all of the files that generate the "views" from the top bar to reside in their own client sub-directory. I renamed all of the files as well to better represent what they files do along with new comments in each.  This is a precursor for me going after #1498


jwrober:
If I rebase off of `master` will that revert my move of the file here?

lmoureaux:
> If I rebase off of `master` will that revert my move of the file here?

It will try to replay the moves and stop if it can't (then you can solve by hand and tell it to continue).

jwrober:
Ok that was much easier than I expected it to. Simple rebase worked fine.

client/views/view_nations_data.cpp
```diff
@@ -11,6 +11,11 @@
     \_____/ /                     If not, see https://www.gnu.org/licenses/.
       \____/        ********************************************************/
 
+/*
+ * This file contains functions used to gather varying data elements for use
+ * in the nations view (formally known as the plrdlg - player dialog).
```
lmoureaux, line 16: You can use Doxygen markers around file-level comments like this one, so the description appears in the generated docs (not that I've seen anyone use it though). [Relevant documentation.](https://www.doxygen.nl/manual/commands.html#cmdfile)

lmoureaux (approved):
When you write `formally known as the foobar` in the comments, I wonder why you don't use this for the file name. For `plrdlg` I understand, but what about the `city report`?

jwrober:
> When you write `formally known as the foobar` in the comments, I wonder why you don't use this for the file name. For `plrdlg` I understand, but what about the `city report`?

I named all the files `view_[view name]` to make them consistent. And used view because that is what we have been calling them in the docs to make the game less about an IDE feeling (e.g. a report...) and more about a view into an aspect of your game. Someone new to our code, but used to the legacy code will get a quick update with the comment. That is why I did it that way.

PR #1791, opened 2023-01-30
Remove the HAVE_CONFIG_H macro

lmoureaux:
The macro was always defined in the build system, so all #ifdef HAVE_CONFIG_H were always taken (except the one in cvercmp).  Dropping it simplifies the code.


jwrober:
Update copyright dates? Also clang format is not happy it seems.

lmoureaux:
> Update copyright dates?

I don't think this change is copyrightable. Should update all files once and for good.

Code formatting fixed.

jwrober (approved):

Issue #1742, opened 2023-01-15
Update how we do versions


jwrober:
**Is your feature request related to a problem? Please describe.**
Right now we have a somewhat complex version string. I think we can do better to make it easier to read.

**Describe the solution you'd like**
Change the version to go like this:

`[major version].[minor version].[gh hex hash]-[string].[level]`

e.g. `3.0.729e3fb-rc.2` or `3.0-729e3fb.stable.0`

**Describe alternatives you've considered**

**Additional context**
Came up as part of review of #1726

lmoureaux:
Do we need to keep the git hash in the version string? It makes the versions not monotonous.

jwrober:
I'd be fine with that too.

jwrober:
I'm wondering if this is closed/resolved. We support a pretty good versioning system now.

PR #1792, opened 2023-02-01
Remove unused IS_BETA_VERSION and related functions

lmoureaux:
We don't use this code. Cleanup step before #1742.


jwrober:
Needs `clang-format`

jwrober:
Also, are we going to backport 1742, if so do we need this one too?

lmoureaux:
> Also, are we going to backport 1742, if so do we need this one too?

Been wondering the same.

jwrober:
I vote for back-port this and #1742

lmoureaux:
Added #1742 to 3.0-stable then, though it's a bit risky so close to a release.

jwrober:
After further thought, taking this out of back-ports. Too risky so close to release.

PR #1794, opened 2023-02-01
Remove ColorDodge in the city screen and other places

### LINKED ISSUE #1106: Some tileset sprites have a strange transparency effect applied when city view is open
**Describe the bug**
When a city view is open, some sprites (I've noticed this with units, cities, and some tile extras so far) become partially transparent.

**To Reproduce**
1. Observe a unit
2. Open a city view
3. Observe the same unit, now half transparent

**Expected behavior**
They shouldn't be half-transparent

**Screenshots**
Regular view:
freeciv21-client_QzK8MCIOgm
freeciv21-client_OF5l18NEfj
freeciv21-client_3ZJZktjO2N

With city view open:
freeciv21-client_zao8Zj5V1E
freeciv21-client_AeUgIYj86G
freeciv21-client_R618UwXdOw

 - OS: Windows 10
 - Freeciv21 version: beta 2
 - Ruleset/Longturn game (if applicable): not applicable

**Additional context**
It seems like it applies to units regardless of where they are, but for cities and tile extras only if those are outside of the worked area of the currently opened city.

lmoureaux:
The fog effect from the city screen is gone, will do it properly if we want to add it back.

Closes #1106.

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -404,9 +404,8 @@ void progress_bar::paintEvent(QPaintEvent *event)
     }
 
     j = height() - 2 * f_size;
-    p.setCompositionMode(QPainter::CompositionMode_ColorDodge);
-    QFontMetrics fm(*sfont);
 
+    QFontMetrics fm(*sfont);
     if (fm.horizontalAdvance(s1) > rx.width()) {
       s1 = fm.elidedText(s1, Qt::ElideRight, rx.width());
     }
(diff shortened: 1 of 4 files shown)


jwrober (approved):
I think this generally fixes the issue. However I would like to see some kind of shadow effect or something to help highlight the city center on the map. Maybe as part of the city dialog redo as we get rid of the side bars and make then drawers on the left.

lmoureaux:
We have the grid. I wanted to try with only this for a while, to see how it works out. Also, implementing fog didn't seem very easy.

daavko (approved):
Fixes the issue, which is good. Would be nice to at least have an option to draw everything outside the worked area as fogged, but if it's hard to implement then it is what it is. I'll make a feature request after this gets merged so the idea doesn't get completely lost.

jwrober:
Merging. @daavko don't forget the enhancement request :)

PR #1795, opened 2023-02-02
Update WASM Qt to 5.15.8

lmoureaux:
The 5.15.2 branch was removed recently.

Let's see what happens to the CI.


jwrober (approved):

lmoureaux:
Backporting since it happens in `stable` as well

PR #1796, opened 2023-02-02
Fix hud_message_box sometimes showing at the top left of the screen

### LINKED ISSUE #1422: The quit dialog is decoupled from the client window
**Describe the bug**
The quit dialog can get disconnected from the client window

**To Reproduce**
I don't know how to reproduce it at this time. I am on dual-screen so that might have something to do with it. The screenshot taken actually started with the client on one screen and the dialog on the other. I moved the client window over to the same screen to get the picture.

**Expected behavior**

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.0-alpha4+

**Additional context**
None

lmoureaux:
Turned out that calling show() twice moves the widget around. Don't call it from set_text_title() and always call it externally instead.

Closes #1422.

Backporting not envisioned because if I missed a `show()`, that dialog wouldn't show up at all.

### DIFF
--- a/client/dialogs.cpp
+++ b/client/dialogs.cpp
@@ -3394,6 +3394,7 @@ disband_box::disband_box(const std::vector<unit *> &punits, QWidget *parent)
   setDefaultButton(pb);
   connect(pb, &QAbstractButton::clicked, this,
           &disband_box::disband_clicked);
+  pb->show();
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jwrober (approved):
Agree for non back-port as well. This bug was annoying, but not that big a deal.

Issue #1591, opened 2022-12-11
City screen iteration 2


lmoureaux:
**Is your feature request related to a problem? Please describe.**

The city screen in 3.0 is the first iteration since it is on map. While generally functional and working relatively well, some issues have appeared:

* #611 
* #653 
* #728 
* #1407 
* #1490 
* #1511 
* #1518 
* #1584 
* #1588 
* #1589 

This abundance of issues hints that changes to the general layout of the screen are needed.

**Describe the solution you'd like**

We should design and implement a second iteration of the city screen to address the issues listed in the introduction. The first iteration of the city screen was done by changing the layout of the existing widgets until they all fit (and IMHO achieved this quite well). For the second iteration, I think the widgets themselves should be modified to increase information density.

**Describe alternatives you've considered**

Gradually improving the current layout. In view of the issues, I think more radical changes are needed.

psampathkumar:
![image](https://user-images.githubusercontent.com/30164619/207038610-7aa39f0f-15fc-4919-86e7-04700eed29ca.png)
This was one mockup we had. The idea was we have everything in the same place somewhere on the side (for both unit and cities) like civ6 does it.

jwrober:
I would add #944 to this one.

lmoureaux:
Had a discussion on Discord about this, going around an experiment currently found in `wip/city-dialog-phase-2` in my remote. The city dialog in this branch looks like this:

![image](https://user-images.githubusercontent.com/22327575/208177667-06158731-cdca-4148-ad43-34438f4ab1c9.png)

It scales reasonably well down to 768px high, but doesn't look especially good and has more tabs.

The main lessons I draw from this are:
* No big refactor for 3.0
* There is indeed a need for more compact widgets, especially the governor settings and general info
* Pushing things to the sides works, but only two sides: left/below (maybe 2 for small screens? but that's Plan B)
* Consider tabbing present/supported/buildings
* Widgets should expand to use more space when available (and not just leave gaps between them)
* The map should be zoomed in/out to adjust for the city visible area
* Widgets with variable width (happiness) are annoying

These aren't new but it's good to have them written down somewhere.

Mu-00:
[...]

PR #1801, opened 2023-02-03
City dialog: push to the sides

### LINKED ISSUE #728: Change the City Dialog layout so it looks more centered
As stated in the issue title, change the way the city dialog is shown on the screen. Right now there is an artificial spacer in place between the city worked tiles and the two side panels and when shown on the screen makes it look like the whole thing is off center.

Instead of an artificial spacer, determine the size of the city center is in pixels and then tuck the two side panels to the left and right side. Then as zoom in/out happens, the side panels are always right next to the city center. The city center is already centered on the screen, so this is more of a movement of the side panels, mainly the left one.

Doing nothing is also an option. This request is more about making the client look polished and professional.

**Additional context**
None, but here are a couple screen shots showing the space on the left side due to the artificial spacer.

image

image

### LINKED ISSUE #1714: Suggestions for new city dialog left "side" bar
Master branch is working to develop a new city dialog. Some comments on the side bar as it stands as of https://github.com/longturn/freeciv21/commit/b688f424a69b55abcafdeeab6997e1464df3a10a.

1. IMO, the dialog should be closer to the Map View button. Even better if it "unfolds" from the top menu in that area when invoked

image

2. As shown in the image above, if there are no supported improvements or units, nothing is shown but the scroll bar. Suggest that the bar is always the same size like this:

image

and when "blank" has text that says something like "no supported improvements" or "no supported units". Might also be good if we can add a separator in there to add a visual differentiation between improvements, wonders and units.

3. Lastly, I think the city icon and the supported items bar should be coupled together with no separation between them.

lmoureaux:
This is one step further towards #1591. It pushes the panels to the side and addresses #1714.

Closes #728.
Closes #1714.


jwrober (changes requested):
This is a good start. I have some suggestions.

jwrober:
Suggestions:

- Consolidate all the tabs into one bank on the left. General / Production / Citizens / Governor.
- Move the city Next/Previous/Close button to above the city name
- Have a look at the sprites that go along with the varying values in the screen shot below. They don't update properly if you switch between themes.

image

lmoureaux:
> * Consolidate all the tabs into one bank on the left. General / Production / Citizens / Governor.

This would make it harder to develop the new layout, so I'm not doing this.

> * Move the city Next/Previous/Close button to above the city name

Done along with a second pass on the rest of the layout.

> * Have a look at the sprites that go along with the varying values in the screen shot below. They don't update properly if you switch between themes.

Opened #1803 about this.

jwrober (approved):
Second commit is much better.

PR #1802, opened 2023-02-03
New Units View

jwrober:
Implement a new page view for units. See #1498

ToDo:

- [x] Gather the proper values for `upgradable` and `in progress`

To get to the view as it stands in this draft, you right-click on the top bar. Left click opens the old widget.

Sample:
image

### DIFF
--- a/client/repodlgs_common.cpp
+++ b/client/repodlgs_common.cpp
@@ -156,23 +156,21 @@ void get_economy_report_units_data(struct unit_entry *entries,
 std::vector<unit_view_entry>
 get_units_view_data(struct unit_view_entry *entries, int *num_entries_used)
 {
-  int count, in_progress, gold_cost, food_cost, shield_cost;
-  bool upgradable = false;
-
   *num_entries_used = 0;
 
   if (nullptr == client.conn.playing) {
     return {};
   }
 
-  count = 0;       // Count of active unit type
-  in_progress = 0; // Count of being produdced
-  gold_cost = 0;   // Gold upkeep
-  food_cost = 0;   // Food upkeep
-  shield_cost = 0; // Shield upkeep
-
   unit_type_iterate(unittype)
   {
+    int count = 0;           // Count of active unit type
+    int in_progress = 0;     // Count of being produced
+    int gold_cost = 0;       // Gold upkeep
+    int food_cost = 0;       // Food upkeep
+    int shield_cost = 0;     // Shield upkeep
+    bool upgradable = false; // Unit type is upgradable
+
     unit_list_iterate(client.conn.playing->units, punit)
     {
       if (unit_type_get(punit) == unittype) {
@@ -187,13 +185,19 @@ get_units_view_data(struct unit_view_entry *entries, int *num_entries_used)
     }
     unit_list_iterate_end;
 
-    city_list_iterate(client.conn.playing->units, pcity)
+    city_list_iterate(client.conn.playing->cities, pcity)
     {
-      if (VUT_UTYPE == pcity->production.kind) {
+      if (pcity->production.value.utype == unittype
+          && pcity->production.kind == VUT_UTYPE) {
         in_progress++;
       }
     }
     city_list_iterate_end;
+
+    // Skip unused unit types
+    if (count == 0 && in_progress == 0) {
+      continue;
+    }
 
     entries[*num_entries_used].type = unittype;
     entries[*num_entries_used].count = count;
--- a/client/repodlgs_common.cpp
+++ b/client/repodlgs_common.cpp
@@ -151,6 +151,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
 }
 
 /**
+   Returns an array of units data. Number of units in
+   the array is added to num_entries_used.
+ */
+void get_units_view_data(struct unit_view_entry *entries,
+                         int *num_entries_used)
+{
+  int count, gold_cost, food_cost, shield_cost;
+
+  *num_entries_used = 0;
+
+  if (nullptr == client.conn.playing) {
+    return;
+  }
+
+  unit_type_iterate(unittype)
+  {
+    count = 0;
+    gold_cost = 0;
+    food_cost = 0;
+    shield_cost = 0;
+
+    city_list_iterate(client.conn.playing->cities, pcity)
+    {
+      unit_list_iterate(pcity->units_supported, punit)
+      {
+        if (unit_type_get(punit) == unittype) {
+          count++;
+          gold_cost += punit->upkeep[O_GOLD];
+          food_cost += punit->upkeep[O_FOOD];
+          shield_cost += punit->upkeep[O_SHIELD];
+        }
+      }
+      unit_list_iterate_end;
+    }
+    city_list_iterate_end;
+
+    if (count == 0) {
+      continue;
+    }
+
+    entries[*num_entries_used].type = unittype;
+    entries[*num_entries_used].count = count;
+    entries[*num_entries_used].gold_cost = gold_cost;
+    entries[*num_entries_used].food_cost = food_cost;
+    entries[*num_entries_used].shield_cost = shield_cost;
+    (*num_entries_used)++;
+  }
+  unit_type_iterate_end;
+
+  std::sort(entries, entries + *num_entries_used,
+            [](const auto &lhs, const auto &rhs) {
+              return QString(utype_name_translation(lhs.type))
+                     < QString(utype_name_translation(rhs.type));
+            });
+}
+
+/**
    Sell all improvements of the given type in all cities.  If
  "redundant_only" is specified then only those improvements that are replaced
  will be sold.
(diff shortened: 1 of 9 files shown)


lmoureaux:
About the todo list:

* **Get unit wait time table populated:** Could we show only the first/last unit to end UWT? Then it could fit in the table format. This can also be left out of the first iteration
* **Get upkeep icons in the header (from the old widget):** Can be done later
* **Gather the proper values for upgradable and in progress:** Not sure what this means but sounds pretty much needed
* **Add support for global observe (see all units everywhere like in cities view):** Can be done later
* **Get the slots to work:** Or buttons in follow-up PRs?
* **Tear out the old widget code & following:** Great! If some functionality isn't there in the first iteration, maybe this can wait

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
   unit_type_iterate_end;
 }
 
+/**
+   Returns an array of units data. Number of units in
+   the array is added to num_entries_used.
```
lmoureaux, line 155: New code should have the *'s
```suggestion
 * Returns an array of units data. Number of units in
 * the array is added to num_entries_used.
```

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
...
+ */
+void get_units_view_data(struct unit_view_entry *entries,
+                         int *num_entries_used)
+{
+  int count, gold_cost, food_cost, shield_cost;
+
+  *num_entries_used = 0;
```
lmoureaux, line 162: This doesn't follow the documentation ("Number of units in the array is added to num_entries_used.")

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
   unit_type_iterate_end;
 }
 
+/**
+   Returns an array of units data. Number of units in
+   the array is added to num_entries_used.
+ */
+void get_units_view_data(struct unit_view_entry *entries,
+                         int *num_entries_used)
```
lmoureaux, line 158: Would prefer a `std::vector<unit_view_entry> get_units_view_data()`, but it seems you're copying some other code that uses C-style pointers.

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
...
+{
+  int count, gold_cost, food_cost, shield_cost;
```
lmoureaux, line 160: Best declare variables on first use (reduces the number of variables one needs to track when reading the code).

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
...
+  if (nullptr == client.conn.playing) {
+    return;
+  }
+
+  unit_type_iterate(unittype)
+  {
+    count = 0;
+    gold_cost = 0;
+    food_cost = 0;
+    shield_cost = 0;
```
lmoureaux, line 173: This is most likely the cause of the wrong math. Should be outside the loop.

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
...
+
+    city_list_iterate(client.conn.playing->cities, pcity)
+    {
+      unit_list_iterate(pcity->units_supported, punit)
```
lmoureaux, line 177: I think we have `client.conn.playing->units`

client/repodlgs_common.cpp
```diff
@@ -150,6 +150,63 @@ void get_economy_report_units_data(struct unit_entry *entries,
...
+    entries[*num_entries_used].food_cost = food_cost;
+    entries[*num_entries_used].shield_cost = shield_cost;
+    (*num_entries_used)++;
+  }
+  unit_type_iterate_end;
+
+  std::sort(entries, entries + *num_entries_used,
+            [](const auto &lhs, const auto &rhs) {
+              return QString(utype_name_translation(lhs.type))
+                     < QString(utype_name_translation(rhs.type));
```
lmoureaux, line 206: Actually we need to sort this in a locale-aware way, otherwise `Éléphants` will be at the end, while it should be between `Dragons` and `Fret`. The code would be:
```cpp
return QString::localeAwareCompare(utype_name_translation(lhs.type), utype_name_translation(rhs.type)) > 0;
```

client/repodlgs_common.h
```diff
@@ -23,6 +23,11 @@ struct unit_entry {
   int count, cost, total_cost;
 };
 
+struct unit_view_entry {
```
lmoureaux, line 26: Would be useful to document what this is.

client/views/view_units.cpp
```diff
@@ -0,0 +1,466 @@
...
+{
+  ui.setupUi(this);
+
+  QStringList slist;
+  slist << _("Unit Type") << _("Upgradable ★") << _("In Progress ⚒")
+        << _("Active ⚔") << _("Shield Upkeep") << _("Food Upkeep")
+        << _("Gold Upkeep") << QLatin1String("");
+  ui.units_widget->setColumnCount(slist.count());
+  ui.units_widget->setHorizontalHeaderLabels(slist);
+  ui.units_widget->setSortingEnabled(0);
```
lmoureaux, line 90:
```suggestion
  ui.units_widget->setSortingEnabled(false);
```

client/views/view_units.cpp
```diff
@@ -0,0 +1,466 @@
...
+  ui.upg_but->setText(_("Upgrade"));
+  ui.upg_but->setDisabled("True");
```
lmoureaux, line 92:
```suggestion
  ui.upg_but->setDisabled(true);
```

client/views/view_units.cpp
```diff
@@ -0,0 +1,466 @@
...
+  ui.find_but->setText(_("Find Nearest"));
+  ui.find_but->setDisabled("True");
```
lmoureaux, line 94:
```suggestion
  ui.find_but->setDisabled(true);
```

client/views/view_units.cpp
```diff
@@ -0,0 +1,466 @@
...
+  ui.disband_but->setText(_("Disband All"));
+  ui.disband_but->setDisabled("True");
```
lmoureaux, line 96:
```suggestion
  ui.disband_but->setDisabled(true);
```

client/views/view_units.cpp
```diff
@@ -0,0 +1,466 @@
...
+    struct unit_type *putype = pentry->type;
+    cid id;
+
+    auto sprite = get_unittype_sprite(tileset, putype, direction8_invalid());
+    id = cid_encode_unit(putype);
+
+    ui.units_widget->insertRow(max_row);
+    for (j = 0; j < 7; j++) {
+      item = new QTableWidgetItem;
+      switch (j) {
```
lmoureaux, line 158: Using a `switch` instead of a loop sounds a bit complicated... The code shared between the iterations is:
```cpp
      item = new QTableWidgetItem;
      ...
      item->setTextAlignment(Qt::AlignVCenter | Qt::AlignHCenter);
      ui.units_widget->setItem(max_row, j, item);
```
I'm not sure it's worth the complexity of a loop and switch...

client/views/view_units.cpp
```diff
@@ -0,0 +1,466 @@
...
+      item_totals->setTextAlignment(Qt::AlignCenter);
+      item_totals->setText(_("-------\n0"));
+      // item->setData(Qt::DisplayRole, 0);
+      break;
+    case 2:
+      // In Progress
+      item_totals->setTextAlignment(Qt::AlignCenter);
+      item_totals->setText(_("-------\n0"));
+      // item->setData(Qt::DisplayRole, 0);
+      break;
```
lmoureaux, line 240: Here a loop would make sense since the exact same code is repeated many times. But would require putting the totals in an array...

client/views/view_units.ui
```diff
@@ -0,0 +1,156 @@
...
+       <enum>QAbstractItemView::SingleSelection</enum>
+      </property>
+      <property name="selectionBehavior">
+       <enum>QAbstractItemView::SelectRows</enum>
+      </property>
+      <property name="showGrid">
+       <bool>false</bool>
+      </property>
+      <property name="sortingEnabled">
+       <bool>true</bool>
```
lmoureaux, line 141: This contradicts what you set in the code, I'd keep a single source of truth.

lmoureaux (commented):
Stylistic and code-wise comments as requested. I may have found the bug that makes the totals wrong.

client/repodlgs_common.cpp
jwrober: I did copy another older function. Happy to change the return type.

client/views/view_units.cpp
jwrober: I used code from the economics view that does this loop/switch deal.

jwrober:
Take 2 adds support for # in progress and upgradable. I may have introduced a regression in the `get_units_view_data()`

jwrober:
With take 3, we now have a fully populated table of all the same data as the old widget. Thanks to @lmoureaux!

PR #1807, opened 2023-02-06
Units View #2

jwrober:
See #1498

- [x] Get unit wait time table populated
- [x] Get upkeep icons in the header (from the old widget)
- [x] Gather the proper values for `upgradable` and `in progress`

### DIFF
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -31,14 +31,8 @@ units_view::units_view() : QWidget()
 {
   ui.setupUi(this);
 
-  const QPixmap *spr;
-  QImage img;
-  QRect crop;
-  QImage cropped_img;
-  QPixmap pix;
-  QStringList slist;
-
   // Configure the units table
+  QStringList slist;
   slist << _("Unit Type") << _("★ Upgradable") << _("⚒ In Progress")
         << _("⚔ Active") << _("Shield Upkeep") << _("Food Upkeep")
         << _("Gold Upkeep") << QLatin1String("");
@@ -63,16 +57,17 @@ units_view::units_view() : QWidget()
   ui.uwt_label->setText("Units Waiting:");
 
   // Add shield icon for shield upkeep column
-  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "upkeep.shield",
-                                    "citybar.shields", "", "", false);
-  img = spr->toImage();
-  crop = zealous_crop_rect(img);
-  cropped_img = img.copy(crop);
-  pix = QPixmap::fromImage(cropped_img);
+  const QPixmap *spr =
+      tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "upkeep.shield",
+                                  "citybar.shields", "", "", false);
+  QImage img = spr->toImage();
+  QRect crop = zealous_crop_rect(img);
+  QImage cropped_img = img.copy(crop);
+  QPixmap pix = QPixmap::fromImage(cropped_img);
   ui.units_widget->horizontalHeaderItem(4)->setIcon(pix);
 
   // Add food icon for food upkeep column
-  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "citybar.food",
+  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "upkeep.food",
                                     "citybar.food", "", "", false);
   img = spr->toImage();
   crop = zealous_crop_rect(img);
@@ -81,9 +76,7 @@ units_view::units_view() : QWidget()
   ui.units_widget->horizontalHeaderItem(5)->setIcon(pix);
 
   // Add gold icon for gold upkeep column
-  // FIXME: We don't have a sprite for a gold coin on its own. This looks
-  // better than using get_tax_sprite(tileset, O_GOLD)
-  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "citybar.trade",
+  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "upkeep.gold",
                                     "citybar.trade", "", "", false);
   img = spr->toImage();
   crop = zealous_crop_rect(img);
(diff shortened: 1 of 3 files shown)


client/views/view_units.cpp
```diff
@@ -17,26 +18,78 @@
...
  */
 units_view::units_view() : QWidget()
 {
   ui.setupUi(this);
 
+  const QPixmap *spr;
+  QImage img;
+  QRect crop;
+  QImage cropped_img;
+  QPixmap pix;
```
lmoureaux, line 38: Please don't get into this habit of declaring everything at the top. Just declare variables on first use, possibly using `const` and `auto`, it reduces the risk of using uninitialized variables.

client/views/view_units.cpp
```diff
@@ -17,26 +18,78 @@
...
+  ui.uwt_widget->setSortingEnabled(false);
+  ui.uwt_label->setText("Units Waiting:");
+
+  // Add shield icon for shield upkeep column
+  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "upkeep.shield",
+                                    "citybar.shields", "", "", false);
+  img = spr->toImage();
+  crop = zealous_crop_rect(img);
+  cropped_img = img.copy(crop);
+  pix = QPixmap::fromImage(cropped_img);
```
lmoureaux, line 71: Ooof we need a shield icon that's not the upkeep one

client/views/view_units.cpp
```diff
@@ -17,26 +18,78 @@
...
+  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "citybar.food",
+                                    "citybar.food", "", "", false);
+  img = spr->toImage();
+  crop = zealous_crop_rect(img);
+  cropped_img = img.copy(crop);
+  pix = QPixmap::fromImage(cropped_img);
+  ui.units_widget->horizontalHeaderItem(5)->setIcon(pix);
+
+  // Add gold icon for gold upkeep column
+  // FIXME: We don't have a sprite for a gold coin on its own. This looks
```
lmoureaux, line 84: `upkeep.gold`?

client/views/view_units.cpp
```diff
@@ -17,26 +18,78 @@
...
+  ui.uwt_widget->setHorizontalHeaderLabels(slist);
+  ui.uwt_widget->setSortingEnabled(false);
+  ui.uwt_label->setText("Units Waiting:");
+
+  // Add shield icon for shield upkeep column
+  spr = tiles_lookup_sprite_tag_alt(tileset, LOG_VERBOSE, "upkeep.shield",
+                                    "citybar.shields", "", "", false);
+  img = spr->toImage();
+  crop = zealous_crop_rect(img);
+  cropped_img = img.copy(crop);
```
lmoureaux, line 70: May want to write an overload of `zealous_crop_rect` taking a `QImage` instead of repeating this code. I hate `zealous` by the way, because it results in inconsistent sprite sizing all over.

client/views/view_units.cpp
```diff
@@ -206,9 +259,68 @@ void units_view::update_view()
...
+      struct city *pcity_near = get_nearest_city(punit, &pcity_near_dist);
+      ui.uwt_widget->setItem(
+          units_count, 1,
+          new QTableWidgetItem(
+              get_nearest_city_text(pcity_near, pcity_near_dist),
+              pcity_near_dist));
+
+      ui.uwt_widget->setItem(
+          units_count, 2,
+          new QTableWidgetItem(move_points_text(punit->moves_left, false),
```
lmoureaux, line 299: Aren't UWT units supposed to have full MP anyway?

client/views/view_units.cpp
```diff
@@ -217,44 +329,37 @@ void units_view::update_view()
...
-      break;
-    case VUT_UTYPE:
-      counter = ui.eco_widget->item(curr_row, 3)->text().toInt();
-      ui.bdisband->setEnabled(true);
-      break;
-    default:
-      qCritical("Not supported type: %d.", selected.kind);
-    }*/
+    if (counter > 0) {
+      ui.disband_but->setDisabled(false);
```
lmoureaux, line 355: Why can't we disband multiple unit types at once? If you want the user to only ever act on a single unit type, you can change the selection mode of the table to `SingleSelection`.

lmoureaux (commented):
Code comments only, didn't run it. Will do a feature review with #1808.

client/views/view_units.cpp
jwrober: Let's talk about this one in #1808

client/views/view_units.cpp
jwrober: `SingleSelection` is on by default. This is checking to see if there any of the unit active. I didn't code to do more than one mass disband at once. This is probably a new feature issue post merge.

client/views/view_units.cpp
jwrober: I copied this verbatim from legacy mir3x code. Maybe write a small helper function to do the work instead? If you are not very happy with `zealous` then I don't suggest we use it here, but can at least reduce the dups.

client/views/view_units.cpp
jwrober: I don't think there is a sprite for that, hence the ask of hugomflavio for one.

client/views/view_units.cpp
jwrober: I see that as an artwork request for an update post initial merge.

client/views/view_units.cpp
lmoureaux: It's kind of needed here because the code uses sprites not meant to be used separately (the upkeep ones). I suggest adding the following next to `zealous`:
```cpp
QPixmap zealous_crop(const QPixmap &pix)
{
  return pix.copy(zealous_crop_rect(pix.toImage()));
}
```

client/views/view_units.cpp
lmoureaux: `upkeep.gold`, for amplio2 in upkeep.png:
image

jwrober: But that is only for amplio2, correct? Was looking for something tileset agnostic. The three in use are in `misc`.

lmoureaux: `upkeep.shield` is not, though the alt `citybar.shields` is (but since `upkeep.shields` is mandatory, the alt will never be used). In any case they can be overridden by the tileset by not importing `misc`.

Issue #944, opened 2022-03-17
Integrate cities tab with the city dialog


pranavtifr:
**Is your feature request related to a problem? Please describe.**
In GTK Client, because of the fact that double clicking the city name in cities tab opens the city dialog separate window, it is really convenient to double click a city name  use the city dialog and move on to the next city in cities tab. In FC21, we switch to view mode when double clicking the city. You then have to go back to cities tab to continue working on other cities. 

**Describe the solution you'd like**
FC21 is more view mode centric than GTK, and we need to preserve that. One idea, which was discussed on discord, is to open a list of city names when double clicking from cities tab (This list has the same sorting as the cities tab) The list is minimal, and shows only city names and the currently sorted column. Clicking on the name on this list switches to the city and opens that cities dialog. You can also select neighboring cities by clicking on the cities in view mode. Pressing escape, will take you back to the cities tab. 

The city list will open on the right like this, 
![image](https://user-images.githubusercontent.com/30164619/158839882-19460682-0ca7-4819-9208-91a8e03dad64.png)

**Describe alternatives you've considered**
The positioning of the city list, is upto debate, We can also open it on the left like this. 
![image](https://user-images.githubusercontent.com/30164619/158840378-2fac9bf2-796a-4178-9f65-4b49a7db481b.png)

or position it in the bottom with a scrollbar. 
![image](https://user-images.githubusercontent.com/30164619/158841055-6cb68a5b-a1ee-4b2f-91b2-86ca815ba3e6.png)
This way, we can show more columns too if needed.

lmoureaux:
If we think that there would be an animated transition, what would be the most natural option? I think that in this case the city list should be full width or full height. To me it feels more natural to have a navigational element (here the city list) appear on the left, like bookmarks in Firefox.

PR #1808, opened 2023-02-07
Units View #3

### LINKED ISSUE #1498: Change units view from a widget to a page/table

jwrober:
On top of #1807

Absent some housekeeping, effectively closes #1498 

- [x] Get the slots to work
- [x] Tear out the old widget code (or rearrange so old widget is a right click)
- [x] Change the action button to a regular left click on the top bar
- [x] Make sure F2 opens the new view
- [x] Update the client manual


lmoureaux (changes requested):
*Functional review only, didn't look at the code yet.*

Long term, I would suggest to merge the two widgets in a tree view, with the first level being unit types and individual units visible at the second level. This could be integrated with the map view in the same way as the list of cities (#944).

Both tables appear to have an empty column to the right. Clicking on the header makes it expand to the full width of the widget.

**Bug:** Clicking on the table headers changes the font.

I would prefer if the first column could be left-aligned.

**Top Panel**

**Bug:** Global observers don't see anything. Either the view should be disabled, or it should be useful.

I got the impression that the `Find Nearest` button switches view after the map sliding animation has started. The button is also confusing: "nearest" to what? The currently invisible tile around which the map is centered? Maybe it would be more useful to show the Advanced Unit Selection dialog with the unit type preselected.

The `Upgrade` button could be more informative about how many units will be upgraded and the total cost.

**Bottom Panel**

**Bug:** The displayed times are not updated. I see `QObject::startTimer: Timers cannot have negative intervals` in the console, maybe this is related.

**Bug:** Double clicking a unit should focus it on the map.

Is MP relevant here? Shouldn't it always be full for units under UWT?

`Near Location` should be `Location`, because it turns to `in Cityname` when the unit is inside a city.

`Time left` doesn't follow capitalization rules.

The Time Left column could use the same logic as the Turn Done button to avoid showing seconds when hours are left.

jwrober:
> Bug: Clicking on the table headers changes the font.

The same happens on the economics view, which this view takes a lot of lift from. Not going to fix it now. See related issue.

jwrober:
> I got the impression that the Find Nearest button switches view after the map sliding animation has started. The button is also confusing: "nearest" to what? The currently invisible tile around which the map is centered? Maybe it would be more useful to show the Advanced Unit Selection dialog with the unit type pre-selected.

The nearest unit is the one closest to where you are on the map when you enter the units view. This is same functionality to the legacy fc GTK client. The client should close the units view and open the map to show the sliding animation as it centers on the unit and stays on the map when done. I suggest an enhancement request tagged here if you want to change. As it is now, its still a big improvement over the old widget.

jwrober:
> Bug: The displayed times are not updated.

That's not a bug. I am only passing a string to the table. It updates if you go off the view and come back, but does not do an active countdown timer in the view like the old UWT table did. I always felt the countdown timer to be superfluous so didn't take the time to code a QTimer in the new code.

jwrober:
> Bug: Double clicking a unit should focus it on the map.

That's not a bug, that's a feature that didn't exist in the old widget so wasn't coded in the new view.

jwrober:
> Is MP relevant here? Shouldn't it always be full for units under UWT?

MPs should not be shown in the UWT table. I pulled it out.

jwrober:
> I would prefer if the first column could be left-aligned.

It is in `units_view.cpp` line 155 == `item->setTextAlignment(Qt::AlignLeft);`

jwrober:
> Both tables appear to have an empty column to the right. Clicking on the header makes it expand to the full width of the widget.

The Economics view has this too. Recall I used that as the template.

jwrober:
> Long term, I would suggest to merge the two widgets in a tree view, with the first level being unit types and individual units visible at the second level. This could be integrated with the map view in the same way as the list of cities.

As I have been looking over things today, I think that your idea of using a `QTreeView` and `QAbstractListModel` like what the cities view does is the better way to go. I'm going to push an update that adds tool tips to the buttons and fixes a couple other items. My ask/suggestion when you get to code review is to keep that in mind. When we get this guy merged, I'll take a crack at both a refactor of this and the econ view as well. Will enable possible future features like a pop-up menu in units for more functions.

Units View Bugs
- [x] Add tooltips to the buttons to aid the user in what they do
- [x] UWT Near Location should be Location

lmoureaux:
> > Bug: The displayed times are not updated.
> 
> That's not a bug. I am only passing a string to the table. It updates if you go off the view and come back, but does not do an active countdown timer in the view like the old UWT table did. I always felt the countdown timer to be superfluous so didn't take the time to code a QTimer in the new code.

I would argue that it is a bug. If I leave the list open with 20 minutes to go, go do something else and check again, the displayed values will be off by several minutes. I think the problem with the previous widget was that the precision was too high, but if we restrict this to the same as in the Turn Done widget it makes sense to keep updating. It would update every second only for units that are close to come back to life, while others would be updated every minute at worst.

lmoureaux:
> > Bug: Double clicking a unit should focus it on the map.
> 
> That's not a bug, that's a feature that didn't exist in the old widget so wasn't coded in the new view.

Then it was already a bug ^^ but ok to leave it out then

lmoureaux:
> > Long term, I would suggest to merge the two widgets in a tree view, with the first level being unit types and individual units visible at the second level. This could be integrated with the map view in the same way as the list of cities.
> 
> As I have been looking over things today, I think that your idea of using a `QTreeView` and `QAbstractListModel` like what the cities view does is the better way to go.

I wouldn't go as far as a full `QAbstractListModel`, `QTreeWidget` sounds good enough (until we start to have performance issues).

lmoureaux:
> Will enable possible future features like a pop-up menu in units for more functions.

We have [`unit_quick_menu`](https://github.com/longturn/freeciv21/blob/master/client/utils/unit_quick_menu.h) for this :)

jwrober:
> I would argue that it is a bug. If I leave the list open with 20 minutes to go, go do something else and check again, the displayed values will be off by several minutes. I think the problem with the previous widget was that the precision was too high, but if we restrict this to the same as in the Turn Done widget it makes sense to keep updating. It would update every second only for units that are close to come back to life, while others would be updated every minute at worst.

Ok point taken. I'll do another commit to get the timer to update following same algorithm as the turn done button for consistency.  I'll also look at global observe based on feedback from Discord.

lmoureaux:
Would be great to make sure this new Coverity warning is fixed in this PR:
```
________________________________________________________________________________________________________
*** CID 1504578:  High impact quality  (Y2K38_SAFETY)
/client/views/view_units.cpp: 300 in units_view::update_waiting()()
294     
295           time_t dt = time(nullptr) - punit->action_timestamp;
296           if (dt < 0 && !can_unit_move_now(punit)) {
297             char buf[64];
298             format_time_duration(-dt, buf, sizeof(buf));
299             ui.uwt_widget->setItem(units_count, 3,
>>> CID 1504578:  High impact quality  (Y2K38_SAFETY)
>>> A "time_t" value is stored in an integer with too few bits to accommodate it.  The expression "dt" is cast to "int".
300                                    new QTableWidgetItem(buf, dt));
301           }
302     
303           ++units_count;
304         }
305       }
```
This points out a bug, `dt` is being passed as `QTableWidgetItem`'s second argument `type`. Not sure what was intended and didn't check that the code still exists in this PR.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
 // client
 #include "climisc.h"
-#include "repodlgs_g.h"
 
 struct unit_type;
 
-/****************************************************************************
-  Tab widget to display units view (F2)
-****************************************************************************/
+/*
```
lmoureaux, line 20:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Structure of data for the Units View.
+ * See get_units_view_data()
+ */
+struct unit_view_entry {
+  struct unit_type *type;
+  int count, in_prod, total_cost, food_cost, gold_cost, shield_cost;
+  bool upg;
+};
+
+/*
```
lmoureaux, line 30:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Structure of unit waiting data for the Units View.
+ * See get_units_waiting_data()
+ */
+struct unit_waiting_entry {
+  struct unit_type *type;
+  QString timer, city_name;
+};
+
+/*
```
lmoureaux, line 39:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Function to pull data on units for the base units widget
+ */
```
lmoureaux, line 41: Prefer having the documentation headers next to the implementation.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+std::vector<unit_view_entry>
+get_units_view_data(struct unit_view_entry *entries, int *num_entries_used);
+
+/*
```
lmoureaux, line 45:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Function to pull data on units subject to unitwaittime
+ */
```
lmoureaux, line 47: Prefer having the documentation headers next to the implementation.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+std::vector<unit_waiting_entry>
+get_units_waiting_data(struct unit_waiting_entry *entries,
+                       int *num_entries_used);
+
+/*
```
lmoureaux, line 52:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Function used by update_queue
+ */
```
lmoureaux, line 54: Prefer having the documentation headers next to the implementation.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+void units_view_dialog_update(void *unused);
+
+/*
```
lmoureaux, line 57:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Full destructor function
+ */
```
lmoureaux, line 59: Prefer having the documentation headers next to the implementation.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+void popdown_units_view();
+
+/*
```
lmoureaux, line 62:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Helper function to crop a sprite
+ */
```
lmoureaux, line 64: Prefer having the documentation headers next to the implementation.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+QPixmap crop_sprite(const QPixmap *sprite);
```
lmoureaux, line 65: Could this be in `canvas.{h,cpp}` for other code to benefit from it?

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+
+/*
```
lmoureaux, line 67:
```suggestion
/**
```

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+ * Function to help us find the nearest unit
+ */
```
lmoureaux, line 69: Prefer having the documentation headers next to the implementation.

client/views/view_units.h
```diff
@@ -14,13 +14,65 @@
...
+struct unit *find_nearest_unit(const struct unit_type *utype,
+                               struct tile *ptile);
+
+/*
```
lmoureaux, line 73:
```suggestion
/**
```

client/views/view_units.cpp
```diff
@@ -1,28 +1,37 @@
 /*
  * SPDX-License-Identifier: GPL-3.0-or-later
+ * SPDX-FileCopyrightText: Freeciv21 and Freeciv contributors
  * SPDX-FileCopyrightText: James Robertson <jwrober@gmail.com>
  */
 
 /*
```
lmoureaux, line 7:
```suggestion
/**
```

client/views/view_units.cpp
```diff
@@ -1,28 +1,37 @@
...
- * \file This file contains functions to generate the table based GUI for
+ * \class units_view
+ * This file contains functions to generate the table based GUI for
  * the units view (F2).
  */
```
lmoureaux, line 11: You already have this comment in the `.h`, don't duplicate it

client/views/view_units.cpp
```diff
@@ -33,59 +42,56 @@ units_view::units_view() : QWidget()
...
+        << _("⚔  Active") << _("Shield Upkeep") << _("Food Upkeep")
         << _("Gold Upkeep") << QLatin1String("");
   ui.units_label->setText(QString(_("Units:")));
   ui.units_widget->setColumnCount(slist.count());
   ui.units_widget->setHorizontalHeaderLabels(slist);
   ui.units_widget->setSortingEnabled(false);
+  ui.units_widget->setAlternatingRowColors(true);
   ui.upg_but->setText(_("Upgrade"));
+  ui.upg_but->setToolTip(_("Upgrade selected unit."));
+  ui.upg_but->setToolTipDuration(5000);
```
lmoureaux, line 55: The default duration is calculated automatically based on the length of the tooltip, why change it? The formula is `10000 + 40 * qMax(0, text.length()-100)`, would give 10000 in this case.

client/views/view_units.cpp
```diff
@@ -33,59 +42,56 @@ units_view::units_view() : QWidget()
...
   ui.upg_but->setDisabled(true);
   ui.find_but->setText(_("Find Nearest"));
+  ui.find_but->setToolTip(_("Center the map on the nearest unit in relation "
+                            "to where the map is now."));
+  ui.find_but->setToolTipDuration(10000);
```
lmoureaux, line 60: This is the default duration, why set it explicitly?

client/views/view_units.cpp
```diff
@@ -33,59 +42,56 @@ units_view::units_view() : QWidget()
...
   ui.find_but->setDisabled(true);
   ui.disband_but->setText(_("Disband All"));
+  ui.disband_but->setToolTip(_("Disband all of the selected unit."));
+  ui.disband_but->setToolTipDuration(5000);
```
lmoureaux, line 64: This is shorter than the default duration, why set it explicitly?

client/views/view_units.cpp
```diff
@@ -113,27 +119,28 @@ void units_view::init() { queen()->game_tab_widget->setCurrentIndex(index); }
...
+  int total_count = 0;  // Sum of unit type
+  int upg_count = 0;    // Sum of upgradable
+  int in_progress = 0;  // Sum of in progress by unit type
+  int total_gold = 0;   // Sum of gold upkeep
+  int total_shield = 0; // Sum of shield upkeep
+  int total_food = 0;   // Sum of food upkeep
+  int max_row = 0;      // Max rows in the table widget
+  int i, j;
+
   for (i = 0; i < entries_used; i++) {
```
lmoureaux, line 144:
```suggestion
  for (int i = 0; i < entries_used; i++) {
```

client/views/view_units.cpp
```diff
@@ -144,16 +151,16 @@ void units_view::update_view()
 
     ui.units_widget->insertRow(max_row);
     for (j = 0; j < 7; j++) {
```
lmoureaux, line 153:
```suggestion
    for (int j = 0; j < 7; j++) {
```

client/views/view_units.cpp
```diff
@@ -144,16 +151,16 @@ void units_view::update_view()
...
       switch (j) {
       case 0:
         // Unit type image sprite and name
         if (sprite != nullptr) {
           item->setData(Qt::DecorationRole, sprite->scaledToHeight(h));
         }
         item->setData(Qt::UserRole, id);
-        item->setTextAlignment(Qt::AlignLeft);
         item->setText(utype_name_translation(putype));
+        item->setTextAlignment(Qt::AlignLeft | Qt::TextAlignmentRole);
```
lmoureaux, line 163: Surprised by the `TextAlignmentRole`, was it that or swapping `setText` and `setTextAlignment` that made text left aligned?

client/views/view_units.cpp
```diff
@@ -262,57 +282,66 @@ void units_view::update_view()
  */
 void units_view::update_waiting()
 {
-  int units_count = 0;
+  QTableWidgetItem *item;
```
lmoureaux, line 285: Declare variables when they are first used.

client/views/view_units.cpp
```diff
@@ -262,57 +282,66 @@ void units_view::update_view()
...
-        char buf[64];
-        format_time_duration(-dt, buf, sizeof(buf));
-        ui.uwt_widget->setItem(units_count, 3,
-                               new QTableWidgetItem(buf, dt));
-      }
+  get_units_waiting_data(unit_entries, &entries_used);
+
+  max_row = 0;
+  int i, j;
+  for (i = 0; i < entries_used; i++) {
```
lmoureaux, line 300:
```suggestion
  for (int i = 0; i < entries_used; i++) {
```

client/views/view_units.cpp
```diff
@@ -262,57 +282,66 @@ void units_view::update_view()
...
+    struct unit_waiting_entry *pentry = unit_entries + i;
+    struct unit_type *putype = pentry->type;
+    cid id;
+
+    auto sprite = get_unittype_sprite(tileset, putype, direction8_invalid());
+    id = cid_encode_unit(putype);
 
-      ++units_count;
+    ui.uwt_widget->insertRow(max_row);
+    for (j = 0; j < 3; j++) {
```
lmoureaux, line 309:
```suggestion
    for (int j = 0; j < 3; j++) {
```

client/views/view_units.cpp
```diff
@@ -262,57 +282,66 @@ void units_view::update_view()
...
-                               new QTableWidgetItem(buf, dt));
-      }
+  get_units_waiting_data(unit_entries, &entries_used);
+
+  max_row = 0;
+  int i, j;
+  for (i = 0; i < entries_used; i++) {
+    struct unit_waiting_entry *pentry = unit_entries + i;
+    struct unit_type *putype = pentry->type;
+    cid id;
```
lmoureaux, line 303: Declare when first used

client/views/view_units.cpp
```diff
@@ -262,57 +282,66 @@ void units_view::update_view()
...
+
+    auto sprite = get_unittype_sprite(tileset, putype, direction8_invalid());
+    id = cid_encode_unit(putype);
 
-      ++units_count;
+    ui.uwt_widget->insertRow(max_row);
+    for (j = 0; j < 3; j++) {
+      item = new QTableWidgetItem;
```
lmoureaux, line 310: Declare when first used
```suggestion
      auto item = new QTableWidgetItem;
```

client/views/view_units.cpp
```diff
@@ -394,23 +425,48 @@ void units_view::disband_units()
...
- * Find nearest unit
+ * Find nearest selected unit, closes units view when button is clicked.
  */
 void units_view::find_nearest()
 {
-  // go look at old fc gtk and see how its done there
+  struct universal selected;
+  struct tile *ptile;
+  struct unit *punit;
+  const struct unit_type *utype;
```
lmoureaux, line 435: Declare variables when first used

client/views/view_units.cpp
```diff
@@ -394,23 +425,48 @@ void units_view::disband_units()
...
+ */
 void units_view::upgrade_units()
 {
   QString b, c;
   hud_message_box *ask = new hud_message_box(king()->central_wdg);
+  struct universal selected;
   int price;
   const struct unit_type *upgrade;
-  const Unit_type_id type = utype_number(utype);
+  const struct unit_type *utype;
```
lmoureaux, line 464: Declare variables when first used

client/views/view_units.cpp
```diff
@@ -434,7 +490,200 @@ void units_view::upgrade_units()
...
           [=]() { dsend_packet_unit_type_upgrade(&client.conn, type); });
   ask->show();
-} */
+}
+
+/****************************************
+ * Helper functions related to units view
+ ****************************************/
+
+/*
```
lmoureaux, line 499:
```suggestion
/**
```

client/views/view_units.cpp
```diff
@@ -434,7 +490,200 @@ void units_view::upgrade_units()
...
+{
+  QImage img = sprite->toImage();
+  QRect crop = zealous_crop_rect(img);
+  QImage cropped_img = img.copy(crop);
+  QPixmap pix = QPixmap::fromImage(cropped_img);
+
+  return pix;
+}
+
+/*
```
lmoureaux, line 512:
```suggestion
/**
```

client/views/view_units.cpp
```diff
@@ -434,7 +490,200 @@ void units_view::upgrade_units()
...
+
+  std::sort(entries, entries + *num_entries_used,
+            [](const auto &lhs, const auto &rhs) {
+              return QString::localeAwareCompare(
+                         utype_name_translation(lhs.type),
+                         utype_name_translation(rhs.type))
+                     < 0;
+            });
+
+  return {};
```
lmoureaux, line 619: Guess you shouldn't use `std::vector` if you're anyway always returning an empty one

client/views/view_units.cpp
```diff
@@ -434,7 +490,200 @@ void units_view::upgrade_units()
...
```
lmoureaux, line 681: Guess you shouldn't use `std::vector` if you're anyway always returning an empty one

lmoureaux (commented):
Many comments, but you'll see that they are quite repetitive and minor. With this fixed we're good to go

client/views/view_units.cpp
jwrober: I'm not sure it works. Will need to come back at it when we change it to a `QTreeView`. I'm going to leave as is for now.

jwrober:
Once this is merged, I will open issues for:

1. (bug) Getting global observing done right.
2. (bug) First column text should be left-aligned.
3. (enh) Convert to `QTreeView`
4. (enh) Look at changing the find nearest button to bring up the advanced unit selection box.
5. (enh) The Upgrade button could be more informative about how many units will be upgraded and the total cost.
6. (enh) UWT Double clicking a unit should focus it on the map.

lmoureaux (approved):

PR #1811, opened 2023-02-08
Rewrite the notify report with widgets

lmoureaux:
Use a QLabel instead of painting it directly. Enable text interaction so the text can be copied.


jwrober:
What is the notify report? Not sure what to evaluate on this.

lmoureaux:
Demographics, top five cities, wonders, historian reports

client/notifyreport.cpp
```diff
@@ -9,135 +9,68 @@
...
 
-  for (i = 0; i < str_list.count(); i++) {
-    x = qMax(x, fm.horizontalAdvance(str_list.at(i)));
-    y = y + 3 + fm.height();
-  }
-  y += fm.height();
-  x = x + 15;
-}
+  auto cw = new close_widget(this);
+  layout->addWidget(cw, 0, 1, Qt::AlignTop | Qt::AlignRight);
```
jwrober, line 49: In the other widgets with the `close_widget`, this method is used to place it `cw->put_to_corner();` that places the widget in the upper right corner. The way this is shown here is different. For consistency I think the close widget should be pushed to the corner.  Otherwise, looks great.

jwrober (changes requested):

jwrober (approved):
The requested change looks good. Since this is a refactor, might think about two things:
1. Move the file to the `client/widgets` directory and name it simply `notify`.
2. Add a \File: comment at the top for doxygen documentation

lmoureaux:
Since you were confused by the name `notify_widget`, I suggest to rename it `report_widget`.

jwrober:
That would work too. The \File comment will also help too, which is something I think we should look at as a common coding guideline.

lmoureaux:
> The \File comment will also help too, which is something I think we should look at as a common coding guideline.

It's often redundant when we have only one class per file, which is good practice. (Not that I'm against it, but it's not always needed.)

lmoureaux:
Pushed refactoring further

jwrober (approved):
Looks good to me.

PR #1813, opened 2023-02-11
Don't hard code master for autorevision and source tarballs

lmoureaux:
We released 3.0-rc.3 with code from master. This should not happen.

Also update the CI so it can push new version numbers to `stable` when we make RC releases and document when `stable` gets created (as a first draft to be revisited when we reach 3.1 and make a new `stable` branch).

The branch part was tested in https://github.com/lmoureaux/freeciv21/pull/4.

Backporting recommended.

### DIFF
--- a/docs/Contributing/stable-branch.rst
+++ b/docs/Contributing/stable-branch.rst
@@ -1,5 +1,6 @@
 .. SPDX-License-Identifier: GPL-3.0-or-later
 .. SPDX-FileCopyrightText: James Robertson <jwrober@gmail.com>
+.. SPDX-FileCopyrightText: Louis Moureaux <m_louis30@yahoo.com>
 
 Maintaining the Stable Branch
 *****************************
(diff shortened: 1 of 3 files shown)


docs/Contributing/stable-branch.rst
```diff
@@ -14,6 +15,8 @@ occasions where we are introducing major breaking changes that take time to reso
 ``stable`` branch. As development occurs on the ``master`` branch, there are going to be times when we want to
 back-port a commit (a single patch) or a Pull Request (a collection of commits) over to the ``stable`` branch.
 
+The ``stable`` branch is created when preparing for the first Release Candidate towards a stable release, see
+:doc:`release`.
 This page documents the rules and procedures for maintaining the ``stable`` branch.
```
jwrober, line 20: This is not part of the PR, but since you are in there would you change to 

This page documents the rules and procedures for the release process, which includes some maintenance activities for the ``stable`` branch.

jwrober (approved):
You don't have to do the update, but would be nice.

docs/Contributing/stable-branch.rst
lmoureaux: I don't think your sentence fits the page about `stable`. I added a mention of `stable` branch in The Release Process.

jwrober: Looks good.

PR #1820, opened 2023-02-17
AutoRevision update

lmoureaux:
This contains a collection of updates to AutoRevision:

* Pick the latest tag in the current branch, not overall, to fix issues encountered by panch with tags from my fork
* Sanitize the tag name and fail early if we don't like it
* Add support for `-dev` mode and turn `master` to dev mode

Caveat: When creating the `v3.1-alpha.1` tag on top of a `v3.1-dev`, the version will stay at `v3.1-dev` until the AutoRevision PR gets merged. Even after it is merged, someone doing a git checkout of `v3.1-alpha.1` will get the `v3.1-dev` tag.


jwrober (approved):
It works for me. I do agree with the other issue opened that this is starting to get unwieldy and we need a different approach.

lmoureaux:
Suggest to backport [6e41925](https://github.com/longturn/freeciv21/pull/1820/commits/6e41925f1b43dfda376302ec82ed9447255f662b)

PR #1821, opened 2023-02-18
Docs: Update coding guidelines

jwrober:
Mostly based on commentary while working on new units view. No related issue.

### DIFF
--- a/docs/Coding/guidelines.rst
+++ b/docs/Coding/guidelines.rst
@@ -42,7 +42,7 @@ New code follows the `SPDX standard <https://spdx.dev/ids/>`_:
 
 .. code-block:: cpp
 
-    /**
+    /*
      * SPDX-License-Identifier: GPL-3.0-or-later
      * SPDX-FileCopyrightText: Freeciv21 and Freeciv Contributors
      * SPDX-FileCopyrightText: Author Name <how-to-contact@example.com>
@@ -156,8 +156,7 @@ initial value right before it is used.
     ... some code
     ... some code
 
-    int i;
-    for (i = 0; i < max_item; i++) {
+    for (int i = 0; i < max_item; i++) {
       ... do something in the loop
     }
 
@@ -294,7 +293,9 @@ Containers in the Standard Library should be preferred over Qt ones:
 One notable exception is ``QStringList``, which should be preferred over other constructs because it
 integrates better with Qt.
 
-The main point here is to avoid using function parameters to return values.
+This can have consequences on function signatures. For instance, a common C idiom is to pass an output array
+and a size as function parameter, whereas this is achieved in C++ by returning a ``std::vector`` (or, if
+performance is critical, passing a reference to a vector that the function fills).
 
 
 Use ``<algorithm>``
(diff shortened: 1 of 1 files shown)


docs/Coding/guidelines.rst
```diff
@@ -30,18 +32,21 @@ manually. The downside is that the tool is really picky and will sometimes want
...
 
 .. code-block:: cpp
 
-    // SPDX-License-Identifier: GPL-3.0-or-later
-    // SPDX-FileCopyrightText: 2022 Author Name <how-to-contact@example.com>
+    /**
+     * SPDX-License-Identifier: GPL-3.0-or-later
+     * SPDX-FileCopyrightText: Freeciv21 and Freeciv Contributors
+     * SPDX-FileCopyrightText: Author Name <how-to-contact@example.com>
+     */
```
lmoureaux, line 49: Aaah no, no `/**` Doxygen here, we don't want this in the docs!

I have a slight preference towards `//`-comments because they're used in Kde, but I've done both and I'd say both are accepted.

docs/Coding/guidelines.rst
```diff
@@ -128,6 +135,32 @@ definition, as well as for less complex constructs such as enumerations and vari
...
+When encountering this older style or writing new code, it is best to define the variable and give it an
+initial value right before it is used.
+
+.. code-block:: cpp
+
+    ... some code
+    ... some code
+
+    int i;
+    for (i = 0; i < max_item; i++) {
```
lmoureaux, line 160:
```suggestion
    for (int i = 0; i < max_item; i++) {
```

docs/Coding/guidelines.rst
```diff
@@ -261,6 +294,9 @@ Containers in the Standard Library should be preferred over Qt ones:
 One notable exception is ``QStringList``, which should be preferred over other constructs because it
 integrates better with Qt.
 
+The main point here is to avoid using function parameters to return values.
```
lmoureaux, line 297: Not really the main point (which is rather avoiding explicit memory management), but a nice side-effect. Would tone down a bit.

docs/Coding/guidelines.rst
jwrober: I was going with the `/*` style for multi-line comments and `//` style for single-line comments. 

> Aaah no, no /** Doxygen here, we don't want this in the docs!

What do you mean here?

docs/Coding/guidelines.rst
jwrober: If you have a wording suggestion, I'd gladly add it in next commit.

lmoureaux: Feel like it's a bit specific for this style guide, but here we go:
> This can have consequences on function signatures. For instance, a common C idiom is to pass an output array and a size as function parameter, whereas this is achieved in C++ by returning a `std::vector` (or, if performance is critical, passing a reference to a vector that the function fills).

docs/Coding/guidelines.rst
lmoureaux: The SPDX headers should be normal `/*` (or `//`) comments, not `/**` or `///`

jwrober: ok for the record we are all over the map in existing code

PR #1831, opened 2023-02-22
Docs: Stop creating PDF version, build is failing

jwrober:
No related issue. Just noticed that recent changes were not loading.


lmoureaux:
Let's fix instead, #1832

PR #1837, opened 2023-02-23
Add guidelines for math usage in the documenation

### LINKED ISSUE #1719: Document the usage of latex math
Freeciv21 is math heavy and our documentation has a lot of mathematical formulas in it to help players understand how varying elements of the game are calculated. RTD allows us to use latex formatting for the `:math:` directive, however it is a bit of alchemy to a regular documentation writer unless you have used latex before.

**Do you have suggestions?**
In the contributing section, could have a page that explains how to format math formulas using the latex feature of the `:math:` directive.

lmoureaux:
Keep the topic short because math should be used sparingly. In particular, a full Latex reference is out of scope here; Latex is widespread enough that there are countless resources on the topic. The syntax is also way to complex to explain in a single page.

Closes #1719.

### DIFF
--- a/docs/Contributing/style-guide.rst
+++ b/docs/Contributing/style-guide.rst
@@ -244,7 +244,7 @@ code could be typeset in monospace with ``\texttt{}``, or defining a few symbols
 a long reasoning. The main guideline for formulas is to take your time to make them as readable as possible.
 
 Formulas use the ``:math:`` role or the ``.. math::`` directive. These blocks support most of the LaTeX
-mathematics syntax, for which you will find many references online. The ``:math:`` role is used for inline
+`mathematics syntax <https://en.wikibooks.org/wiki/LaTeX/Mathematics>`_. The ``:math:`` role is used for inline
 math in a paragraph. For instance, ``:math:`a+b=1``` becomes :math:`a+b=1`. The directive is used for longer
 or more important formulas that come on their own line:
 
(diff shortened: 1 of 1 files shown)


jwrober:
That was fast :)

jwrober (approved):
I think this is good. My only comment would be to aid the reader with a link to a "good source" to help those that might be overwhelmed with all the LaTeX documentation to know which one to use.

I do like the "use sparingly" comment. Maybe more emphasis on usage in  the coding section is appropriate, while in manuals is probably not appropriate as an example?

lmoureaux:
> My only comment would be to aid the reader with a link to a "good source" to help those that might be overwhelmed with all the LaTeX documentation to know which one to use.

Added a Wikibooks I learned a lot from, it's however more a reference than a tutorial.

> Maybe more emphasis on usage in the coding section is appropriate, while in manuals is probably not appropriate as an example?

Given how often we get questions about the price of unit upgrades, I think we shouldn't push back too hard on  math in "normal" docs.

jwrober:
> Given how often we get questions about the price of unit upgrades, I think we shouldn't push back too hard on math in "normal" docs.

Simple formulas I think are fine anywhere. It's the super complex ones that we have in the coding area that I think need to stay there. "Normal" docs should be readable by anyone with Algebra I level math knowledge (e.g. what every US high school teaches).

lmoureaux:
Wording suggestion?

jwrober:
When math formulas are used on non-technical pages (such as any one of the manuals), the reasoning should be relatively simple following elementary algebra [https://en.wikipedia.org/wiki/Elementary_algebra]

PR #1841, opened 2023-02-25
Move governor settings widget to its own class with a .ui

lmoureaux:
As part of the general city dialog rework in https://github.com/longturn/freeciv21/issues/1591, move the widget used to set
governor parameters to its own file. Implement the layout in a .ui file for
easier edits and rework it:

* Add labels with slider values.
* Set slider pageSteps to 1 for more precise edits.
* Expose max_growth, allow_specialists, and allow_disorder.
* Add accelerator shortcuts.
* Improve typography, in particular removing usage of the notify_label font.

The widget is designed to be used horizontally in the future and thus doesn't
fit very well in the current column-based layout.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -137,6 +137,7 @@ target_sources(
   views/view_units.cpp
   widgets/decorations.cpp
   widgets/city/city_icon_widget.cpp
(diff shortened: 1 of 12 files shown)


jwrober (approved):
Looks great.

PR #1843, opened 2023-03-01
Update AutoRevision.txt to 3.1-dev

jwrober:
No related issue.


cmake/AutoRevision.txt
```diff
@@ -1,2 +1,2 @@
-8dcf1
-v3.0-dev
+9e418
```
lmoureaux, line 1: Don't update the commit, it's the branching point

jwrober: Oh didn't realize that. I thought it was latest id. I'll fix.

PR #1873, opened 2023-03-20
Fix loading unhappy sprites when there is more than one

### LINKED ISSUE #1859: Supported units that cause unhappiness not shown in city dialog
**Describe the bug**
If you have units in enemy territory that cause unhappiness to the city that is supporting them, the city dialog does not show which unit(s) is causing the unhappiness. There doesn't seem to be any way to see that.

**To Reproduce**
1. Open a game
2. Get some units from a known city into enemy territory or an area that would cause unhappiness
3. Open the city dialog for the city the unit is supported by

**Expected behavior**
There is tileset feature to show an angry face / angry citizen embellishment next to the unit just like we have for gold, shield and food upkeep.

**Screenshots**
N/A

 - OS: Linux
 - Freeciv21 version: 3.0-rc.3

**Additional context**
Noticed while playing LT76Team game

lmoureaux:
We were looking for "upkeep.unhappy1", which is actually "upkeep.unhappy" for
historical reasons. So start by looking for "upkeep.unhappy2". Do the same for
other upkeep sprites.

Closes #1859.

Backport candidate.

### DIFF
--- a/client/tileset/tilespec.cpp
+++ b/client/tileset/tilespec.cpp
@@ -2757,18 +2757,18 @@ static void tileset_lookup_sprite_tags(struct tileset *t)
     auto sprite = load_sprite(t, buffer);
     if (sprite) {
       t->sprites.upkeep.output[o].push_back(sprite);
-    }
-    // Start from "upkeep.food2"; there is no "upkeep.food1"
-    for (i = 2;; i++) {
-      buffer =
-          QStringLiteral("upkeep.%1%2")
-              .arg(get_output_identifier(static_cast<Output_type_id>(o)),
-                   QString::number(i + 1));
-      auto sprite = load_sprite(t, buffer);
-      if (!sprite) {
-        break;
+      // Start from "upkeep.food2"; there is no "upkeep.food1"
+      for (i = 2;; i++) {
+        buffer =
+            QStringLiteral("upkeep.%1%2")
+                .arg(get_output_identifier(static_cast<Output_type_id>(o)),
+                     QString::number(i));
+        auto sprite = load_sprite(t, buffer);
+        if (!sprite) {
+          break;
+        }
+        t->sprites.upkeep.output[o].push_back(sprite);
       }
-      t->sprites.upkeep.output[o].push_back(sprite);
     }
   }
   output_type_iterate_end;
--- a/client/tileset/tilespec.cpp
+++ b/client/tileset/tilespec.cpp
@@ -2741,7 +2741,8 @@ static void tileset_lookup_sprite_tags(struct tileset *t)
   if (!t->sprites.upkeep.unhappy.back()) {
     tileset_error(t, LOG_FATAL, "Missing sprite upkeep.unhappy");
   }
-  for (i = 1;; i++) {
+  // Start from "upkeep.unhappy2"; there is no "upkeep.unhappy1"
+  for (i = 2;; i++) {
     buffer = QStringLiteral("upkeep.unhappy%1").arg(QString::number(i));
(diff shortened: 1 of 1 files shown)


jwrober (changes requested):
I know it would be a bit janky, but I think we need to start with `0/blank`, skip `1` with an `if()` and then continue from `2` onwards. That way we get all the sprites to show appropriate instead of completely skipping the first sprite called altogether.

lmoureaux:
> I know it would be a bit janky, but I think we need to start with `0/blank`, skip `1` with an `if()` and then continue from `2` onwards. That way we get all the sprites to show appropriate instead of completely skipping the first sprite called altogether.

The sprites to load are `upkeep.unhappy`, `upkeep.unhappy2`, `upkeep.unhappy3`, etc. There is no `upkeep.unhappy0` because no sprite is shown when there is no upkeep. Adding a sprite at 0 would be a new feature... I'm not opposed to it but it would need to be optional (couldn't be part of the loop) and the patch shouldn't be backported.

jwrober:
Right I agree with you. Not asking for a new `.unhappy0`. Just asking to go from `upkeep.unhappy` to `upkeep.unhappy2` and skip `1`. Right now your loop starts at `upkeep.unhappy.2` and ignores the first entry.

lmoureaux:
hmm then what you actually want is start at 1 and have a special case for the name

lmoureaux:
> Just don't think a "fix" should ignore a valid sprite.

Coming back to this... which valid sprite is ignored by the fix?

jwrober:
Column 0, e.g. the very first one on the left.

lmoureaux:
This is what the loop would look like:
```cpp
  for (i = 1;; i++) {
    if (i == 1) {
      buffer = QStringLiteral("upkeep.unhappy");
    } else {
      buffer = QStringLiteral("upkeep.unhappy%1").arg(QString::number(i));
    }
    auto sprite = load_sprite(t, buffer);
    if (!sprite) {
      break;
    }
    t->sprites.upkeep.unhappy.push_back(sprite);
  }
  if (t->sprites.upkeep.unhappy.empty()) {
    tileset_error(t, LOG_FATAL, "Missing sprite upkeep.unhappy");
  }
```
It's functionally identical to the current code. Not sure if it's more readable...

jwrober:
The current code does a `for (i = 2...` instead of a `for (i = 1...` and then adds another to `i` with `QString::number(i + 1));`, which on first pass `i` would be 3. I think what you have above looks much better to me.

lmoureaux:
Aaah I see it now, I was supidly assuming that the two loops (unhappy and food) were identical, but they're not!

jwrober:
This code update still skips the graphic for 1 upkeep. Amplio and Hexemplio both show upkeep as a quantity, so skipping the first icon won't be accurate.

lmoureaux:
I have to admit it's a bit weird, but there is no `upkeep.something1`:
https://github.com/longturn/freeciv21/blob/8da632265df6a3c8b2b1c51d1f850dc5c0d0da93/data/amplio2/upkeep.spec#L27-L36

jwrober:
Agree dealing with the legacy tileset code is a pain. What about using the code you put in an earlier comment to account for the lack of a `1` for `upkeep`? I think that would be better than simply skipping it altogether.

lmoureaux:
I don't skip it, the special case for the first sprite is handled above the loop as it was before (I didn't touch that part of the code). Or is it just a stylistic preference?

jwrober (approved):
Ok, with some help from friends I better understand what is going on here. Approved.

PR #1875, opened 2023-03-20
Fix military upkeep tooltip

lmoureaux:
There was probably a call to city_refresh_from_main_map when it was written (or I was using a governor).

See #1873.

### DIFF
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -1884,30 +1884,33 @@ QString text_happiness_units(const struct city *pcity)
   }
 
   // And finally how many unhappy we get in this city
-  int count = 0;
+  int count = 0, happy_upkeep = 0;
   unit_list_iterate(pcity->units_supported, punit)
   {
     int dummy = 0;
-    if (city_unit_unhappiness(punit, &dummy) > 0) {
+    if (auto n = city_unit_unhappiness(punit, &dummy); n > 0) {
       count++;
+      happy_upkeep += n;
     }
   }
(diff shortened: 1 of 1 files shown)


client/text.cpp
```diff
@@ -1884,30 +1884,33 @@ QString text_happiness_units(const struct city *pcity)
...
-    str += QString(PL_("resulting in <b>%1 additional unhappy citizen.</b>",
-                       "resulting in <b>%1 additional unhappy citizens.</b>",
-                       pcity->unit_happy_upkeep))
-               .arg(pcity->unit_happy_upkeep);
+    // TRANS: "This city supports X agressive military units, resulting in up
+    //        to %1 unhappy citizens." (second part)
+    str +=
+        QString(
+            PL_("resulting in up to <b>%1 additional unhappy citizen.</b>",
+                "resulting in up to <b>%1 additional unhappy citizens.</b>",
```
jwrober, line 1911: I think the `</b>` should be before the period. You don't need to bold the period. Otherwise rest looks great to me.

jwrober (approved):

client/text.cpp
lmoureaux: This would be [against typographic rules](https://writing.stackexchange.com/a/14161).

lmoureaux:
I suppose this should be backported

client/text.cpp
jwrober: Interesting.

PR #1877, opened 2023-03-20
Fix installing dependent modpacks with relative URLs

lmoureaux:
The URL of the dependency wasn't used correctly. Noticed while developing RoundHex.

Old fix I never committed. Test by pasting the RoundHex modpack URL into the installer:
* https://raw.githubusercontent.com/lmoureaux/RoundSquare/hex/RoundHex.json

Backport proposed.


jwrober (approved):
Works for me.

Issue #1250, opened 2022-08-11
Retrofit existing docs to updated style guide


jwrober:
**What should be documented? Is there something wrong in the documentation?**
Existing documents in the documentation repository need to be updated to follow new style guidelines.

**Do you have suggestions?**
* Figures need to be converted to use `:numref:`
* Capitalization rules
* Oxford Comma rules
* The usages of contractions (don't vs do not)

jwrober:
PR #1311 tackles `docs/Contributing`

PR #1878, opened 2023-03-22
Docs/Getting style guide update

jwrober:
Part of #1250


jwrober:
This isn't ready yet. Just noticed that the images in windows-install need to be converted to figures. Also the images themselves are really old and could use an update.

jwrober:
Ok, much better. Turns out the images are still valid so no need to make the repo larger with new versions.

daavko (approved):
LGTM

PR #1879, opened 2023-03-23
Docs/Manuals style update

jwrober:
Part of #1250 

Also came back to `Coding` to include the `:table:` directive for `:numref:` and updated the style guide to call out the use of `:figure:` and `:table:`.

### DIFF
--- a/docs/Manuals/modpack-installer.rst
+++ b/docs/Manuals/modpack-installer.rst
@@ -8,8 +8,7 @@ Using the Modpack Installer Utility
 ***********************************
 
 The modpack installer is a simple tool to download custom Freeciv21 content called :strong:`modpacks` from the
-Internet and automatically install them into the correct location to be used by the Freeciv21 game interface
-and server.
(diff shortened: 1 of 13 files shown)


docs/Manuals/modpack-installer.rst
```diff
@@ -1,15 +1,15 @@
...
+
 
 Using the Modpack Installer Utility
 ***********************************
 
-The modpack installer is a simple tool to download custom Freeciv21 content called :strong:`modpacks` from
-the Internet, and automatically install them into the correct location to be used by the Freeciv21 client and
-server.
+The modpack installer is a simple tool to download custom Freeciv21 content called :strong:`modpacks` from the
+Internet and automatically install them into the correct location to be used by the Freeciv21 game interface
```
daavko, line 11: Would it be better to say "used by Freeciv21" here?

daavko (commented):
one note, other than that LGTM

PR #1897, opened 2023-04-05
Packaged for Nix/ NixOS

dmadisetti:
Packaged for the Nix package manager (which is usable on Mac / NixOS/ general linux)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -279,6 +279,20 @@ jobs:
           name: macOS-dmg
           path: Freeciv21-*.dmg
 
+  nix:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - uses: cachix/install-nix-action@v19
+      - run: nix flake check
+      # TODO: We can add the build to a cache for others (see cachix.org)
+      # It's free for FOSS.
+      - run: nix build .#freeciv21
+
   wasm:
     name: "WebAssembly"
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ build.ninja
 # Build directories
 /build*
 
+# Nix build artifacts
+result
+
 # IDE files
 .cache
 .clangd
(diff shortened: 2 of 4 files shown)


lmoureaux:
Thank you for the contribution! I'm glad you could figure out the `-DFREECIV_DOWNLOAD_FONTS=0` :+1:
I have a few questions, please forgive my ignorance about Nix (looks like it's packaged for Ubuntu/Debian, I'll give it a try):
* How can we test this to make sure we don't introduce regressions? Maybe upstream provides a Github action we can use?
* You mentioned MacOS. Do you know if Nix can be used to create dmgs? We're currently using vcpkg and it's both very slow and quite unstable

dmadisetti:
Yes! We can use: github.com/cachix/install-nix-action. Tests can be added to the derivation as well to ensure valid builds too.

Hmm, I don't know enough about dmgs to really comment. But Nix is developed specifically to address instability

dmadisetti:
Not sure if it'll run for MacOS. It should? I have no way of testing beyond a quick cross compilation check. I guess the CI will tell. The only dependency I couldn't resolve was Qt5's Karchive, but that's already present on MacOS? Wasn't clear on that.  I only looked at x86 as well, but that could potentially be expanded.

I can add a snippet to https://github.com/longturn/freeciv21/blob/master/docs/Getting/compile.rst as well. Once this is merged, nix users can just do `nix run github:longturn/freeciv21#` .

dmadisetti:
@lmoureaux Can you run the CI?

lmoureaux:
Thanks for the updates, I had missed them since this was a busy week. I'll look more over the week-end.

dmadisetti:
No problem, thanks. 

How is KF5Archive dependency resolved on macos? I don't see any explicit call in the other workflows and nix doesn't package it for mac

lmoureaux:
It's resolved through vcpkg, which has an automatic way of discovering dependencies from CMake. It compiles it from source

lmoureaux:
I suggest we leave MacOS out and merge this. Let's also make it a weekly build so we don't use compute needlessly but still check for regressions (like `windows_clang_msvc`).

jwrober:
Would appreciate if we put the `flake.nix` in the `dist` directory. This is where we store flatpak and snapcraft yaml files. Docs updates are always encouraged.

jwrober:
appreciate the submission @dmadisetti

lmoureaux:
I can't edit the branch in this PR -> superseded by #1934

PR #1906, opened 2023-04-07
Fix units view to show when observing

### LINKED ISSUE #1826: Units view does not do global observing correctly
**Describe the bug**
When global observing with the new units view, the view is empty.

**To Reproduce**
1. Open a local game
2. In chat enter `/observe` to enter global observe
3. Open units view

**Expected behavior**
Similar to the cities and economics view, I would expect to see all units for all nations.

**Screenshots**
N/A

 - OS: All supported
 - Freeciv21 version: v3.1-dev (master) at https://github.com/longturn/freeciv21/commit/20ff9bdf0c7441535aec6aa0798cd7f22c77e659
 - Ruleset/Longturn game (if applicable): Any

**Additional context**

jwrober:
Closes #1826

### DIFF
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -138,7 +138,7 @@ void units_view::update_view()
 
   for (int i = 0; i < entries_used; i++) {
     struct unit_view_entry *pentry = unit_entries + i;
-    struct unit_type *putype = pentry->type;
+    const struct unit_type *putype = pentry->type;
     cid id;
 
     auto sprite = get_unittype_sprite(tileset, putype, direction8_invalid());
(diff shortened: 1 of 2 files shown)


client/views/view_units.cpp
```diff
@@ -531,61 +531,65 @@ void get_units_view_data(struct unit_view_entry *entries,
...
-    int in_progress = 0;     // Count of being produced
-    int gold_cost = 0;       // Gold upkeep
-    int food_cost = 0;       // Food upkeep
-    int shield_cost = 0;     // Shield upkeep
-    bool upgradable = false; // Unit type is upgradable
-
-    unit_list_iterate(client_player()->units, punit)
+    if (client_has_player() && pplayer != client_player()) {
+      continue;
+    }
```
lmoureaux, line 538: How is this different from doing the following?
```cpp
auto pplayer = client_player();
if (!pplayer) {
  return;
}
```

jwrober: I didn't know that was an option I guess. Is the code better your way?

jwrober: I test it with your code and it breaks the view.

PR #1913, opened 2023-04-13
Remove extra column in econ view

### LINKED ISSUE #1818: The Economics View has an extra unused header column
**Describe the bug**
The Economics View has an extra header column that doesn't show any data and if clicked will expand to fill in the width of the table.

**To Reproduce**
1. Open a game
2. Click on economics view
3. Look at the right edge of the table header and notice the extra "box", click on it

**Expected behavior**
No extraneous columns in a table

**Screenshots**
N/A

 - OS: Linux, but assume all
 - Freeciv21 version: 3.0-rc.2+

**Additional context**
Will also show in new Units View (#1808) when merged as it was based on econ view.

jwrober:
Do some other cleanup items too such as making the icon bigger, centering it in the cell and set alternating row colors. 

Closes #1818

### DIFF
--- a/client/views/view_economics.cpp
+++ b/client/views/view_economics.cpp
@@ -29,9 +29,14 @@ eco_report::eco_report() : QWidget()
 {
   ui.setupUi(this);
 
+  QFont font = ui.eco_widget->horizontalHeader()->font();
+  font.setWeight(QFont::Bold);
+  ui.eco_widget->horizontalHeader()->setFont(font);
+
   QStringList slist;
   slist << _("Type") << Q_("?Building or Unit type:Name") << _("Redundant")
-        << _("Count") << _("Cost") << _("Upkeep Total");
+        << _("Count") << _("Cost") << _("Total Upkeep");
+
   ui.eco_widget->setColumnCount(slist.count());
   ui.eco_widget->setHorizontalHeaderLabels(slist);
   ui.eco_widget->setAlternatingRowColors(true);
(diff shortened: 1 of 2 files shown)


client/views/view_economics.cpp
```diff
@@ -31,9 +31,10 @@ eco_report::eco_report() : QWidget()
 
   QStringList slist;
   slist << _("Type") << Q_("?Building or Unit type:Name") << _("Redundant")
-        << _("Count") << _("Cost") << _("U Total") << QLatin1String("");
+        << _("Count") << _("Cost") << _("Upkeep Total");
```
lmoureaux, line 34: Wondering what sounds best, `Total Upkeep` or `Upkeep Total`?

client/views/view_economics.cpp
```diff
@@ -90,32 +91,37 @@ void eco_report::update_report()
     ui.eco_widget->insertRow(i);
     for (j = 0; j < 6; j++) {
       item = new QTableWidgetItem;
+      QLabel *lbl = new QLabel;
```
lmoureaux, line 94: You leak the label when `j != 0` (memory leak): it is neither deleted nor used. Simply create it when getting the sprite.

client/views/view_economics.cpp
```diff
@@ -124,40 +130,44 @@ void eco_report::update_report()
...
+    auto sprite = get_unittype_sprite(tileset, putype, direction8_invalid())
+                      ->scaledToHeight(h);
+
+    cid id = cid_encode_unit(putype);
 
     ui.eco_widget->insertRow(i + max_row);
     for (j = 0; j < 6; j++) {
       item = new QTableWidgetItem;
-      item->setTextAlignment(Qt::AlignHCenter);
+      QLabel *lbl = new QLabel;
```
lmoureaux, line 142: Another leaked label

client/views/view_economics.cpp
jwrober: How is that different than the `QTableWidgetItem`?

client/views/view_economics.cpp
jwrober: I think its probably better as "Total Upkeep" I'll change

PR #1916, opened 2023-04-18
Feature/flatpak

jwrober:
Support the flatpak system for Freeciv21. Will only work on stable 3.0 until we can get Qt6 working in `master`.

### DIFF
--- a/dist/flatpack.yaml
+++ b/dist/flatpack.yaml
@@ -1,0 +1,70 @@
+##########
+# Freeciv21 - flatpack.yaml
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: James Robertson <jwrober@gmail.com>
+#
+# Build Steps:
+#   - (once) sudo apt install flatpak-builder
+#   - (once) sudo flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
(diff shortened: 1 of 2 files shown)


dist/flatpack.yaml
```diff
@@ -0,0 +1,70 @@
...
+      - cp -R fonts/Libertinus/* /app/share/freeciv21/fonts/Libertinus
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=home
+  - --filesystem=~/.local/share/freeciv21:create
```
lmoureaux, line 66: We also need `~/.config`, do we get it by default?

dist/flatpack.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+##########
+# END flatpack.yaml
+##########
```
lmoureaux, line 70: That's kinda obvious, isn't it? :smile:

jwrober: Never assume my friend. LoL

dist/flatpack.yaml
jwrober: its built in. the one for `.local` is to ensure that we can create the `freeciv21` directory.

lmoureaux (approved):

PR #1918, opened 2023-04-21
Flatpak master

jwrober:
Add support for Flatpak to master branch. Note that it doesn't work right now as the platform file needs to have cmake 3.21 (has 3.20 now).


lmoureaux:
Where does CMake come from and can we get an idea of when we'll get a newer version?

jwrober:
> Where does CMake come from and can we get an idea of when we'll get a newer version?

The Platform file is the VM and is where CMake comes from. I have no idea if there are some kind of release notes posted somewhere. Could look/search or just run the script once a month to see when it "works". Then again with the KDE folks actively working on Qt6 updates to the Frameworks, we may get KArchive to Qt6 sooner than later and can push to Qt6 LTS on our end.

PR #1923, opened 2023-04-24
Units and Econ View Updates

jwrober:
With the work done in #1913, wanted to set units view to match. Also fixed the potential memory leak issue raised in the former PR.  No related issue for this PR.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -445,7 +445,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Run clang-format style check for C/C++
-      uses: jidicula/clang-format-action@v4.9.0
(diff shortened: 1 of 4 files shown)


lmoureaux (approved):
Seems to work in a very quick test. We'll want to adjust the row height to (multiples of) the actual icon size to minimize artifacts.

jwrober:
I have code to do the adjusting automagically via Qt functions, is that not good enough? Happy to fix if there is a better way.

lmoureaux:
> I have code to do the adjusting automagically via Qt functions, is that not good enough? Happy to fix if there is a better way.

I think you adjust to the font height, which makes sense for small icons but not for big ones. Not very important anyway, let's merge and keep for later

PR #1926, opened 2023-04-25
team mates can see same unit info as player owner

### LINKED ISSUE #282: Allow allies/team mates to see all details of units same as home nation sees
Comes in real handy with team based games.

jwrober:
Closes #282

### DIFF
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -2838,6 +2838,9 @@ void send_unit_info(struct conn_list *dest, struct unit *punit)
       }
     } else if (pplayer->team == powner->team) {
       send_packet_unit_info(pconn, &info);
+      if (pdata != nullptr) {
+        BV_SET(pdata->can_see_unit, player_index(pplayer));
+      }
     } else if (can_player_see_unit(pplayer, punit)) {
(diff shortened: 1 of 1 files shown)


lmoureaux:
Middle click doesn't seem to be using the newly available information.

jwrober:
Yea I def dorked that up. I'll look to push another commit later this week.

server/unittools.cpp
```diff
@@ -2836,6 +2836,11 @@ void send_unit_info(struct conn_list *dest, struct unit *punit)
       if (pdata != nullptr) {
         BV_SET(pdata->can_see_unit, player_index(pplayer));
       }
+    } else if (pplayer->team == powner->team) {
```
lmoureaux, line 2839: To facilitate maintenance I would prefer this (from `player.h`):
```suggestion
    } else if (players_on_same_team(pplayer, powner)) {
```
Could we merge this with the previous `if`? After all, a player is always on its own team...

lmoureaux:
Can you rebase for RTD?

lmoureaux (approved):
Let's merge this

PR #1931, opened 2023-05-05
Add +duplicates_ok to tilespec files

jwrober:
When opening the client from command line, we get a lot of noise of errors that isn't needed. Added `+duplicate_ok` cap string to the tilespec files we ship.

No related issue.

### DIFF
--- a/data/alio.tilespec
+++ b/data/alio.tilespec
@@ -2,7 +2,7 @@
 [tilespec]
 
 ; Format and options of this tilespec file:
-options = "+Freeciv-tilespec-Devel-2019-Jul-03"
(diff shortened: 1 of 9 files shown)


lmoureaux:
Why do we have duplicate tags in the first place? The warning may be telling us that we're shipping duplicate/unused sprites.

jwrober:
> Why do we have duplicate tags in the first place? The warning may be telling us that we're shipping duplicate/unused sprites.

I tried to figure it out and couldn't so thought suppressing the messages was better.

lmoureaux:
The warnings come from `trident` and `hex2t`. I think they're related to the mix of old-style and new-style wonder files.

jwrober:
Closing this as no longer needed. Will work to fix the tilesets now that we have #1935

PR #1933, opened 2023-05-06
Menu, Actions and Event capitalization standardization

jwrober:
No related issue. Took opportunity to clean up the capitalization and also some menu item names. No functional change, just cosmetic.


client/menu.cpp
```diff
@@ -258,7 +258,7 @@ void gov_menu::update_all()
 /**
    Instantiate a new goto and act sub menu.
  */
-go_act_menu::go_act_menu(QWidget *parent) : QMenu(_("Go to and..."), parent)
+go_act_menu::go_act_menu(QWidget *parent) : QMenu(_("Goto And..."), parent)
```
lmoureaux, line 261: Should we really merge `goto` or is it informal usage? My spellchecker isn't happy.

lmoureaux (commented):
Do we need to update the game manual to match?

Also one specific comment (appearing twice).

jwrober:
> Do we need to update the game manual to match?

Yea. I am working on another commit for the manual.

client/menu.cpp
jwrober: I did a quick search, seems "go to" is the correct way. I'll fix.

Issue #1895, opened 2023-04-05
Build error - Incompatible lua dependency though matches required


3dsf:
**Describe the bug**
Documentation state lua 5.3 or lua 5.4 should be compatible, though build fails

**To Reproduce**
1. Follow install procedures on git page
2. `cmake --build build`
**Errors that look like this**

> [7/295] Building CXX object common/scriptcore/CMakeFiles/scriptcore.dir/luascript.cpp.o
> FAILED: common/scriptcore/CMakeFiles/scriptcore.dir/luascript.cpp.o 
> /usr/bin/c++ -DLUA_LIB -DPUBLIC -DQT_CORE_LIB -DQT_DEPRECATED_WARNINGS -DQT_DISABLE_DEPRECATED_BEFORE=0x051500 -DQT_NETWORK_LIB -DQT_NO_DEBUG -DSOL_ALL_SAFETIES_ON=1 -DSOL_NO_CHECK_NUMBER_PRECISION=1 -I/home/3dsf/git/freeciv21/common/scriptcore -I/home/3dsf/git/freeciv21/build/common/scriptcore -I/home/3dsf/git/freeciv21/common -I/home/3dsf/git/freeciv21/build/common -I/home/3dsf/git/freeciv21/common/aicore -I/home/3dsf/git/freeciv21/common/networking -I/home/3dsf/git/freeciv21/utility -I/home/3dsf/git/freeciv21/build/utility -I/home/3dsf/git/freeciv21/dependencies/tolua-5.2/include -I/home/3dsf/git/freeciv21/dependencies/sol2 -isystem /usr/include/qt5 -isystem /usr/include/qt5/QtCore -isystem /usr/lib64/qt5/mkspecs/linux-g++ -isystem /usr/include/qt5/QtNetwork -std=gnu++17 -Wall -Wno-sign-compare -Warray-bounds -Wdelete-incomplete -Wint-in-bool-context -Wreturn-type -Wsequence-point -Wswitch -Wunused-value -Wmaybe-uninitialized -Wstringop-truncation -fPIC -MD -MT common/scriptcore/CMakeFiles/scriptcore.dir/luascript.cpp.o -MF common/scriptcore/CMakeFiles/scriptcore.dir/luascript.cpp.o.d -o common/scriptcore/CMakeFiles/scriptcore.dir/luascript.cpp.o -c /home/3dsf/git/freeciv21/common/scriptcore/luascript.cpp
[...]

PR #1936, opened 2023-05-06
Add a Fedora CI and review the dependency list

lmoureaux:
fter the debacle of 3.0 (https://github.com/longturn/freeciv21/issues/1895), I feel like it's a good idea to systematically check more environments. This introduces a check for the stable version of Fedora.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -279,6 +279,43 @@ jobs:
           name: macOS-dmg
           path: Freeciv21-*.dmg
 
+  fedora:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'schedule'
+    container: fedora:latest
+    steps:
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -279,6 +279,43 @@ jobs:
...
+            qt5-qtsvg-devel \
+            kf5-karchive-devel \
+            lua-devel \
+            sqlite-devel \
+            SDL2_mixer-devel \
+            readline-devel \
+            zlib-devel \
+            libunwind-devel \
+            elfutils-libs \
+            python3-sphinx
```
jwrober, line 305: If not needed in instructions, prolly not here either.

jwrober (approved):
Looks good and good idea. Only one small change.

.github/workflows/build.yaml
lmoureaux: Good catch :+1:

PR #1937, opened 2023-05-06
Remove infra points

### LINKED ISSUE #1929: Get rid of "infrapoints"
Apparently when we forked fc 3.0 beta, there was a pending server option called `infrapoints` that was never implemented. In authoring a server manual (#1927) I came across this feature option.

Since it isn't actually implemented, we should get rid of it.

**Additional context**

lmoureaux:
Infra points were an unfinished feature that would let players add extras to tiles without moving units there, "buying" them. We're not planning to support this.

This patch introduces an incompatible change: rulesets using Nation_Intelligence `Infra Points` cannot be loaded by new servers. Since this was only available in 3.0 (without UI support), this is an accepted risk.

Closes #1929.


jwrober (approved):
Looks good to me. Agree to risk at this early stage of 3.1 dev. 

There is also mention of infrapoints in new server manual that should be removed. That is what led to the issue.

Issue #1453, opened 2022-11-02
Update the summary of shipped rulesets to be more "rst" friendly


jwrober:
**What should be documented? Is there something wrong in the documentation?**
Following along from #1452. The short summary of many of the shipped rulesets don't parse well to be reStructuredText friendly.

**Do you have suggestions?**
Let's walk the rulesets we ship and adjust to include fewer new lines (\n) and whitespace.

Issue #1452, opened 2022-11-02
Detailed long description of rulesets need to be made "rst" friendly


jwrober:
**What should be documented? Is there something wrong in the documentation?**
We are working on a project to provide extremely detailed documentation of the shipped rulesets, plus other important rules such as aviation, ltt and ltx. In many cases those rules have a README file included that we want to include in our docs.

**Do you have suggestions?**
The files are formatted in plain text, however they also have formatting in them that makes them unsuitable for using in a reStructuredText file. Work to change them so they are readable as is, and also can be included as part of our docs system.

PR #1945, opened 2023-05-15
Civ1 and Civ2 helptext updates for RTD

jwrober:
Associated with output from pyciv. Part of #1452 and #1453.


data/civ1/buildings.ruleset
```diff
@@ -336,12 +336,12 @@ The amount of stored food will be set to half full whenever a city\
 ; In Civ1, city size reduction does not generate food like this.
 ; Dare I ask where this food comes from?? :-)
 _("\
-Famine can occur when food stock reaches zero due to a food deficit\
- (which a Granary merely delays), in which case it leads to the loss\
+Famine can occur when food stock reaches zero due to a food deficit,\
+ which a Granary merely delays. In which case it leads to the loss\
```
lmoureaux, line 340: While not incorrect, using a full stop before `in which case` seems to be far less common than using a comma.

lmoureaux (changes requested):
You seem to dislike colons and semicolons very much. These signs add nuance through punctuation: the colon introduces a list or explanation, and the semicolon is a pause slightly longer than a comma but shorter than a full stop and can be used, for instance, to introduce exceptions or caveats to the preceding clause. [Here](https://www.merriam-webster.com/words-at-play/a-guide-to-using-semicolons) is a guide from a reference institution for the English language; maybe we should add it to our guidelines?

lmoureaux (approved):

PR #1954, opened 2023-05-19
Multiplayer helptext updates for RTD

jwrober:
Associated with output from pyciv. Part of https://github.com/longturn/freeciv21/issues/1452 and https://github.com/longturn/freeciv21/issues/1453.

### DIFF
--- a/data/multiplayer/README.multiplayer
+++ b/data/multiplayer/README.multiplayer
@@ -1,24 +1,23 @@
-----------------------------------------------------------------------
-                   Multiplayer Freeciv Ruleset
-----------------------------------------------------------------------
+Multiplayer Freeciv21 Ruleset
+-----------------------------
 
-The 'multiplayer' ruleset is intended to rectify some game balance
+The "multiplayer" ruleset is intended to rectify some game balance
 issues encountered on multiplayer servers with the classic ruleset.
 It is based on the 'rewonder' ruleset designed by Madeline Book for
-Freeciv 2.1.
+legacy Freeciv 2.1.
 
 This ruleset is also suited for Longturn and Play-By-Email games.
 
 
-1. Changes compared to the classic ruleset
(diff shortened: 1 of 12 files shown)


data/multiplayer/README.multiplayer
```diff
@@ -1,24 +1,23 @@
...
 
-1. Changes compared to the classic ruleset
-   ---------------------------------------
+Changes Compared to the Classic Ruleset
+=======================================
 
 The differences can be extracted by
 
-#> diff -ur data/classic data/multiplayer
+$ diff -ur data/classic data/multiplayer
```
lmoureaux, line 17: Does this get rendered as fixed-width text?

data/multiplayer/README.multiplayer
```diff
@@ -93,63 +92,68 @@ As for the remaining great wonders:
...
 
 To play a game with the multiplayer rules, start the server with:
-       ./fcser -r data/multiplayer.serv
+
+$ ./freevic21-server -r data/multiplayer.serv
+
 or a single player game with
-       ./freeciv-gtk3 -r data/multiplayer.serv
+
+$ ./freeciv21-client -r data/multiplayer.serv
```
lmoureaux, line 146: Same question here

lmoureaux (approved):

data/multiplayer/README.multiplayer
jwrober: Nope, neither does the `#>`. that is why I went with the `$` make it look more like a command prompt.

data/multiplayer/README.multiplayer
jwrober: same answer. I didn't want to add a bunch of rst type code to the plain text file so tried to add simple stuff that would at least look like plain text but still render ok in RTD.

PR #1958, opened 2023-05-25
Fix displaying message turn and timestamp

### LINKED ISSUE #1942: ec_info option does not show turn and timestamp
**Describe the bug**
The server option `ec_info` is supposed to show the turn number and timestamp of messages in the chatline widget, but it doesn't.

**To Reproduce**
1. Open a game with the option enabled (LT78 is a good recent example)
2. Click on the server/chatline widget
3. See error

**Expected behavior**
There should be a turn number and server timestamp before each message line in the chatline when connecting to the game to show past messages.

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.1-alpha.1 and 3.0-stable

**Additional context**

lmoureaux:
The previous code was doing some dark magic to obtain an ok-looking result, but the generated HTML closing tags weren't in the correct order. In addition, it was neglecting any leftover out of any tags. This was working fine because most message content is enclosed within a color tag, but hid the timestamps that are not located inside a tag.

Simplify the implementation by merging the HTML tags together (in the correct order!) as they are generated. This way applying them becomes a simple string insertion task.

Also use the opportunity to refactor some related code slightly. We should eventually write a simple direct featured-text-to-HTML converter instead of using the intermediate "tags" step (whose architecture is a copy of Gtk's rich text functions and doesn't play well with Qt's).

Closes #1942.

### DIFF
--- a/client/chatline.cpp
+++ b/client/chatline.cpp
@@ -33,6 +33,7 @@
 #include "colors_common.h"
 #include "connectdlg_common.h"
 #include "control.h"
+#include "featured_text.h"
 #include "game.h"
 #include "views/view_map_common.h"
 // gui-qt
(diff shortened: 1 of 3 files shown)


jwrober (approved):
I can't speak too much on the code, however the change does produce the expected output.

PR #1960, opened 2023-05-27
NightStalker: make default chat color more neutral

lmoureaux:
When the turn and time are displayed next to messages, they use the default color. The saturated green we had earlier was clashing with message colors, in particular the ones in red. Switch to a slightly softened white to avoid this.

See commit f6d7d609b9a8c2a2445eac09b9f93e842da0644b.


jwrober (approved):
LGTM. Much better than the green that I could barely see :)

PR #1963, opened 2023-05-27
Guess the default username only after loading options

### LINKED ISSUE #1898: New user -- Username Issues
**Describe the bug**
Username required to start all games.  Issue not reported on console but at bottom of GUI.  Seen post from an unfortunate soul, #1466, which told me I could update it in the options menu.  After updating the username in the Options menu (**Apply** and **Save**) and trying to start a game, resulted in continued games not starting.  Closed and reopened client, then the username was respected   

**To Reproduce**
1. Start Client
2. Click 'Start new Game'  -- Game does not start
3. Update name in 'Options', Apply, Save
4. Click 'Start new Game'  -- Game does not start
5. Close and reopen client
6. Click 'Start new Game'  -- Game starts

**Expected behavior**
Game to start on first start of game press or maybe an add name dialog

 - OS: Fedora 37
 - Freeciv21 version: 3.1 dev
 - Ruleset/Longturn game (if applicable): NA

**Additional context**
I think this is a bug, but might also be an enhancement request for an enter name dialog if the username is empty.  
I'm happy you have decided not to auto-populate the computers users login name.

lmoureaux:
Loading options resets them to their defaults, so we can only override them after they are loaded. This has the side-effect of using a default username if the user sets it to something invalid (which is good because Freeciv21 is unusable without a username).

Closes #1898.

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -508,8 +508,12 @@ int client_main(int argc, char *argv[])
   configure_fonts();
   init_themes();
 
+  options_load();
+
   // Set default username
-  (void) user_username(gui_options->default_user_name, MAX_LEN_NAME);
+  if (!is_valid_username(gui_options->default_user_name)) {
+    (void) user_username(gui_options->default_user_name, MAX_LEN_NAME);
+  }
   if (!is_valid_username(gui_options->default_user_name)) {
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Works for me.

Issue #1899, opened 2023-04-05
Fonts not loaded on fresh compile


jwrober:
**Describe the bug**
This may also be an issue with install from one of our packages. If you have a "blank slate" so to speak with a new machine and compile new with no rc file, the Libertinus font is not loaded. You have to do a "reset" to get the options to use the font.

**To Reproduce**
1. Start with a brand new computer that has never seen fc21 before
2. Compile and install
3. Start game

**Expected behavior**
We want our selected font to be first class and always show.

**Screenshots**
This is me trying Fedora for the first time on a VM.

![image](https://user-images.githubusercontent.com/23037938/230232817-f48f7901-d748-4aa9-9996-fa4dbcf1b749.png)

**Platform and version (please complete the following information):**
 - OS: Linux (assume others impacted)
 - Freeciv21 version: v3.1-dev
 - Ruleset/Longturn game (if applicable): n/a

**Additional context**
none

lmoureaux:
May be related to #1898

jwrober:
Also may be a duplicate of #1036

lmoureaux:
Cannot reproduce on Ubuntu with current master (I was only able to get rid of Libertinus once I completely removed it from my system).

lmoureaux:
Is this still a bug? It degrades the UX enough that I'd like a fix for 3.1.

jwrober:
Yes I think so. Good call.

lmoureaux:
Couldn't reproduce with the 3.1-rc.2 snap.

jwrober:
I'm going to create some new test vm's - windows and unix and test our packages. If you can't repro from snap maybe we fixed this already and didn't realize it.

jwrober:
works as expected on a brand new win 11 vm w/ the rc2 package. the user id is also found for the network dialog

jwrober:
With some extensive testing I'm going to call this closed. 3.1-rc2 has this fixed in it.

PR #1965, opened 2023-05-27
Don't configure options twice

lmoureaux:
Options are configured in client_main, no need to repeat this in ui_main. load_fonts is also called as needed from configure_fonts, so no need to call it from ui_main either.

Found while investigating #1899.


jwrober (approved):
LGTM

PR #1966, opened 2023-05-28
Prevent the tile info tooltip from getting too wide

### LINKED ISSUE #1901: Tile info popup overflows the window when city has too many visible improvements
**Describe the bug**
If a city has many visible improvements (wonders or things like city walls, coastal defense, etc.) and you middle-click it to open the tile information popup, it overflows the window. This makes any tile information unreadable since the popup gets centered, so the information is pushed beyond the left window border.

**To Reproduce**
1. Build a city
2. Build many visible improvements (wonders or things like city walls, coastal defense, etc.) in it
3. Middle-click on the city to open the tile information popup
4. See that it overflows the window

**Expected behavior**
The list of visible buildings should wrap when it becomes too large for the window. Ideally this would also not take the complete window width, but reasonable percentage of it (say, max 50% of window width) so it stays readable and isn't just a super long noodle.

**Screenshots**
Confidential, sent to a place where the devs can see them

 - OS: Windows 10
 - Freeciv21 version: 3.0 rc3 (almost certainly present in newest `stable` and `master` as well, there have been no changes to this popup for a long time)
 - Ruleset/Longturn game (if applicable): not applicable

**Additional context**
No additional context

lmoureaux:
Also refactor the widget slightly to improve code quality and rely more on Qt.

Closes #1901.

### DIFF
--- a/client/views/view_map.cpp
+++ b/client/views/view_map.cpp
@@ -680,11 +680,13 @@ bool mapview_is_frozen() { return (0 < mapview_frozen_level); }
 /**
    Constructor for info_tile
  */
-info_tile::info_tile(struct tile *ptile, QWidget *parent) : QLabel(parent)
+info_tile::info_tile(struct tile *ptile, QWidget *parent)
+    : QLabel(parent), itile(ptile)
 {
-  setParent(parent);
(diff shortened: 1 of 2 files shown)


jwrober (approved):
This is a great patch. Like the wrap now.

Issue #1050, opened 2022-06-25
Fix issues with #871


pranavtifr:
PR #871 was rolled back. Fix issues with #871 and merge again.

PR #1967, opened 2023-05-28
Don't wait forever when trying to connect

lmoureaux:
This is one of the failure modes of bad connections. Fail to connect after 30s (Qt default). Related to #450.

To test, try to connect to an unused game port on LT.net.

### DIFF
--- a/client/clinet.cpp
+++ b/client/clinet.cpp
@@ -132,8 +132,8 @@ static int try_to_connect(const QUrl &url, char *errbuf, int errbufsize)
   }
 
   client.conn.sock->connectToHost(url.host(), url.port());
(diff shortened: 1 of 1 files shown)


jwrober (changes requested):
30s is a very long time to wait with no feedback. Can we add a counter or something to the very bottom of the dialog with a connection counter/countdown? Some visual indicator that the client is trying to connect.

lmoureaux:
This would require going async (#1050), but refactoring the connection code wouldn't be easy.
Could also reduce the timeout.

jwrober:
How about we reduce to 15s?

lmoureaux:
Finally came back to this and decided that 10s should be enough for everyone.

jwrober (approved):

PR #1978, opened 2023-06-12
Ensure that the help dialog doesn't get bigger than the screen it is on

lmoureaux:
* When restoring, cap the size to the screen's available area
* When showing for the first time, use the current screen instead of the first one

To backport or not to backport?


jwrober:
Yea this feels like a good backport candidate.

jwrober (approved):

PR #1979, opened 2023-06-12
Don't give a free tech when inciting a city

### LINKED ISSUE #1944: Inciting a city always gives a tech
**Describe the bug**

It's a LT patch we apparently forgot to port. Inciting a city unconditionally grants a tech.

**To Reproduce**
1. Be behind in tech (/take an AI if you can't)
2. Incite a city
3. Get a tech

**Expected behavior**

No tech is granted, this is way too powerful.

 - OS: any
 - Freeciv21 version: 3.0, 3.1alpha1
 - Ruleset/Longturn game (if applicable): all

**Additional context**

Simply remove this line and the comment above:
https://github.com/longturn/freeciv21/blob/cef4ff5cec1c732758ec1b0599bcde7cc9d9f80b/server/diplomats.cpp#L1129

lmoureaux:
This makes inciting way too powerful, allowing "tech trading" in games where it ought to be disabled.

Closes #1944.

### DIFF
--- a/server/diplomats.cpp
+++ b/server/diplomats.cpp
@@ -1125,9 +1125,6 @@ bool diplomat_incite(struct player *pplayer, struct unit *pdiplomat,
   pcity->shield_stock = 0;
   nullify_prechange_production(pcity);
 
-  // You get a technology advance, too!
-  steal_a_tech(pplayer, cplayer, A_UNSET);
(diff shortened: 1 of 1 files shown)


lmoureaux:
I think this should be backported.

PR #1991, opened 2023-07-01
Change the default background color for Classic

### LINKED ISSUE #1981: So... much... white...
image

lmoureaux:
Use a yellowish color instead of gray. This works better with other widgets.

Closes #1981.

No backport needed, I don't think the issue is present in 3.0.

### DIFF
--- a/data/themes/gui-qt/Classic/resource.qss
+++ b/data/themes/gui-qt/Classic/resource.qss
@@ -32,22 +32,18 @@ QMainWindow, minimap_panel {
 /* Help dialog tile output 3 labels */
 *[foodlab="true"] {
   background:#00ee01;
-  color: black;
 }
 
 *[shieldlab="true"] {
   background: #e4e4e4;
-  color: black;
 }
 
 *[ecolab="true"] {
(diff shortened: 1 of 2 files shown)


jwrober:
The tagged issue only talks about the city dialog, which this patch does correct. I'm wondering however if we should also look at the Nations view. Its all white too, while Units, Cities and Econ have the "classic" color. The background and wireframe of Messages is black (or some other really dark color). Wonder if the frame can be left alone and then add the "classic" color as the background there as well.

lmoureaux:
Nations looks alright for me...
image

lmoureaux:
@jwrober ping :wink:

jwrober:
Oops. Let me re-download and look at this again.

jwrober:
Hm...
image

jwrober:
image

lmoureaux:
I see the same now. Had a stall install around.

lmoureaux:
Player list and messages widget fixed.

jwrober (approved):
Latest changes look great.

PR #1995, opened 2023-07-07
Manual Pages

### LINKED ISSUE #1947: Update manual pages
From the fork from legacy freeciv, we have Linux style man pages in `doc/man`. 

**Do you have suggestions?**
Thinking these should be updated and included in the `.deb` package at a min. If macOS supports man pages, can include in the `.dmg` as well.

jwrober:
Closes #1947 

- Adds final page for freeciv21-ruleup
- Add sphinx to the list of packages to install in CI
- Updates the install target to include the manual pages on Linux

### DIFF
--- a/cmake/FreecivDependencies.cmake
+++ b/cmake/FreecivDependencies.cmake
@@ -171,7 +171,7 @@ if (FREECIV_ENABLE_FCMP_CLI OR FREECIV_ENABLE_FCMP_QT)
   find_package(SQLite3 REQUIRED)
 endif()
 
-# Add Spinx as a soft requirement so we can generate man pages and add to installer.
+# Add Sphinx as a soft requirement so we can generate man pages and add to installer.
 find_package(Sphinx QUIET)
 if (SPHINX_FOUND)
   message(STATUS "Sphinx Found.")
--- a/cmake/FreecivDependencies.cmake
+++ b/cmake/FreecivDependencies.cmake
@@ -170,3 +170,9 @@ endif()
 if (FREECIV_ENABLE_FCMP_CLI OR FREECIV_ENABLE_FCMP_QT)
   find_package(SQLite3 REQUIRED)
 endif()
+
+# Add Spinx as a soft requirement so we can generate man pages and add to installer.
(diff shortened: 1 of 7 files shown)


cmake/FreecivDependencies.cmake
```diff
@@ -170,3 +170,9 @@ endif()
 if (FREECIV_ENABLE_FCMP_CLI OR FREECIV_ENABLE_FCMP_QT)
   find_package(SQLite3 REQUIRED)
 endif()
+
+# Add Sphinx as a soft requirement so we can generate man pages and add to installer.
```
lmoureaux, line 174: Add a message when Sphinx is not found, similar to what we do for stack unwinding libraries?

cmake/FreecivInstall.cmake
```diff
@@ -160,6 +160,13 @@ if(UNIX AND NOT APPLE)
     message(STATUS "CMAKE_INSTALL_PREFIX changed from the default to /usr.")
   endif()
 
+  # Generate the manual pages with Sphinx
+  if(SPHINX_FOUND)
+    message(STATUS "Creating manual pages...")
+    execute_process(
+        COMMAND sphinx-build -b man ${CMAKE_SOURCE_DIR}/docs ${CMAKE_BINARY_DIR}/man)
```
lmoureaux, line 167: Does this run at configure time or at build time? I would prefer build time...

jwrober: It seems to run at configure time, even though the code is in the install section. So it really should run at build or install time, but not sure how to tell cmake to do that.

lmoureaux: The CMake interface to execute commands at build time is through [`add_custom_command`](https://cmake.org/cmake/help/latest/command/add_custom_command.html) and [`add_custom_target`](https://cmake.org/cmake/help/latest/command/add_custom_target.html).

jwrober: but... I don't think we need a custom command or target and adding one seems like overkill

PR #1996, opened 2023-07-08
Fix penalizing buy-then-change-production

### LINKED ISSUE #1915: Queued shields are halved for unique units same as changing type
**Describe the bug**
If you build a unique unit - Navy Troops in LTX as an example - and then continue to build Navy Troops in other cities the shields will store as you cannot actually build more than one unique unit. If you learn a tech (Amphibious Warfare) and want to build Marines with those stored shields, the client cuts the shored shields in half at swap same as if you were building a unit and changed to a city improvement.

**NOTE**: This could be a "feature" and not a "bug" per se, depending on how you are trying to exploit the game mechanics.

**To Reproduce**
1. Set many cities to build Navy Troops or another unique unit
2. Finish one of them in a city, but leave all the other cities still trying to complete the unit
3. Learn Amphib or another tech you want to use to build many units via build slots with the stored shields
4. Swap the construction to the new unit type (e.g. Marines)
5. Notice that you lost 50% of the stored shields for those cities that had more shields stored than the cost of a Marines unit. Those cities with fewer stored shields for one unit didn't have the 50% penalty.

**Expected behavior**
This is problematic on a couple of fronts:
1. This is clearly an exploit of the game engine mechanics
2. The "tax" if you want to call it that isn't evenly applied. It only happens if you have more shields than the cost of a single unit stored up.

There are a couple ways we could go about "fixing" this issue:
1. Change the code in a way that if you build a unique unit and the same unit is being built in other cities an "advisor" kicks in at TC and automatically changes construction to the best defensive unit available. This prevents the storage of the shields completely.
2. Enact some kind of unique-unit to regular-unit change in production tax just like we have already if you change prod between classes of items (e.g. you are building a unit and swap to a city improvement). For consistency, keep it at 50% just like the rest.

I personally like the first. This is an exploit we probably should look to eliminate.

**Screenshots**
image
image

vs

image
image

 - OS: N/A - happens in all clients
 - Freeciv21 version: v3.0 and v3.1-dev
 - Ruleset/Longturn game (if applicable): LTX, but don't think this is ruleset dependent

**Additional context**

lmoureaux:
When a player buys a unit that cannot be built then switches to building something else, penalize half of the shields that were bought. This changes the previous overjealous behaviour where half the total amount of shields would be lost.

Closes #1915.

### Testing done

* Playing normally works as expected.
* Added the `Unique` flag to warriors in `classic` and started a game. Bought a couple Warriors before I had any, then changed production to Settlers. Half of the shields I had bought disappeared as expected.

### DIFF
--- a/client/packhand.cpp
+++ b/client/packhand.cpp
@@ -791,7 +791,7 @@ void handle_city_info(const struct packet_city_info *packet)
 
   pcity->airlift = packet->airlift;
   pcity->did_buy = packet->did_buy;
-  pcity->did_buy_production = packet->did_buy_production;
+  pcity->bought_shields = packet->bought_shields;
   pcity->did_sell = packet->did_sell;
   pcity->was_happy = packet->was_happy;
 
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -1828,12 +1828,12 @@ int city_change_production_penalty(const struct city *pcity,
     penalized_shields = pcity->before_change_shields;
   }
 
-  // Penalize 50% if you buy unit last turn but didnt built
-  if ((pcity->did_buy_production
+  // Penalize 50% if you bought a unit but didn't build it
+  if ((pcity->bought_shields > 0
        && ((target->kind == VUT_UTYPE
             && target->value.utype != pcity->changed_from.value.utype)))) {
-    unpenalized_shields = 0;
-    penalized_shields = pcity->before_change_shields;
+    unpenalized_shields -= pcity->bought_shields;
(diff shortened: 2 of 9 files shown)


jwrober:
Does this change the network protocol? I'll do a more careful review tomorrow.

lmoureaux:
Yes, but in a backward compatible way (though I realize I didn't test... connecting to a running game should be enough).

jwrober (approved):
Works for me.

PR #2000, opened 2023-07-12
Add <releases> tag to client metainfo.xml file

jwrober:
No related issue. Will make future flatpak updates easier.

### DIFF
--- a/dist/net.longturn.freeciv21.client.metainfo.xml
+++ b/dist/net.longturn.freeciv21.client.metainfo.xml
@@ -37,4 +37,7 @@
     <content_rating type="oars-1.0">
       <content_attribute id="social-chat">mild</content_attribute>
     </content_rating>
+    <releases>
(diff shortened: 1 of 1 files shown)


lmoureaux (changes requested):
Can we generate this automatically so we don't need to remember to update it for every release?
I guess we should set it consistently in all metainfo files, not only the client one.

lmoureaux:
Oh, and since there's a validation tool for Flatpak, it would be good to run it in the CI (can be a later project though).

jwrober:
Note this is for `stable`. For now I think leaving it manual in stable branch is the better approach. I don't want to upset the apple cart so to speak. I want to automate it in all the files w/ CI checks in develop/master branch where we have plenty of time to work out any bugs.

lmoureaux (approved):
Sorry had missed the target branch!

PR #2004, opened 2023-07-19
Distribution Updates

### LINKED ISSUE #1997: Re-evaluate how we do .desktop and .metainfo.xml files
As part of the creation of a flatpak package for fc21, it came up during review of our [PR](https://github.com/flathub/flathub/pull/4315) that we are not following established norms as it relates to the configuration of our `.desktop` and `.metainfo.xml` files.

1. We don't have a top level ID for the game - e.g. `net.longturn.freeciv21`. We have `net.longturn.freeciv21.client` and similar for `.server`, `.ruledit`, and `.modpack-qt`.
2. We hardcode a path to the icons that live inside of the fc21 share directory. Seems like the standard is to place them in `share/icons/hicolor/[size]/apps`

To get the flatpack to work (not approved yet), I had to do some serious hacking on the files to get the to conform. 

I think we need to re-evaluate how we are creating and installing these files.

1. Do we create a `.directory` file for Freeciv21 so menus would show Games/Freeciv21/... and the 4 icons?
2. Start shipping the `data/icons` directory in our packages/distribution and with proper file naming put the files into `share/icons/hicolor/[size]/apps` directory.
3. Update the `.desktop` files so the game (client) is a top level ID - e.g. `net.longturn.freeciv21` and the rest of the programs are below it - e.g. `net.longturn.freeciv21.server`. Alternately we say the client is `.net.longturn.freeciv21.freeciv21`, but I think this is overkill.
4. Update the `.desktop` files to simply refer to a PNG file that now resides in `share/icons/hicolor/[size]/apps`, which means no full path. So need to change how we do `configure_file()`.
5. Update the `.metainfo.xml` files to match what we did in the `.desktop` ones.
6. Change the icon location for the `.metainfo.xml` to be local and not remote

What we have now "works", but isn't really cross-platform/package capable. Better to use best practices and follow established standards.

**Additional context**
None

### LINKED ISSUE #2001: Standardize the long description across all platforms
We have a standard long description that is spread across many locations:

1. Repo Readme
2. Manual Pages
3. metainfo files
4. snap package
5. flathub package

**Do you have suggestions?**
Standardize it so it is the same across all places that it will be seen by users.

### LINKED ISSUE #2003: Icons are everywhere...
We have icons in:

- /data
- /data/icons
- /data/misc
- /dist

Consolidate all icons to live in `/data/icons`. We also should look at getting rid of the low res 16x16 and 32x32 icons as well. Might look at creating high-res 512x512

None

**Additional context**
Looked at this more closely as part of evaluating work on #1997

jwrober:
Closes #1997 
Closes #2001 
Closes #2003 

Taking an opportunity to clean up where we store icons as well as how we do the desktop and metainfo files.


lmoureaux (commented):
As we briefly discussed on Discord, I think it would be good to have a list documenting all the places where the description is pasted. A reminder comment can then be added next to each instance to avoid them getting out-of-sync.

Since files were moved around, I'd like to make sure that all installers and packages get the correct icons. Windows and Mac OS installers can be downloaded from the CI, but snap/flatpak/deb should also be confirmed to be working.

jwrober:
I can confirm that the deb package is fine. We currently don't support snap and flatpak on development branch, but I can look to see how they are impacted to be sure noting is horribly broken. I also agree on the list of places the standard description is placed so we can keep changes in sync for any future updates. More commits coming later today.

jwrober:
Snap and windows work well too. Flatpak is still based on `stable` so won't be impacted here.

PR #2008, opened 2023-07-22
Improve texts in the research view

lmoureaux:
Improve the following texts in the research view header following [a comment by Blauwal on Discord](https://discord.com/channels/378908274113904641/1106327029223272518/1132109288521867346):
* The bulb production details (top left): add a missing newline when tech upkeep is enabled
* The current status (top right in the progress bar): make it work better when tech leak and tech upkeep are used together
* The number of bulbs/turns required to reach the current goal (bottom right): make it less confusing by taking the real cost of the current tech into account

Some small code improvements here and there as well.

I think the risks of backporting this outweigh the benefits.

### DIFF
--- a/client/text.cpp
+++ b/client/text.cpp
@@ -750,10 +750,14 @@ int get_bulbs_per_turn(int *pours, bool *pteam, int *ptheirs)
  */
 int turns_to_research_done(const struct research *presearch, int per_turn)
 {
-  if (per_turn > 0) {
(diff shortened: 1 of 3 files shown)


jwrober:
Is there a savegame or something I can use to test this?

lmoureaux:
Unfortunately not, I was testing by observing Blauwal in LT79.

jwrober (approved):
Code looks like it works. I personally would like to actually see it work for real before we backport.

PR #2009, opened 2023-07-24
(stable) Dist updates

jwrober:
Follows #2004, but for stable branch.


lmoureaux:
Will our flathub configuration need changes after this?

jwrober:
> Will our flathub configuration need changes after this?

Yes, sure will. Will update as soon as this is done. For example, we don't need the patch any longer.

PR #2012, opened 2023-07-25
Ensure icon png available for win32

jwrober:
Found a regression from #2004. Fixing here. No related issue.


lmoureaux (approved):
Why does Windows need a jpg in the first place?

jwrober:
For some reason the client uses `misc/icons.spec` to get the main icon on the program. I found it on my main Linux side, but was playing around on my Msys2 machine and noticed the same. If the icon file is missing, the client displays as a red square.

Looks like I didn't do something right. Msys2 install not working. Will push a fix.

PR #2014, opened 2023-07-26
Use scriptlet to get current version for snapcraft.yaml

### LINKED ISSUE #1735: Integrate snapcraft.yaml with cmake configure_file()
New `snapcraft.yaml` file is hard coded to a version tag. We want to have the file use the cmake `configure_file()` function going forward.

Rename `snapcraft.yaml` as `snapcraft.yaml.in` and then update our `CMakeLists.txt` file to configure it correctly in the `build` directory.

**Additional context**
Right now, `snapcraft.yaml` will only work on the `stable` branch. We need to wait until the snap team gets Qt support for Ubuntu 22.04 via `core22`. We have a higher cmake version requirement in `master` branch that breaks the snap capability.

jwrober:
Closes #1735

Part of this might be good for `stable`

### DIFF
--- a/dist/snapcraft.yaml
+++ b/dist/snapcraft.yaml
@@ -10,6 +10,7 @@
 #   - mkdir -p build/snap/local
 #   - cp dist/snapcraft.yaml build/snap
 #   - cp data/icons/128x128/freeciv21-client.png build/snap/local
+#   - cp cmake/AutoRevision.txt build/snap/local
 #   - cd build
 #   - snapcraft
 #   - sudo snap install --devmode ./freeciv21_*.snap
@@ -27,10 +28,7 @@ description: |
 
   The code is maintained by the team over at Longturn.net and is based on the QT framework. The game supports both hex and square tiles and is easily modified to create custom rules.
 
-# TODO: At some point, will want to integrate into cmake process and configure_file()
-version: '3.1-dev'
-grade: devel
-
+adopt-info: freeciv21
 license: GPL-3.0-or-later
 website: https://longturn.net
 contact: longturn.net@gmail.com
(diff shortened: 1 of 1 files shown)


dist/snapcraft.yaml
```diff
@@ -48,10 +46,13 @@ icon: snap/local/freeciv21-client.png
 
 parts:
   freeciv21:
+    override-pull: |
+      snapcraftctl pull
+      snapcraftctl set-version `sed -n '2p' snap/local/AutoRevision.txt`
+      snapcraftctl set-grade "devel"
```
lmoureaux, line 52: Shouldn't this also be computed dynamically?

jwrober: Suggestions on how to do that?

lmoureaux: The three possible values are `devel`, `candidate`, and `stable`. What about setting `candidate` for rc releases and `stable` for stable releases?

jwrober: I guess I could write an `if` statement that uses `awk` or something to tease out part of the version. `alpah` then gives `devel`, `rc` gives `candidate` and finally the absence of those gives `stable`.

lmoureaux:
Anything I can help with to move this PR forward?

jwrober:
I was looking at it yesterday after reviewing your other PR. I will have some time this week to poke on it. Any suggestions you may have are always welcome.

lmoureaux (approved):

lmoureaux:
> Part of this might be good for stable

We can add to the backport list if you think this will rebase fine. Else we need a second PR

PR #2018, opened 2023-08-01
Remove spurious qCritical()

lmoureaux:
Spotted by soundnfury.


jwrober (approved):
I saw those messages too. Good catch.

PR #2022, opened 2023-08-22
Make units on cities visible when city drawing is disabled

rsubtil:
Suggested on Discord. This will draw units stationed in cities when drawing cities is disabled in the settings.

### DIFF
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -791,8 +791,9 @@ unit *find_visible_unit(const ::tile *ptile)
     }
   }
 
-  // If a city is here, return nothing (unit hidden by city).
-  if (tile_city(ptile)) {
+  // If a city is here (and UI is drawing them), return nothing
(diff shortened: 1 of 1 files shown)


lmoureaux (changes requested):
I think this is a great addition, thanks!
It seems you inadvertently changed the newlines in the Windows batch script to start the client, could you revert to the original?

rsubtil:
Oops, yeah sorry, didn't notice it! :sweat_smile:

lmoureaux (approved):

lmoureaux:
Feel free to submit a PR if you want to add yourself to [AUTHORS](https://github.com/longturn/freeciv21/blob/master/AUTHORS), we try to document copyright holders there

rsubtil:
Thanks! I might actually make some more significant contributions due to my existing work, so in the future I can do that.

PR #2025, opened 2023-08-27
Fix disembarking without losing MP through adjacent transport

lmoureaux:
Say we have two transports A and B next to the coast C, with a land
unit in A. Consider the following orders for the land unit:

1. Unload from A to B->tile
2. Move from B->tile to C

This sequence should not succeed because the unit should still be loaded
after the first step. However, in practice the unit was not loaded after
step 1, resulting in an invalid state that confused ORDER_MOVED. Do the
same when bumping units after a transport is killed.

The fix to that is obviously to load the unit after step 1. There is a
catch though: this is (and has always been) done without considering
action enablers. It might be possible to exploit this to bypass loading
rules.

Path finding support is also included in this PR.
Backport candidate.

### DIFF
--- a/common/map.cpp
+++ b/common/map.cpp
@@ -759,19 +759,9 @@ int tile_move_cost_ptrs(const struct civ_map *nmap, const struct unit *punit,
 
   } else if (!is_native_tile_to_class(pclass, t2)
              || !is_native_tile_to_class(pclass, t1)) {
-    if (tile_city(t1) == nullptr) {
-      /* Loading to/disembarking from transport. */
-
-      // UTYF_IGTER units get move benefit.
-      return (utype_has_flag(punittype, UTYF_IGTER) ? MOVE_COST_IGTER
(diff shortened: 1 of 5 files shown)


jwrober:
Does this bug only impact combat units, or would it also work on engineers or other non-combat units? I can't recreate on an unpatched install.

lmoureaux:
All units are affected as long as they have at least 3 MP (1 for ship hopping, 1 to disembark, 1 to attack). I used Mechs in my tests.

jwrober:
Ok I recreated with a Mech too.

jwrober (approved):
This does resolve the issue. I am a little concerned about the FIXME added to `path_finder.cpp`

lmoureaux:
> I am a little concerned about the FIXME added to path_finder.cpp

It might be possible to move from a ship into a transport plane or something like this that would normally be forbidden.

PR #2034, opened 2023-09-09
Don't overwrite the "square" tileset option on startup

lmoureaux:
And actually never overwrite tileset options when loading a tileset. Updating the tileset options should only be done when the user explicitly changes them.


jwrober (approved):
WFM

PR #2036, opened 2023-09-20
fcdb.rst update

jwrober:
Update the fcdb.rst file based on recent Discord conversation where the instructions were not clear for a new user. Also update the CPackConfig.cmake file to specifically ask for lua5.3 package.

Recommend we backport the commit to CPackConfig.cmake to stable branch.


docs/Manuals/Server/fcdb.rst
```diff
@@ -46,9 +46,11 @@ storing passwords with MD5 hashes.
 
 .. code-block:: sh
 
-    $ sudo apt install luarocks
-    $ sudo luarocks --lua-version 5.3 install md5
-    $ sudo luarocks --lua-version 5.3 install luasql-sqlite3
+    $ sudo apt install luarocks lua5.3 liblua5.3-dev libsqlite3-dev
+    $ lua -v  # output should be 5.3.x
```
lmoureaux, line 50: This will not necessarily be true if the user changed the default Lua version via `update-alternatives` (do `update-alternatives --display lua-interpreter`). Lua 5.3 can be invoked using `lua5.3`, but then the check becomes tautological.
I would suggest to drop this check: Freeciv21 is currently always compiled against Lua 5.3.

cmake/CPackConfig.cmake
```diff
@@ -108,7 +108,7 @@ if(UNIX AND NOT APPLE)
   set(CPACK_COMPONENTS_GROUPING ALL_COMPONENTS_IN_ONE)
   set(CPACK_DEB_COMPONENT_INSTALL "ON")
   set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS "ON")
-  set(CPACK_DEBIAN_PACKAGE_DEPENDS "fonts-linuxlibertine (>= 5.3.0)")
+  set(CPACK_DEBIAN_PACKAGE_DEPENDS "fonts-linuxlibertine (>= 5.3.0), lua (>= 5.3.0)")
```
lmoureaux, line 111: Our package doesn't depend on the Lua interpreter. If installing it to silence `luarocks` warnings, just have the user install it alongside `luarocks` in the instructions.

docs/Manuals/Server/fcdb.rst
```diff
@@ -35,7 +35,7 @@ Freeciv21 server. A good default location is the user's :file:`$HOME` directory.
 
     [fcdb]
     backend="sqlite"
-    database="/my/path/to/freeciv21.sqlite"
+    database="/home/user/freeciv21.sqlite"
```
lmoureaux, line 38: I think we should already highlight that `user` needs to be changed.

docs/Manuals/Server/fcdb.rst
```diff
@@ -75,9 +78,12 @@ command:
     /fcdb lua sqlite_createdb()
 
 
-Now you can create some users by connecting with the client. Due to the :code:`--Newusers` flag, when you
-connect with the client with a previously unknown username, the server will prompt for a password and save the
-new account to the database.
+You should now see a :file:`/home/user/freeciv21.sqlite` file as per the same path given in the
+:file:`fc_auth.conf` file.
```
lmoureaux, line 82: Oh cool it creates it automagically! :+1:

lmoureaux (changes requested):
My main concern is the introduction of a new dependency to a Lua interpreter, which is not needed for anyone.

lmoureaux (approved):

Issue #2041, opened 2023-10-05
Large cities lose ability to adjust specialists


jwrober:
**Describe the bug**
If you have a very large city (e.g. 40+) you cannot adjust specialists on the city view. The citizen icons are all stacked/smashed together and you cannot click on an individual citizen icon to adjust specialists (taxmen, scientist, entertainer).

**To Reproduce**
1. Open the city dialog of a really big city (40+)
2. Take a citizen off the working area, creating an entertainer
3. Try to click the entertainer to change to an taxman
4. See error/issue

**Expected behavior**
The icons should be easy to click on and make adjustments.

This is a bug, but also an opportunity to enhance the citizen icons.

**Screenshots**

**Platform and version (please complete the following information):**
 - OS: N/A
 - Freeciv21 version: both v3.0+ and 3.1-alphas are impacted
 - Ruleset/Longturn game (if applicable): N/A

**Additional context**
Discord conversation: https://discord.com/channels/378908274113904641/402213035865014272/1159279524698140703

jwrober:
Been thinking about this some more lately. I think an easy solution is to break the citizen icons into rows of 10. When the city grows to 11, a second row is created and pushes the rest of the items on the panel down a bit. Keep adding a row at every +10. Plenty of real estate on the new city dialog to do this, even a gigantic size 100 city would only have 10 rows of citizen icons.  Could even go with 15 or maybe 20, but 10 seems like an easy split point.

lmoureaux:
I was thinking in a similar direction, simply breaking the bar when it would become too long for the available width. I find your proposal better ergonomics since it would facilitate counting citizens. I'd go for a break at 20 though, to avoid cities with 4 or 5 rows in LTT/X.

On thing to consider is how citizens are distributed in this space. I think we should still have happy citizens on the left and specialists on the right, like so:
```
HHHHCCCCCCCUUUUUUSSS
HHHCCCCCCC
```
A downside of this layout is that counting happy and unhappy citizens becomes more difficult. One could also lay it out with the second line centered:
```
HHHHHHCCCCCCCUUUUSSS
     HCCCCCCCUU
```
It's more readable but will look a bit strange with the rest of the column left-aligned.

jwrober:
[...]

PR #2042, opened 2023-10-05
Fix changing specialists for city sizes >= 40

lmoureaux:
Fix suggested by daavko.
See #2041.


daavko (commented):
Can't really test it at this time, but the code change looks good to me

jwrober (approved):

PR #2043, opened 2023-10-10
Update american.ruleset

Rorr4:
Added more leaders and more nations to civilwar_nations

### DIFF
--- a/data/nation/american.ruleset
+++ b/data/nation/american.ruleset
@@ -17,8 +17,7 @@ legend=_("The United States of America achieved its independence from\
 leaders = {
  "name",                        "sex"
  "George Washington",           "Male"
- "Martha Washington",    	"Female"
- "Alexander Hamilton",          "Male"
+ "Martha Washington",    	      "Female"
  "Benjamin Franklin",           "Male"
  "John Adams",                  "Male"
  "Thomas Jefferson",            "Male"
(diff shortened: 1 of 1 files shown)


data/nation/american.ruleset
```diff
@@ -17,6 +17,7 @@ legend=_("The United States of America achieved its independence from\
 leaders = {
  "name",                        "sex"
  "George Washington",           "Male"
+ "Martha Washington",    	"Female"
```
lmoureaux, line 20: Can we try to keep the nice alignment intact?

lmoureaux (commented):
Thanks for the update! The addition of a few more women to the list is much appreciated! I have one small coding style comment (in the next section) and a few suggestions to be discussed:

In order to increase the proportion of women, we could consider removing some less known presidents. From my European point of view, Hamilton/Monroe/Jackson--Taylor/Andrew Johnson/Cleveland/McKinley could be candidates for deletion.
We could also include a few important female senators or representatives, maybe Jeannette Rankin, Nancy Kassebaum, Kamala Harris...

data/nation/american.ruleset
Rorr4: I would say remove Hamilton as he was never a president.

jwrober:
See #2044

PR #2051, opened 2023-11-05
Add Docker and Docker Compose support for the server

nat-418:


lmoureaux:
Hi nat! My first question is about your motivation. I'm not sure what the use case is: do you want this just for testing, or should this deployment handle saves, plugging in additional rulesets, doing user authentication? Should the client and modpack installer also work?

nat-418:
> Hi nat! My first question is about your motivation. I'm not sure what the use case is: do you want this just for testing, or should this deployment handle saves, plugging in additional rulesets, doing user authentication? Should the client and modpack installer also work?

I am thinking about self-hosting a server for me and some friends and noticed there was no Docker setup, and I figure that is the way most people are self-hosting their services these days so I decided to share my work on one. If there's anything you recommend I enable or configure let me know. I think I could pretty easily add a volume for saves and rulesets. Not sure  exactly how the auth piece works if you want to explain. Basically, the container is just a simple Debian setup per what I saw in the README.

lmoureaux:
For the usage you describe I think it's important to have persistent saves and [authentication support](https://longturn.readthedocs.io/en/latest/Manuals/Server/fcdb.html). (I suppose it's actually running on k8s? I don't think docker-compose is a good deployment method, would rather create a systemd service.) The ability to mount custom rulesets into the image would be nice but I don't see this as needed for a first iteration.

To facilitate maintenance, could we avoid hard-coding the server version?

Also I think the files should reside in `dist/docker` to avoid cluttering the repo root.

lmoureaux:
Closing as inactive - feel free to reopen if you come back to it

PR #2052, opened 2023-11-21
Stop disbanding units under UWT when bouncing them

lmoureaux:
When a unit affected by unit wait time was bounced, the generated move order was rejected by the move handling code. As a consequence, the unit ended up being disbanded.

This affected units doing terraforming work (e.g. turning land to ocean). This forced players to issue the terraforming orders at least UWT seconds before TC, which forced them to connect at specific times of the day. Since one of the goals of UWT is exactly to avoid this constraint around TC, adding a similar constraint around TC - UWT is the opposite of good.

### DIFF
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -1375,7 +1375,14 @@ void bounce_unit(struct unit *punit, bool verbose, bounce_reason reason,
         packet.orders[i] = steps[i].order;
       }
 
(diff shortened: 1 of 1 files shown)


jwrober (approved):
works for me

lmoureaux:
Also added to backport list, I think it's important to have in the next 3.0 release

PR #2055, opened 2023-11-22
Fix map sharing

### LINKED ISSUE #2045: Map data transfer in need of thorough review
I think that the implementation of one-time transfer of map data and of shared vision should be thoroughly reviewed in its entirety. There seem to be numerous bugs. I’ve seen several in the past, from what I remember; in particular there was this one where with shared vision you were shown wrong assignments of territory to tiles (this particular one might or might not have been fixed meanwhile). Just now I saw another map sharing bug in action: you give your world map to someone, and the city data is from various earlier turns apparently (drastically too low city sizes, drastically different city sizes of cities built around the same time, wrong assignment of nations to cities).

How to reproduce these bugs? Well, play a lot of Freeciv with lots of transfers of world and sea maps as well as shared vision and look very carefully at the map data you get. That said, since there seem to be multiple bugs and these bugs seem to often have non-obvious reasons, it would be better, I think, to really thoroughly inspect the code that implements all this sharing of map data (one-shot and continuous sharing).

What’s the expected behavior? Well, that this whole thing works correctly. :slightly_smiling_face: In particular, the *newest* data available should always be shown. So don’t have old data fetched from another player and have it overwrite newer data from another player or newer data that you already possess.

Regarding platforms, I guess this is a platform-independent problem, but I can tell you that I use `freeciv21-client 3.0.886883.1-patch` via the Snap package on Ubuntu 20.04 with an Intel processor. The ruleset I’m *currently* using is LTX as used in LT79.

lmoureaux:
Sharing maps between players was fundamentally broken. The server records the last turn in which a player saw a tile. This information is only updated when the player stops seeing a tile, and not when it just keeps seeing it because the tile is within its vision range.

The following map sharing bugs existed:
* Tiles seen by the giving player were not sent if the receiving player had last viewed them after the giving player;
* Cities were only updated when destroyed. Renamed or growing cities were not updated.

This commit fixes both bugs. The core logic in `really_give_tile_info_from_player_to_player` was completely overhauled to make it easier to follow and more correct. In addition, `player_tile.site` was turned to a `unique_ptr` to facilitate bookkeeping.

Closes #2045

Testing: Before the patch:
* Watch the occupied city indicator in a city. Move units in and out while giving your map to another player. See the indicator not being updated for the other player.
* Same with city renaming.
* If you are patient enough, repeat with city growth, terrain alterations, etc.

After the patch, the other player should always get the information at the time of signing the treaty.

Not suggesting to backport since the bugs had been there for a very long time.

### DIFF
--- a/common/vision.cpp
+++ b/common/vision.cpp
@@ -62,11 +62,6 @@ bool vision_reveal_tiles(struct vision *vision, bool reveal_tiles)
 }
 
 /**
-   Frees vision site structure.
- */
-void vision_site_destroy(struct vision_site *psite) { delete[] psite; }
-
-/**
    Returns the basic structure.
  */
 struct vision_site *vision_site_new(int identity, struct tile *location,
--- a/common/vision.h
+++ b/common/vision.h
@@ -130,7 +130,6 @@ struct vision_site {
 };
 
 #define vision_site_owner(v) ((v)->owner)
-void vision_site_destroy(struct vision_site *psite);
 struct vision_site *vision_site_new(int identity, struct tile *location,
                                     struct player *owner);
 struct vision_site *vision_site_new_from_city(const struct city *pcity);
--- a/server/citytools.cpp
+++ b/server/citytools.cpp
@@ -2128,7 +2128,7 @@ bool send_city_suppression(bool now)
 static void package_dumb_city(struct player *pplayer, struct tile *ptile,
                               struct packet_city_short_info *packet)
 {
-  struct vision_site *pdcity = map_get_player_city(ptile, pplayer);
+  const vision_site *pdcity = map_get_player_city(ptile, pplayer);
 
   fc_assert_ret(pdcity != nullptr);
   packet->id = pdcity->identity;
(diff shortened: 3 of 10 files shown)


jeltsch:
> Tiles seen by the giving player were not sent if the receiving player had last viewed them after the giving player

Is this really a bug? I would expect that newer information doesn’t get overwritten with older information.

lmoureaux:
> > Tiles seen by the giving player were not sent if the receiving player had last viewed them after the giving player
> 
> Is this really a bug? I would expect that newer information doesn’t get overwritten with older information.

"Seen" tiles are the tiles currently visible by a player (as opposed to tiles that are merely "known")

jwrober:
Looks like the PR is failing the clang-format checker.

lmoureaux:
Uh I see, the usual case not caught by `git-clang-format`

lmoureaux:
clang-format fixed

jeltsch:
> > > Tiles seen by the giving player were not sent if the receiving player had last viewed them after the giving player
> > 
> > Is this really a bug? I would expect that newer information doesn’t get overwritten with older information.
> 
> "Seen" tiles are the tiles currently visible by a player (as opposed to tiles that are merely "known")

Then I don’t understand this even more. If the seen tiles are currently visible to the player, they are viewed *now. How can then the other player have them last viewed at a later time?

lmoureaux:
The two players are usually not sharing vision, as this is related to the treaty clauses to share one's map

lmoureaux:
Sorry, I missed the question in the second part. As an optimisation (I suppose), the "last viewed" info is not updated at every turn change. It gets updated only when vision is lost.

jeltsch:
> Sorry, I missed the question in the second part. As an optimisation (I suppose), the "last viewed" info is not updated at every turn change. It gets updated only when vision is lost.

I must say that finally I’m utterly confused. I don’t know whether I’d like to help here further. This stuff seems so convoluted that I don’t really have further ambition to try to understand it. Sadly, I’m by no means convinced that the fix in this PR is actually suitable.

jeltsch:
Looking again at what “last viewed” means, it starts to make sense to me why the first point mentioned in the PR description could be a bug. Still, parts of the map where the receiving player has currently vision should not be sent. Is this taken into account by the updated code?

lmoureaux:
> parts of the map where the receiving player has currently vision should not be sent. Is this taken into account by the updated code?

Yes, this is the first check:
```cpp
  // No need to transfer if pdest can see the tile
  if (map_is_known_and_seen(ptile, pdest, V_MAIN)) {
    return;
  }
```
I'd be interested if you have ideas for a simpler implementation

jeltsch:
> I'd be interested if you have ideas for a simpler implementation

At least not spontaneously. :slightly_smiling_face:

lmoureaux:
@jwrober can you review then?

jwrober:
> @jwrober can you review then?

Yep. Prolly tomorrow. Will need to setup a test.

lmoureaux:
Thanks!

jwrober (approved):
Works for me. Nice patch!

PR #2057, opened 2023-11-22
Introduce a multi slider widget and use it to choose tax rates

lmoureaux:
I would like to have a generic version of fc_double_edge (the widget to choose tax rates) available. The general idea is that there is a fixed number of items and the user can distribute them across multiple categories. This could be useful for specialists in big cities. In order to level up the field a bit, I intend to make the new widget usable with the keyboard only.

This PR implements the generic widget and uses it in the tax rates dialog. Both keyboard and mouse navigation should work properly. The widget was developed with NightStalker and I quickly checked that it's not doing anything crazy under Classic.

If/when this gets merged, documentation screenshots will need an update.

### DIFF
--- a/client/widgets/multi_slider.cpp
+++ b/client/widgets/multi_slider.cpp
@@ -14,21 +14,38 @@
 #include <qabstractslider.h>
 
 namespace {
-  // Handle metrics
+/// Widget dimensions, in logical pixels
+namespace metrics {
+  /// Gap between the icons and the focus indicator
+  const double focus_bar_gap = 1;
+  /// Height of the focus indicator
+  const double focus_bar_height = 2;
+  /// Gap at the top of the handle bar
   const double handle_bar_gap = 2;
+  /// Width of the handle bar
   const double handle_bar_width = 4;
+  /// Gap between the icons and the handle (circle)
   const double handle_gap = 1;
+  /// Radius of the handle
   const double handle_radius = 8;
+  /// Radius of the small disk inside the handle
   const double handle_indicator_radius = 4;
+  /// Radius of the small disk inside the handle, when the handle is active
   const double handle_active_indicator_radius = handle_indicator_radius + 1;
-  const double handle_extra_height = handle_gap + handle_radius * 2;
+  /// Height added to the icon height by control elements (handle etc)
+  const double extra_height = std::max(focus_bar_gap + focus_bar_height,
+                                       handle_gap + handle_radius * 2);
+} // namespace metrics
 } // anonymous namespace
 
 namespace freeciv {
 
-multi_slider::multi_slider(QWidget *parent): QAbstractSlider(parent)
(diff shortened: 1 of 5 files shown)


client/widgets/multi_slider.cpp
```diff
@@ -0,0 +1,630 @@
...
+const double handle_indicator_radius = 4;
+/// Radius of the small disk inside the handle, when the handle is active
+const double handle_active_indicator_radius = handle_indicator_radius + 1;
+/// Height added to the icon height by control elements (handle etc)
+const double extra_height = std::max(focus_bar_gap + focus_bar_height,
+                                     handle_gap + handle_radius * 2);
+} // namespace metrics
+} // anonymous namespace
+
+namespace freeciv {
```
jwrober, line 54: When we refactor something why don't we use a diff namespace such as fc21 or freeciv21?

jwrober (approved):
This is a great new feature that I can see us re-using all over the UI. I just have the one review question.

Should we also remove the old sickle icons used?

I think we can update docs here too, but also fine with another PR on that "soon" so we don't forget.

client/widgets/multi_slider.cpp
lmoureaux: The `namespace freeciv` is actually new. Could switch to `fc21` but I find it a bit obscure and it's also harder to type for me (I need shift to type numbers). `freeciv21` is probably too long.

lmoureaux:
> Should we also remove the old sickle icons used?

I wasn't sure. I think the new handles are graphically less engaging, but the sickles were also not great. I went with what was easier to code. Better styling support would be good...

client/widgets/multi_slider.cpp
jwrober: Ok learned something new.

jwrober:
> > Should we also remove the old sickle icons used?
> 
> I wasn't sure. I think the new handles are graphically less engaging, but the sickles were also not great. I went with what was easier to code. Better styling support would be good...

The new handles are def not sexy, but they do work well. Do you need code to provide better styling support or is this something we can play with in a theme once merged?

jwrober:
> > Should we also remove the old sickle icons used?
> 
> I wasn't sure. I think the new handles are graphically less engaging, but the sickles were also not great. I went with what was easier to code. Better styling support would be good...

The new handles are def not sexy, but they do work well. Do you need code to provide better styling support or is this something we can play with in a theme once merged?

lmoureaux:
> The new handles are def not sexy, but they do work well. Do you need code to provide better styling support or is this something we can play with in a theme once merged?

Changing them would need more code. Even the colors are hard-coded right now

jwrober:
I added an issue to capture the code for theme support. I do think we should either update this PR with the new images for the client manual or do it very soon in another PR.

lmoureaux:
I would say it should come with the theming PR if it comes in time for 3.1. Let's open an issue and add it to the 3.1 milestone

PR #2060, opened 2023-11-22
Allow reopening investigated cities

### LINKED ISSUE #2059: Save city state when investigating
When investigating a city, the information is lost when the city dialog is closed. It can only be preserved with a screenshot.

As a minimal update, the client should remember what you've seen in the current session.

Have the server remember this info too (even across reloads). It would be nice but it's a lot more work.

**Additional context**
Requested by @ec429 [on Discord](https://discord.com/channels/378908274113904641/912500712833974322/1176957704674226247).

lmoureaux:
The client was already remembering information about investigated cities, but there was no way of seeing it once out of the city screen. Allow reopening the city screen by clicking on the city.

Closes #2059.

Could be backported even though it's technically a new feature, it was so simple...

### DIFF
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -1172,9 +1172,7 @@ void control_mouse_cursor(struct tile *ptile)
     if (nullptr != punit && unit_owner(punit) == client_player()) {
       // Set mouse cursor to select a unit.
       mouse_cursor_type = CURSOR_SELECT;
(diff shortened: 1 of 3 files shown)


jwrober (approved):
WFM, agree to NOT backport as a new feature even though the change turned out to be pretty easy.

PR #2073, opened 2023-12-05
Bump stable client meta file to 3.0.2

jwrober:
Task 30 in the back port. Manually bump the version in the metainfo file.


lmoureaux:
Sorry that I took so long to approve, the release date is already in the past... Guess we should do this right before the release? (ie we need to fix the macOS failures, backport them, then perform basic checks on as many platforms as possible)

jwrober:
I agree. I can push another commit as the last step before we tag the release. Let's get the mac build fixed.

PR #2077, opened 2023-12-17
Fix SDL autodetection

lmoureaux:
Moving to the official SDL2 config resulted in a few variables now following the standard CMake naming conventions. Use the new names when checking whether SDL is present.


jwrober (approved):
WFM, thanks for the quick fix.

PR #2078, opened 2023-12-17
Update release process 

jwrober:
to include managing updates to vcpkgGitCommitId in CI.

Also fix a small typo to the new Workarounds page.

### DIFF
--- a/docs/Contributing/release.rst
+++ b/docs/Contributing/release.rst
@@ -28,6 +28,10 @@ These are the general steps to prepare and finalize a release:
    post to the ``#releases-project`` channel in the ``LT.DEV`` section on the Longturn Discord server. The
    main purpose of the post is to alert the developers of a pending cadence threshold and to gauge if we
    need to delay in any way or if we are good to proceed as normal.
+#. A release manager or another regular contributor will update the ``vcpkgGitCommitId`` value in the CI/CD
+   build file (:file:`.github/workflows/build.yaml`). Grab the Commit ID of the most recent release from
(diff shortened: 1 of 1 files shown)


docs/Contributing/release.rst
```diff
@@ -28,6 +28,10 @@ These are the general steps to prepare and finalize a release:
    post to the ``#releases-project`` channel in the ``LT.DEV`` section on the Longturn Discord server. The
    main purpose of the post is to alert the developers of a pending cadence threshold and to gauge if we
    need to delay in any way or if we are good to proceed as normal.
+#. A release manager or another regular contributor will update the ``vcpkgGitCommitId`` value in the CI/CD
+   build file (:file:`.github/workflows/build.yaml`). Grab the Commit ID of the most recent release from
+   https://github.com/microsoft/vcpkg/releases. Commit and push a PR. Ensure all of the CI/CD Action runners
+   complete successfully.
```
lmoureaux, line 34: This ID is also present in `vcpkg.json`. I think we can also put a version tag there, which would be easier to understand (but needs testing).

jwrober: The doco for the `vcpkg.json` file led me to believe that it isn't something to change often -- https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json

But maybe it makes sense to keep the two values in sync?

lmoureaux: Oh I see, yes I agree that if it's the minimum version it doesn't have to be changed often.

lmoureaux (approved):

PR #2079, opened 2023-12-23
Fix macOS CI again

lmoureaux:
Use the `SDL2_mixerConfig.cmake` provided by SDL2_mixer on macOS instead of our `FindSDL2_mixer.cmake`. Document the find module as being there only to support older library versions.


lmoureaux:
I hope this version is good to merge

lmoureaux:
> WASM still not picking up SDL, but I don't really think its an issue to worry about.

SDL2_mixer is used in a wholly different way on emscripten, so the following confusing output is printed:
```
-- Could NOT find SDL2_mixer (missing: SDL2_mixer_INCLUDE_DIRS) 
-- Building with audio support enabled
```
I can see that it's building `audio_sdl.cpp` so sound support is enabled.

PR #2083, opened 2023-12-24
Implement two broken governor options

### LINKED ISSUE #2058: Newly exposed governor settings not doing anything
**Describe the bug**

Some new governor settings have been exposed in UI in #1841. They do not seem to do anything.

**To Reproduce**
1. Open a moderately-sized city
2. Play with the new governor settings
3. See specialists when they're not allowed, and suboptimal growth behavior when optimization has been requested

**Expected behavior**
All exposed settings work

 - OS: Linux
 - Freeciv21 version: current `master`

lmoureaux:
This PR fixes the governor option to "optimize" food for growth, trying to find as much food as possible without having excess food when growing. The implementation of this feature isn't optimal but I found it easier to fix the current code than to implement something better.

I also implemented the governor option to disallow using specialists. The field for this was implemented in 2003 and maybe it was working back then, but I don't remember it working in any of the versions I played.

Closes #2058.

### DIFF
--- a/common/aicore/cm.cpp
+++ b/common/aicore/cm.cpp
@@ -16,8 +16,6 @@
 
 // utility
 #include "shared.h"
-#include "support.h"
-#include "timing.h"
 
 // common
 #include "city.h"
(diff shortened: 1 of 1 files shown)


jwrober (approved):
LGTM

PR #2085, opened 2023-12-25
Switch to Anarchy when the current government is lost

### LINKED ISSUE #2047: Government shouldn't be chosen randomly when losing a tech
When a player loses a tech that causes them to lose access to a government, the game currently chooses a random available government. This may be surprising behavior to players, considering other situations that cause spontaneous government change would go directly to Anarchy instead.

The game should go to Anarchy instead of randomly choosing.

Leaving it as-is, which is not great. Also considered leaving it up to ruleset author, but that'd be complicated since they'd have to make sure the player can go to the configured government in any situation where they'd lose a tech.

**Additional context**
Discord discussion https://discord.com/channels/378908274113904641/402213035865014272/1165650726400294932

lmoureaux:
When the technology for the current government is lost, the game used to pick some government at random. Be more punitive and always switch to Anarchy. This is more consistent with other government changes.

Closes #2047.

### DIFF
--- a/server/techtools.cpp
+++ b/server/techtools.cpp
@@ -838,7 +838,7 @@ static void research_tech_lost(struct research *presearch, Tech_type_id tech)
       notify_player(
           pplayer, nullptr, E_NEW_GOVERNMENT, ftc_server,
           _("The required technology for our government '%s' "
-            "was lost. The country has fallen in %s."),
+            "was lost. The country has fallen into %s."),
           government_name_translation(government_of_player(pplayer)),
           government_name_translation(pgov));
       handle_player_change_government(pplayer, government_number(pgov));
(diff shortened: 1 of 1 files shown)


server/techtools.cpp
```diff
@@ -850,30 +832,26 @@ static void research_tech_lost(struct research *presearch, Tech_type_id tech)
...
-      struct government *pgov = pick_random_government(pplayer);
+      // Lost the technology for the government; switch to Anarchy.
+      struct government *pgov = game.government_during_revolution;
 
       notify_player(
           pplayer, nullptr, E_NEW_GOVERNMENT, ftc_server,
           _("The required technology for our government '%s' "
-            "was lost. The citizens have started a "
-            "revolution into '%s'."),
+            "was lost. The country has fallen in %s."),
```
psampathkumar, line 841: I think fallen "into" anarchy, is more grammatically correct ?  Why change the "into" to "in" ?

psampathkumar:
There is one more :p, I thought you'll fix in both places at once :stuck_out_tongue:

lmoureaux:
> There is one more :p, I thought you'll fix in both places at once 😛

Started this PR too long ago, I don't remember what it's doing :D

psampathkumar:
I didnt test it (I dont know how to, dont have a save game where I lose a tech), I just read though the code cuz it caught my attention. LGTM. I just wanted to check if there is specific problems with not defining a gov during revolution which caused them to chose random government in the first place. 

But looks like a gov during revolution is compulsory and the game raises an error otherwise, so I think this should work as intended.

jwrober:
Also looks good to me.

PR #2090, opened 2023-12-27
Fix wrong numbers of turns displayed after buying

### LINKED ISSUE #1848: Rush buying a unit shows complete in 2 turns
**Describe the bug**
Not always, there is some condition that occurs that when you rush buy a unit the client will show 2 turns to finish and zero gold. The unit is finished at turn change as expected, but the city dialog shows an incorrect value.

**To Reproduce**
1. Open a game
2. Open a city building a unit such as a Caravan
3. Rush buy the unit
4. See error -- note that it does not happen at all times

**Expected behavior**
The city dialog shows correct values

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.1-dev, and in 3.0-rc.2
 - Ruleset/Longturn game (if applicable): Discovered in LTX LT76Teams, so don't know if its a rules thing or a client thing for sure.

**Additional context**

lmoureaux:
The client's accounting of the shield stock didn't always take the bought shields into account.

Closes #1848.

Backport candidate.

### DIFF
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -1834,6 +1834,8 @@ int city_change_production_penalty(const struct city *pcity,
             && target->value.utype != pcity->changed_from.value.utype)))) {
     unpenalized_shields -= pcity->bought_shields;
     penalized_shields += pcity->bought_shields;
+  } else {
+    unpenalized_shields += pcity->bought_shields;
   }
 
   /* Do not put penalty on these. It shouldn't matter whether you disband


jwrober (approved):
Works for me. Backport approved

lmoureaux:
Cannot backport because a protocol change is needed for the client to have enough information (would need to backport #1996 and #2018 as well)

jwrober:
Ok then let's not do it.

PR #2092, opened 2023-12-27
 Add "unit" tests checking that shipped rulesets load

### LINKED ISSUE #1049: Fix issues with #869
PR #869 was rolled back in #913. Fix the issues which cropped up and do the merge again.

lmoureaux:
This adds infrastructure for running unit tests and the first tests.

This is a rebase of #869.
Closes #1049.


lmoureaux:
Tests already spotted an issue with Royale :joy:

jwrober:
This doesn't work on windows msys2. Is it supposed to?

lmoureaux:
> This doesn't work on windows msys2. Is it supposed to?

The original PR mentions something along these lines, but I haven't tested.

jwrober:
I tested it on msys2 and the test fails to run. Probably due to the way "install" has to happen and the test probably needs to be run from the install directory?

jwrober:
If it's not really design to run on non-linux, maybe we alter the command so it will skip with a message?

lmoureaux:
IIRC the server was simply idling. I could check in the CI... would be nice to have tests on Windows as well

jwrober (approved):

PR #2093, opened 2023-12-27
Local sockets

### LINKED ISSUE #1051: Fix issues with #888
PR #888 was rolled back in #913. Fix issues with #888 and merge again.

lmoureaux:
This PR restores #888 (rolled back in #913). It introduces the ability to run a server on a "local" socket: a UNIX local domain socket on Linux and a named pipe on Windows. This should allow local games on Windows without annoying firewall interference.

Closes #1051.

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -468,7 +468,8 @@ int client_main(int argc, char *argv[])
   auto positional = parser.positionalArguments();
   if (positional.size() == 1) {
     url = QUrl(positional.constFirst());
-    if (!url.isValid() || url.scheme() != QStringLiteral("fc21")) {
+    // Supported schemes: fc21://, fc21+local://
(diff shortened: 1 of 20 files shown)


jwrober:
I can't tell if this is working. On my Msys2 local I can invoke the server with the `--local` option, but the client does not see it running and only shows connection options for a regular TCP connection.

lmoureaux:
> I can't tell if this is working. On my Msys2 local I can invoke the server with the `--local` option, but the client does not see it running and only shows connection options for a regular TCP connection.

That's how it should be -- a server started on a local socket isn't meant to be used by the "outside world". The client can only connect to such a server when it starts it by itself. If single-player works and starting a single-player game doesn't trigger any firewall/antivirus popup (which it normally does the first time in my Windows VM), then this PR is working.

jwrober (approved):
Based on the conversation I tried a local game from the client. Works as described.

PR #2094, opened 2023-12-27
Remove Manhattan distances from trade routes

### LINKED ISSUE #882: CLASSIC caravan revenue style uses a weird distance
**Describe the bug**

With `caravan_bonus_style CLASSIC`, the revenue is calculated using `map_distance` instead of `real_map_distance`. There is even a comment saying `Should this be real_map_distance?`. `map_distance` corresponds to the Manhattan distance and `real_map_distance` to the Euclidean one.

As a consequence of this, moving a caravan diagonally grants a bonus roughly 2x the bonus of a caravan moving along x or y (for the same number of moves).

**To Reproduce**
1. Read the code (`common/traderoutes.cpp`)

**Expected behavior**

I propose to change this to `real_map_distance` instead. A correction factor could be introduced to maintain the balance of existing rulesets (on average).

 - OS: any
 - Freeciv21 version: `master`
 - Ruleset/Longturn game (if applicable): any of the shipped rulesets with caravans that can `Enter marketplace`

**Additional context**

#880

lmoureaux:
The Manhattan distance is `|deltaX| + |deltaY|`, which only makes sense when you can't move in diagonal. Since caravans can use the diagonal, the "Euclidian" (shortest path) distance is better suited for trade routes.

This introduces an incompatibility by changing (lowering) trade revenue for rulesets currently using CLASSIC. Since trade routes are usually OP, I don't think this is a big deal.

Closes #882.

### DIFF
--- a/common/traderoutes.cpp
+++ b/common/traderoutes.cpp
@@ -449,23 +449,22 @@ int get_caravan_enter_city_trade_bonus(const struct city *pc1,
                                        struct goods_type *pgood,
                                        const bool establish_trade)
 {
-  int md, rmd, trade2, max_trade_2;
+  int rmd, trade2, max_trade_2;
   int tb = 0, bonus = 0;
 
   if (pc2) {
-    md = map_distance(pc1->tile, pc2->tile);
     rmd = real_map_distance(pc1->tile, pc2->tile);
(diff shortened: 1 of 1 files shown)


jwrober (approved):
LGTM

PR #2098, opened 2023-12-29
Fix tests with multi-config build generators

lmoureaux:
And add a test for it!


lmoureaux:
The CI is green :green_apple:

jwrober (approved):
i think this is a good next step. Still more work todo.

PR #2101, opened 2023-12-30
Add a weekly FreeBSD CI run

lmoureaux:
This PR fixes build errors on FreeBSD and adds a weekly CI run to ensure it doesn't break in the future. While debugging FreeBSD, I understood the cause of the test errors on Fedora and also included a fix/workaround for that (FreeBSD got a different solution).

I did a full CI run [in my local](https://github.com/lmoureaux/freeciv21/actions/runs/7361897765) (before squashing debug commits into a nice history, but the code was identical)


jwrober (approved):
Nice!

PR #2119, opened 2024-01-01
CI/CD Windows Clang same as macOS

jwrober:
Change the Windows Clang CI/CD action to use the same `lukka/run-vcpkg@v10` as we use for macOS. Both platforms rely on vcpkg, so it seems like we should use the same tool to install it. Caching works well and we might get better stability with the Windows Clang weekly builds this way, especially with the new release process that keeps the git hash up to date.

It runs fine in my fork - https://github.com/jwrober/freeciv21/actions/runs/7380419411/job/20077801399

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -191,13 +191,13 @@ jobs:
       VCPKG_ROOT: C:/vcpkg
       VCPKG_DEFAULT_BINARY_CACHE: ${{github.workspace}}/vcpkg/bincache
     steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - uses: lukka/run-vcpkg@v10
         name: Install dependencies
         with:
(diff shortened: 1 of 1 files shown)


lmoureaux:
Let's wait until tests pass in the fork

lmoureaux:
> Let's wait until tests pass in the fork

Tests didn't even start:
```
MSBuild version 17.8.3+195e7f5a3 for .NET Framework
MSBUILD : error MSB1009: Project file does not exist.
Switch: test.vcxproj
```
This is not a regression but would be nice to understand

jwrober:
Test won't work in either windows environment without some changes. I'm going to disable in the CI for now

jwrober:
New test in my fork

lmoureaux (approved):
Should work now

jwrober:
Yes it did and we got to see the cache work. Nice!

PR #2121, opened 2024-01-02
Small refactors in utility/shared.h/cpp

lmoureaux:
* Remove a function provided by Qt
* Switch a `const char *` to a `QString` (and update docs)
* Fix a bug spotted by `clangd`


jwrober (approved):
Nice!

PR #2123, opened 2024-01-02
QString-based interface for interpret_tilde

lmoureaux:
On top of #2121.

This PR unifies two functions, simplifying their interface to be more Qt-friendly.


lmoureaux:
clang-format... I knew it!

PR #2124, opened 2024-01-02
snapcraft.yaml - bring to stable 3.0.2

jwrober:
Bring `snapcraft.yaml` up to current stable


jwrober:
> For my own education, why change `source-branch` to `source-commit`?

I wanted to  ensure that I grabbed the branch at the correct state. This is how the flatpak works too FWIW.

PR #2127, opened 2024-01-03
LAN announcements using DNS-SD

lmoureaux:
Use Kde's DNS-SD library to provide functionality equivalent to the UDP-based LAN announcement. This implements server-side support only. Client-side support will be added later. The CI and documentation are not implemented either.

Given all the missing pieces, I consider this patch proof-of-concept quality. To test it, spawn a server and run `avahi-discover` on any machine in the local network. I would still like feedback on whether it's working or not (with the equivalent of `libkf5dnssd-dev` installed).

vcpkg doesn't have a `kf5dnssd` package yet, so we'd need to add it.


lmoureaux:
This won't work with kdns-sd: https://invent.kde.org/frameworks/kdnssd/-/merge_requests/17

PR #2134, opened 2024-01-04
Fix passing the clang-tidy output to Codacy

lmoureaux:
Also upgrade codacy-clang-tidy to 1.3.8.

I was growing a bit worried not to see anything from Codacy. Turns out it's been broken ever since we moved to Ubuntu 22.04 because Codacy can't parse the colored output produced by clang 14 (which is the default). Use `ansifilter` to remove it.
I expect this PR to generate dozens of new warnings.

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -59,7 +59,9 @@ jobs:
 
       - name: Run clang-tidy
         run: |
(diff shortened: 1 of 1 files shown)


lmoureaux:
>  I expect this PR to generate dozens of new warnings.

Except they will only come once we have the code on `master` (this is because the action needs to run in a privileged context to gain access to secrets).

jwrober (approved):

jwrober:
Good catch.

PR #2138, opened 2024-01-06
Apply iwyu recommendations to utility/

lmoureaux:
include-what-you-use is a tool to sanitize includes in order to maximize independence between headers and increase compilation speed by reducing their size. This patch applies IWYU recommendations to utility/ and fixes the build in other files (they were often relying on transitive includes).


jwrober (approved):
Looks like you touched all sort of files outside of `utility/`. I guess that is the "fix the build" comment in action. Like the improved compile speed.

PR #2139, opened 2024-01-06
Fix and modernize listener<>

lmoureaux:
Commit d1f50af71316b3ab817f0438aecd0a639a93af63 broke the listener<> template in a subtle way: the destructor wouldn't remove listeners from the set because dynamic_cast to a derived type returns nullptr in a destructor. This caused a segfault upon saving map images.

In order to solve both the warning that lead to the commit and the new bug, change the set of instances to hold listener<A>* instead of A*, and use dynamic_cast in invoke since the A objects are still valid there.

While I was there, I changed invoke() to be a variadic template. It wasn't one because the code originally had to support C++03.

### DIFF
--- a/client/chatline.cpp
+++ b/client/chatline.cpp
@@ -47,7 +47,6 @@
 
 static bool is_plain_public_message(const QString &s);
 
-FC_CPP_DECLARE_LISTENER(chat_listener)
 QStringList chat_listener::history = QStringList();
 
 namespace {
--- a/client/chatline.h
+++ b/client/chatline.h
@@ -27,12 +27,10 @@ class QPaintEvent;
 class QPainter;
 class QPushButton;
 class QUrl;
-class chat_listener;
 
 void set_chat_colors(const QHash<QString, QString> &colors);
 QString apply_tags(QString str, const struct text_tag_list *tags,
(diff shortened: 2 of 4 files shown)


lmoureaux:
Backport required - or at least revert the equivalent of https://github.com/longturn/freeciv21/commit/d1f50af71316b3ab817f0438aecd0a639a93af63 in `stable`

jwrober (approved):
Looks good. Do we need to worry about the warning from the macOS CI?

blabber:
> Looks good. Do we need to worry about the warning from the macOS CI?

I see these in FreeBSD too. This is nothing to worry about in my opinion. The cause is that macOS and FreeBSD use clang/LLVM as the default compiler.

One could conditionalize the warning using something like this, I am not sure if it is worth it though:

`#if defined(__GNUC__) && !defined(__clang__)
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
#endif`

lmoureaux:
> Looks good. Do we need to worry about the warning from the macOS CI?

Further improving the code helped removing the warning.

jwrober (approved):

Issue #2136, opened 2024-01-04
Modpack installer more like a pkg manager


jwrober:
**Is your feature request related to a problem? Please describe.**
Right now, the modpack installer is kind of dumb. It simply grabs a list of files to download and does it. The problem is it doesn't handle:

- Complex dependency mapping between packs
- File versioning
- File management (e.g. if one version of a pack has a file and a second version doesn't, the old file isn't removed)
- Control and reporting of what is installed
- Better integration into the game interface
- It's ugly and utilitarian
- Easy way for new submissions

**Describe the solution you'd like**
We already have an open issue for the last bullet (better integration) in #1366. We could add a menu option for it as well in game. Would need to ensure that if a pack is installed in game (and probably off start menu) that a game restart isn't required for the game to see the change.

We have basic dep handling, but you can get cross-wise with circular mapping issues. Not sure we need to handle that per se, but better error messaging/handling would be good. Could be a CI/CD action check at PR open.

We don't do any kind of file list management. Easiest solution is to wipe all files on a re-download and start from scratch. Might be a better scalpel method though. e.g. delete old files and only download changed ones.

We fixed basic reporting in #2133. However might need more SQL changes to handle file level reporting.

Lastly, let's add theme support to the fcmp. Take existing theme the game is set to and display. Creates a more integrated experience. Then we don't have to provide `.desktop` and `metainfo` files for the utility.

Right now you need to know the `json` format and git/github to make submissions. As a nice to have, figure out a way to allow for easier submissions of new content. Maybe follow what flathub does with a repo per package? I'm sure there are other options as well.

**Describe alternatives you've considered**
Discussions have also occurred around using an established package manager. The challenge there is:

- Must be super cross-platform like fc21 already is
- Integrate into the game interface
- Not command line, provide a nice UI/UX
- Others I haven't thought of yet.

**Additional context**
Discord conversation starts here: https://discord.com/channels/378908274113904641/912500712833974322/1192480967407968287

jwrober:
In reviewing what Battle for Wesnoth has for their modpack installer:

- Global Search
- Filter by State, Type, Tags, List Order
- Pack details
     - Icon associated with the pack (sexy!)
     - version
     - Author
     - Install/download Size
     - # of installs (showing popularity)
     - Type
     - Inline button for install/remove
- Detailed description when selected
- Ability to install prior/older versions - not sure if this is something we really need

Not that we have to copy them or anything, but some ideas

[...]

PR #2143, opened 2024-01-07
Add an Add Ons button to start screen that loads mp installer

### LINKED ISSUE #1366: Add modpack installer to Start Screen
I am not really sure how many people know about the modpack installer as a way to download added content from our third party set of developers. I think it would be best to promote this more.

Add a "Add Ons" or "Mods" or some other such button to the main start screen. This button will bring up the modpack installer allowing the user to pick what they want.

None really. Obviously leaving it alone or putting some text there is another option, but button seems the best solution.

**Additional context**
I was reviewing our `docs` repo and thought of it while reading the page about using the modpack installer.

jwrober:
Closes #1366

Part of #2136

### DIFF
--- a/client/fc_client.cpp
+++ b/client/fc_client.cpp
@@ -12,6 +12,7 @@
 // Qt
 #include <QApplication>
 #include <QDirIterator>
+#include <QProcess>
 #include <QSettings>
 #include <QStackedLayout>
 #include <QStandardPaths>
@@ -19,6 +20,7 @@
 #include <QTcpSocket>
 #include <QTextBlock>
 #include <QTextCodec>
+
 // utility
 #include "fcintl.h"
 // common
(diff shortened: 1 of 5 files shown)


client/fc_client.cpp
```diff
@@ -687,6 +689,44 @@ void fc_client::start_new_game()
...
+    location = QStandardPaths::findExecutable(modpack_name);
+  }
+
+  // Start it
+  qInfo(_("Starting freeciv21-modpack-qt at %s"), qUtf8Printable(location));
+
+  QProcess *modProcess = new QProcess(this);
+  QStringList arguments;
+  modProcess->start(location, arguments);
+  modProcess->waitForFinished();
```
lmoureaux, line 727: We probably don't want to freeze the main window until the installer is closed, potentially triggering dead app dialogs from the system

lmoureaux (changes requested):
On Windows the modpack installer is optional. Should we make it mandatory?

jwrober:
> On Windows the modpack installer is optional. Should we make it mandatory?

Just noticed this. Yes I think it has to be now. I'll push another change.

jwrober:
Turns out the GUI modpack, isn't optional - https://github.com/longturn/freeciv21/blob/master/tools/fcmp/CMakeLists.txt#L34 - the CLI version is so we are good.

lmoureaux (approved):
We may want to rebrand the installer item as "developer tools" or so then

Issue #1904, opened 2023-04-06
Very large cities cause the right panel in city view to overlap the working area


jwrober:
**Describe the bug**
If you have a very large city (e.g. ~40) the citizens tab will show all of the citizen icons in a row and will make the right panel super wide so it overlaps the working area of the city making it impossible to manage citizen workers on the map.

**To Reproduce**
1. Open a game
2. Click on a very large city to open city dialog
3. See error

**Expected behavior**
The panel should not cover up the map

**Screenshots**
![image](https://user-images.githubusercontent.com/23037938/230501657-030f4c46-72bb-44bf-a4fc-f5d96da0418a.png)

**Platform and version (please complete the following information):**
 - OS: Windows, but probably happens everywhere
 - Freeciv21 version: v3.0 (stable)
 - Ruleset/Longturn game (if applicable): n/a

**Additional context**
none

PR #2144, opened 2024-01-08
Rewrite drawing of citizen sprites to use multiple rows

### LINKED ISSUE #2041: Large cities lose ability to adjust specialists

daavko:
Several things in this PR:
- Citizens are now drawn in multiple rows, where amount per row is controlled via `CITIZENS_PER_ROW`
- Fixed specialist rotating code so the correct specialist is picked even for large cities
- Add size hinting to `city_label` and several other places to prevent them being squished by the unit list when a city has too many units in it

Given that I don't really know C++, this is on a best-effort basis, but hopefully it's at least a bit helpful.

Fixes #2041 
Could partially work for #1904 as well, since that's caused by too many citizens in the second tab

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -1455,7 +1455,7 @@ void city_dialog::update_buy_button()
    Fill a pixmap with citizen sprites
  */
 void city_dialog::fill_citizens_pixmap(QPixmap *pixmap, QPainter *painter,
-                                       citizen_category *categories,
+                                       const citizen_category *categories,
                                        int num_citizens)
 {
   int sprite_width = tileset_small_sprite_width(tileset);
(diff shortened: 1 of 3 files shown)


client/citydlg.cpp
```diff
@@ -1422,81 +1451,100 @@ void city_dialog::update_buy_button()
...
-    }
+        pcity, static_cast<citizen_feeling>(i), categories);
+    this->fill_citizens_pixmap(citizen_pixmap, &painter, categories,
+                               num_citizens);
 
-    lab_table[k]->setPixmap(*citizen_pixmap);
+    lab_table[i]->setPixmap(*citizen_pixmap);
+    lab_table[i]->updateGeometry();
 
-    switch (k) {
```
lmoureaux, line 1481: I'd have used the opportunity to move this `switch` out of the loop. It shouldn't be inside...

client/citydlg.h
```diff
@@ -298,6 +304,8 @@ class city_dialog : public QWidget {
   void update_building();
   void update_info_label();
   void update_buy_button();
+  void fill_citizens_pixmap(QPixmap *pixmap, QPainter *painter,
+                            citizen_category *categories, int num_citizens);
```
lmoureaux, line 308: Since the function doesn't need to modify the `categories`, pass them by pointer-to-`const`:
```suggestion
                            const citizen_category *categories, int num_citizens);
```
This prevents accidental modification and may help the compiler optimize.

lmoureaux (commented):
The code looks sensible and seems to work well.
image
image
The unit area doesn't push the sides aside even with many units in the city and the gigantic `vectron` sprites.

I only have two small comments on your C++, which looks very good already :smile:

client/citydlg.h
daavko: Changed to `const`

client/citydlg.cpp
daavko: Yeah, I didn't really focus on this part during the rewrite. Turns out it's just easier, shorter, and more readable to remove the `switch` and call the functions one by one.

lmoureaux (approved):

Issue #565, opened 2021-07-26
Windows: Tilesets with accents in the path cannot be used


lmoureaux:
**Describe the bug**

[According to `kevin551`](https://discord.com/channels/378908274113904641/402213035865014272/869225691273654312), after downloading Aviation using the Modpack installer on Windows the tileset is not found by the client.

**To Reproduce**
1. Start the client, connect to `LTEx24`
2. Install Aviation from the Modpack installer
3. Try to switch to the Aviation tileset

**Expected behavior**

Aviation is visible in every possible tileset selection option.

**Platform and version (please complete the following information):**
 - OS: Windows (10?)
 - Freeciv21 version: 3.0-alpha3
 - Ruleset/Longturn game (if applicable): `LTEx24`

jwrober:
On Windows, the modpack installer places the files into `C:\Users\<UserID>\AppData\Roaming\freeciv21\3.0`. If you are in a game with the client open when you download something with the modpack installer the client is not aware of the change until you exit the game/close the client and then re-open and connect to a game again.

lmoureaux:
If it's not refreshed automatically, we need a refresh button :sweat_smile:

jwrober:
Behavior is the same on Linux so a refresh option makes sense.

lmoureaux:
Tested on Windows, for me everything was up-to-date after installing without restarting the client. Maybe one needs to reopen the settings window.

However, selecting aviation in the settings didn't work and I had to use the "Choose another tileset" menu.

lmoureaux:
> However, selecting aviation in the settings didn't work and I had to use the "Choose another tileset" menu.

--> #740 

This is just about a refresh button, not for beta1

lmoureaux:
This issue happened for the SG2 game. I was able to reproduce it on `wine` with 3.0-patch.2:
```
This is Freeciv21 version 3.0.16116404.2-patch.
Built against Qt 5.15.11, using 5.15.11

This instance of Freeciv21 searches the following directories for data files:

* Z:/home (user)
* Z:/home/louis (does not exist)
* C:/users/louis/Application Data/freeciv21/3.0 (user)
* D:/a/freeciv21/freeciv21/build/install/data (does not exist)
* C:/users/louis/Local Settings/Application Data/freeciv21 (does not exist)
* C:/ProgramData/freeciv21 (does not exist)
* Z:/home/louis/.wine/drive_c/Program Files/Freeciv21 (user)
[...]

PR #2145, opened 2024-01-08
Refactor encodings

lmoureaux:
This PR tries to sanitize the handling of character encodings:
* Remove support for non-utf-8 internal and data encodings.
* Remove all calls to the C file API and replace them with native Qt APIs. This should prevent path encoding issues like #565.

### Checks

I checked the following under `wine` for files with accents in the name:

* Reading serv files (from the server prompt, with the `-r` option, and from the client)
* Finding tilespec file
* Handling UTF-8 ruleset files (Royale's Ħal Saflieni Hypogeum)

I couldn't check the following because the file names don't have accents in my setup:

* Removing the test file used for cmdlevel hack
* Finding and using the client settings file and modpack DB

### DIFF
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -323,14 +323,7 @@ int client_main(int argc, char *argv[])
     init_ai(ai);
   }
 
-  init_nls();
-#ifdef ENABLE_NLS
-  (void) bindtextdomain("freeciv21-nations", get_locale_dir());
-#endif
-  init_character_encodings();
-#ifdef ENABLE_NLS
-  bind_textdomain_codeset("freeciv21-nations", "UTF-8");
-#endif
+  init_nls("freeciv21-nations");
 
   QCommandLineParser parser;
   parser.addHelpOption();
--- a/client/client_main.cpp
+++ b/client/client_main.cpp
@@ -327,9 +327,9 @@ int client_main(int argc, char *argv[])
 #ifdef ENABLE_NLS
   (void) bindtextdomain("freeciv21-nations", get_locale_dir());
 #endif
-  init_character_encodings(gui_character_encoding, gui_use_transliteration);
+  init_character_encodings();
 #ifdef ENABLE_NLS
-  bind_textdomain_codeset("freeciv21-nations", get_internal_encoding());
+  bind_textdomain_codeset("freeciv21-nations", "UTF-8");
 #endif
 
   QCommandLineParser parser;
--- a/client/connectdlg_common.cpp
+++ b/client/connectdlg_common.cpp
@@ -19,6 +19,7 @@
 #include <QTcpServer>
 #include <QUrl>
 #include <QUuid>
+#include <QtGlobal>
 
 #include <cstring>
 
@@ -48,7 +49,7 @@
 
 bool server_quitting = false;
 
-static char challenge_fullname[MAX_LEN_PATH];
+static QString challenge_fullname{};
 static bool client_has_hack = false;
 
 int internal_server_port;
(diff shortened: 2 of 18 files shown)


client/connectdlg_common.cpp
```diff
@@ -21,6 +21,7 @@
 #include <QUuid>
 
 #include <cstring>
+#include <qglobal.h>
```
jwrober, line 24: I'd recommend `#include <QtGlobal>` instead.

utility/fciconv.h
```diff
@@ -60,20 +60,15 @@
     encoding.
 */
 
-#define FC_DEFAULT_DATA_ENCODING "UTF-8"
```
jwrober, line 63: Why get rid of this? I think we should use global variables like this instead of hard coding in many places. If something is set in one place, that's one thing. But this is set in more than one spot.

client/client_main.cpp
```diff
@@ -327,9 +327,9 @@ int client_main(int argc, char *argv[])
 #ifdef ENABLE_NLS
   (void) bindtextdomain("freeciv21-nations", get_locale_dir());
 #endif
-  init_character_encodings(gui_character_encoding, gui_use_transliteration);
+  init_character_encodings();
 #ifdef ENABLE_NLS
-  bind_textdomain_codeset("freeciv21-nations", get_internal_encoding());
+  bind_textdomain_codeset("freeciv21-nations", "UTF-8");
```
jwrober, line 332: use "FC_DEFAULT_DATA_ENCODING" instead of "UTF-8"

server/srv_main.cpp
```diff
@@ -204,9 +204,9 @@ void srv_init()
   has_been_srv_init = true;
 
   // init character encodings.
-  init_character_encodings(FC_DEFAULT_DATA_ENCODING, false);
+  init_character_encodings();
 #ifdef ENABLE_NLS
-  bind_textdomain_codeset("freeciv21-nations", get_internal_encoding());
+  bind_textdomain_codeset("freeciv21-nations", "UTF-8");
```
jwrober, line 209: use "FC_DEFAULT_DATA_ENCODING" instead of "UTF-8"

tools/ruledit/ruledit.cpp
```diff
@@ -75,9 +75,9 @@ int main(int argc, char **argv)
   (void) bindtextdomain("freeciv21-ruledit", get_locale_dir());
 #endif
 
-  init_character_encodings(FC_DEFAULT_DATA_ENCODING, false);
+  init_character_encodings();
 #ifdef ENABLE_NLS
-  bind_textdomain_codeset("freeciv21-ruledit", get_internal_encoding());
+  bind_textdomain_codeset("freeciv21-ruledit", "UTF-8");
```
jwrober, line 80: use "FC_DEFAULT_DATA_ENCODING" instead of "UTF-8"

utility/fciconv.cpp
```diff
@@ -21,115 +21,45 @@
...
-    if (!internal_encoding) {
-      internal_encoding = local_encoding;
-    }
-  }
-  localCodec = QTextCodec::codecForLocale();
-  dataCodec = QTextCodec::codecForName(data_encoding);
-  internalCodec = QTextCodec::codecForName(internal_encoding);
 #ifdef FREECIV_ENABLE_NLS
-  bind_textdomain_codeset("freeciv21-core", internal_encoding);
+  bind_textdomain_codeset("freeciv21-core", "UTF-8");
```
jwrober, line 35: use "FC_DEFAULT_DATA_ENCODING" instead of "UTF-8"

jwrober (changes requested):
Looks good to me. Made some suggestions I think we need to iterate on.

utility/fciconv.h
lmoureaux: I elected to remove duplication instead and merged all `bind_textdomain_codeset` calls into a single location. The use of UTF-8 is baked pretty much all over the place (`QString::fromUtf8`, `qUtf8Printable`) so I don't think one more is a problem.

jwrober (approved):

PR #2146, opened 2024-01-08
Fix fileinfoname for non-ASCII characters

### LINKED ISSUE #565: Windows: Tilesets with accents in the path cannot be used

lmoureaux:
There was an encoding issue in fileinfoname() that prevented finding files with special characters in the path when the system encoding wasn't UTF-8 (fc_stat would fail as if the file didn't exist because it always received the path in UTF-8). Rewrite the function using only charset-independent Qt APIs.

(I'll want to test the installer in Wine for a final cross-check once actions finish running)

Backport candidate.

Closes #565.

### DIFF
--- a/utility/shared.cpp
+++ b/utility/shared.cpp
@@ -92,8 +92,6 @@ static QStringList scenario_dir_names = {};
 
 static char *mc_group = nullptr;
 
-Q_GLOBAL_STATIC(QString, realfile);
-
 /**
    An AND function for fc_tristate.
  */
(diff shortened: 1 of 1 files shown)


lmoureaux:
Works in `wine`.

jwrober (approved):

jwrober:
Work for me

PR #2150, opened 2024-01-15
Move add-ons button to be with options

jwrober:
Thanks @blabber for the suggestion. No related issue.


lmoureaux (approved):
By the way, I think `add-on` is a more widespread spelling. If we call them add-ons, should we also rename the modpack installer to something like "Add-ons manager"?

jwrober:
I can push another commit to change the spelling. I chose add-ons since its much smaller and easier than modpacks, which to me has always been a weird name/word. I would not be opposed to renaming the modpack installer to something relating to add-ons.

lmoureaux (approved):

blabber:
For me, an add-on is something that extends functionality. In my opinion, the Freeciv modpacks are not add-ons, but "Mods". And that is also a common term in the field of computer games (at least in Germany). So it might even be more appropriate to rename the button to Mods?

PR #2156, opened 2024-01-18
view_units: Fix find, disband and upgrade buttons

blabber:
The checks whether the find, disband and upgrade buttons should be enabled reference the wrong columns in the table.

This commit fixes the column references.

Reported by Corbeau on Discord.

***Screenshots demonstrating the fix***

| Before  | After |
| ------ | ----- |
| find broken | find fixed |
| none broken | none fixed |
| upgrade broken | upgrade fixed |


jwrober (approved):
Nice catch. Thanks for the patch.

PR #2159, opened 2024-01-19
Standardize buttons on start screen

jwrober:
No related issue.

image

image


lmoureaux:
Could we have more padding for long languages or large fonts?
image

Say `padding: 0.2em 1em`:
image

lmoureaux (approved):

PR #2160, opened 2024-01-19
Force Qt to use FreeType on Windows

lmoureaux:
Rendering of the Libertinus fonts was terrible with the (default) DirectWrite backend on Windows. Install a `qt.conf` file forcing Qt to use FreeType.

While reviewing Qt options I also found a way to make the window title bars dark if the user chose a dark theme. I enabled this as well.

Finally I moved files used for the Windows installer to dist/windows in an effort to organize that folder.

### DIFF
--- a/cmake/FreecivInstall.cmake
+++ b/cmake/FreecivInstall.cmake
@@ -41,10 +41,11 @@ if(WIN32 OR MSYS OR MINGW)
   # Custom command files to run the applications
   install(
     FILES
-    ${CMAKE_SOURCE_DIR}/dist/freeciv21-server.cmd
     ${CMAKE_SOURCE_DIR}/data/icons/128x128/freeciv21-client.ico
     ${CMAKE_SOURCE_DIR}/data/icons/128x128/freeciv21-modpack.ico
     ${CMAKE_SOURCE_DIR}/data/icons/128x128/freeciv21-server.ico
+    ${CMAKE_SOURCE_DIR}/dist/windows/freeciv21-server.cmd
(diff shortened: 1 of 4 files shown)


cmake/FreecivInstall.cmake
```diff
@@ -41,10 +41,11 @@ if(WIN32 OR MSYS OR MINGW)
   # Custom command files to run the applications
   install(
     FILES
-    ${CMAKE_SOURCE_DIR}/dist/freeciv21-server.cmd
     ${CMAKE_SOURCE_DIR}/data/icons/128x128/freeciv21-client.ico
     ${CMAKE_SOURCE_DIR}/data/icons/128x128/freeciv21-modpack.ico
     ${CMAKE_SOURCE_DIR}/data/icons/128x128/freeciv21-server.ico
```
jwrober, line 46: FWIW, the icons are Windows specific too

jwrober (changes requested):
Still evaluating, but to be consistent we should place the icons into the windows dir.

cmake/FreecivInstall.cmake
lmoureaux: I'm contemplating creating them on the fly with `icotool`, but I think this would go beyond the scope of this PR and shouldn't be backported.

jwrober: Actually I just realized that these are in the icon dir and not dist so no need to do anything.

jwrober:
My first test run on msys2 didn't get the .conf file installed. I'm trying again.

jwrober (approved):

PR #2161, opened 2024-01-19
cmake: Fix the option to disable man pages

### LINKED ISSUE #2155: Sphinx: Unable to disable making man pages
**Describe the bug**
When people compile the code for themselves in master, they should be able to disable making the RTD man pages, or ignore it when there is a error. 

**To Reproduce**
1. Compile without `sphinx_rtd_theme` in system. 
2. See Error
3. Set `-DFREECIV_ENABLE_MANPAGES` flag to `OFF` in CMake
4. Still See error

**Expected behavior**
Man pages are skipped with the flag. 

**Screenshots**
image

 - OS: Arch Linux
 - Freeciv21 version: [e.g. 3.1-alpha.4]

**Additional context**
It works in stable without this error, with the same PKGBUILD on arch.

blabber:
Fxes the handling of the `FREECIV_ENABLE_MANPAGES` option, leveraging the fact, that an `option` call is a no-op if the value is already set.

We also allow users on non-Unix systems to build manpages (for whatever reason one may want to do that).

Closes #2155

This needs a thorough review as I am non-fluent in cmake and don't have non-Unix systems available to test.

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -1,6 +1,10 @@
 # CMakeLists for Freeciv21 Docs
 
-find_package(Sphinx QUIET)
+if(FREECIV_ENABLE_MANPAGES)
+  find_package(Sphinx REQUIRED)
+else()
+  find_package(Sphinx QUIET)
+endif()
 
 if(SPHINX_FOUND)
   message(STATUS "Sphinx Found, configuring.")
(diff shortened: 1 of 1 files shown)


jwrober:
Well I started as non-fluent in Cmake too and I've got a lot of changes made over time. It grows on ya. Thanks for the PR. I'll have a look soon.

lmoureaux (commented):
I think we need to consider the following cases (assuming UNIX to simplify):
1. The user doesn't configure anything special and has sphinx. In this case we want to build man pages. This is working.
2. The user doesn't configure anything special and doesn't have sphinx. In this case we don't want to build man pages. This is working. (I think the option should still be defined in this case though, just default to `OFF`.)
4. The user sets `FREECIV_ENABLE_MANPAGES` to `OFF`. In this case we don't want to build the pages. This is working.
5. The user sets `FREECIV_ENABLE_MANPAGES` to `ON` and has sphinx. In this case we want to build man pages. This is working.
6. The user sets `FREECIV_ENABLE_MANPAGES` to `OFF` and doesn't have sphinx. In this case we want to build man pages but we can't and should emit an error at configure time. Currently we silently ignore the option.

To support 5 one would simply call `find_package(sphinx REQUIRED)` whenever `FREECIV_ENABLE_MANPAGES` is `ON`.

jwrober:
@blabber see @lmoureaux's comments. Let's support option 5 in the PR to make it complete.

lmoureaux (approved):

lmoureaux:
Thanks for the fix!

Issue #2154, opened 2024-01-18
Sphinx warning: deprecated syntax for intersphinx_mapping


lmoureaux:
**Describe the bug**

sphinx 7.2.6 warns:
```
WARNING: The pre-Sphinx 1.0 'intersphinx_mapping' format is deprecated and will be removed in Sphinx 8. Update to the current format as described in the documentation. Hint: "intersphinx_mapping = {'<name>': ('https://docs.python.org/', None)}".
```
https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping

**To Reproduce**
1. Build on Arch
2. Look out for Sphinx messages

**Expected behavior**
We don't use deprecated syntax.

**Platform and version (please complete the following information):**
 - OS: Arch Linux
 - Freeciv21 version: master, stable likely affected too

PR #2162, opened 2024-01-20
Docs: make external packages optional

lmoureaux:
It was noted in #2155 that when packaging Freeciv21, sphinx_rtd_theme and sphinx_last_updated_by_git were needed at build time to create the man pages. The rtd theme is obviously only needed for HTML output and last_updated_by_git doesn't seem to be used with the man pages. Make them both optional by checking if they can be imported when evaluating the config.

This will allow package maintainers to ship man pages even if one of the two packages are missing from their distribution.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -229,7 +229,7 @@ epub_exclude_files = ['search.html']
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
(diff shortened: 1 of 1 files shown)


jwrober:
If you add `/3/` to this, you can close #2154 

https://github.com/longturn/freeciv21/blob/master/docs/conf.py#L221

lmoureaux:
Added fix for #2154

jwrober (approved):

Issue #2158, opened 2024-01-18
Remove modpack installer shortcut from packages


jwrober:
**Is your feature request related to a problem? Please describe.**
We have recently added a button on the start screen to open the modpack installer, so we don't need windows shortcuts or `.desktop` files for the app any longer.

**Describe the solution you'd like**
Remove these from the NSIS and `/dist` directories. Update varying installation points/packages.

**Describe alternatives you've considered**
Can keep them, but its unnecessary icons in start menus and such.

**Additional context**
None

lmoureaux:
Maybe first add an item to the Game menu so one can also install modpacks during a game?

jwrober:
That's a good idea. Easy to implement the basics, but how do we "refresh" so the client knows something changed?

lmoureaux:
Hmm good point, currently we need to restart the game if we want to see the new modpacks. This is because `get_tileset_list` in `tilespec.cpp` uses a static vector. Removing it could affect performance so should be evaluated "carefully". I'm not sure how sound sets and music sets are discovered, maybe they already magically refresh.

lmoureaux:
Blockers for this have been addressed. Will we still need a modpack installer icon after all? :sweat_smile:

jwrober:
Yep, just don't need a "start menu" type icon.

lmoureaux:
Note that on Wayland window icons are tied to desktop files, so we'll lose the modpack installer icon on Wayland if we remove the `.desktop`. Maybe it's possible to have desktop files that are not in the menu though.

PR #2165, opened 2024-01-21
Add an option to add Modpacks from the Game menu

jwrober:
Also fix two issues where the client would not refresh when content on disk changed after the client was opened. This allows the client to see changes after content is downloaded with the modpack installer. See #2158

### DIFF
--- a/client/audio/audio.cpp
+++ b/client/audio/audio.cpp
@@ -84,8 +84,9 @@ const QVector<QString> *get_soundplugin_list(const struct option *poption)
 const QVector<QString> *get_soundset_list(const struct option *poption)
 {
   static QVector<QString> *sound_list = new QVector<QString>;
-  sound_list->clear();
-  *sound_list = *fileinfolist(get_data_dirs(), SNDSPEC_SUFFIX);
+  auto list = fileinfolist(get_data_dirs(), SNDSPEC_SUFFIX);
+  *sound_list = std::move(*list);
+  delete list;
   return sound_list;
 }
 
@@ -95,8 +96,9 @@ const QVector<QString> *get_soundset_list(const struct option *poption)
 const QVector<QString> *get_musicset_list(const struct option *poption)
 {
   static QVector<QString> *music_list = new QVector<QString>;
-  music_list->clear();
-  *music_list = *fileinfolist(get_data_dirs(), MUSICSPEC_SUFFIX);
+  auto list = fileinfolist(get_data_dirs(), MUSICSPEC_SUFFIX);
+  *music_list = std::move(*list);
+  delete list;
   return music_list;
 }
 
--- a/client/audio/audio.cpp
+++ b/client/audio/audio.cpp
@@ -79,29 +79,14 @@ const QVector<QString> *get_soundplugin_list(const struct option *poption)
 }
 
 /**
-   Returns a static string vector of audiosets of given type available
-   on the system by searching all data directories for files matching
-   suffix.
-   The list is nullptr-terminated.
- */
-static const QVector<QString> *
-get_audio_speclist(const char *suffix, QVector<QString> **audio_list)
-{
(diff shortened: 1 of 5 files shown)


client/audio/audio.cpp
```diff
@@ -78,40 +78,26 @@ const QVector<QString> *get_soundplugin_list(const struct option *poption)
...
    Returns a static string vector of soundsets available on the system.
  */
 const QVector<QString> *get_soundset_list(const struct option *poption)
 {
-  static QVector<QString> *sound_list = nullptr;
-
-  return get_audio_speclist(SNDSPEC_SUFFIX, &sound_list);
+  static QVector<QString> *sound_list = new QVector<QString>;
+  sound_list->clear();
+  *sound_list = *fileinfolist(get_data_dirs(), SNDSPEC_SUFFIX);
```
lmoureaux, line 88: We need to `delete` the pointer returned by `fileinfolist` or we'll have a memory leak. In addition, clearing isn't needed before an assignment.
```suggestion
  auto list = fileinfolist(get_data_dirs(), SNDSPEC_SUFFIX);
  *sound_list = std::move(*list);
  delete list;
```

client/audio/audio.cpp
```diff
@@ -78,40 +78,26 @@ const QVector<QString> *get_soundplugin_list(const struct option *poption)
...
    Returns a static string vector of musicsets available on the system.
  */
 const QVector<QString> *get_musicset_list(const struct option *poption)
 {
-  static QVector<QString> *music_list = nullptr;
-
-  return get_audio_speclist(MUSICSPEC_SUFFIX, &music_list);
+  static QVector<QString> *music_list = new QVector<QString>;
+  music_list->clear();
+  *music_list = *fileinfolist(get_data_dirs(), MUSICSPEC_SUFFIX);
```
lmoureaux, line 99: We need to `delete` the pointer returned by `fileinfolist` or we'll have a memory leak. In addition, clearing isn't needed before an assignment.
```suggestion
  auto list = fileinfolist(get_data_dirs(), MUSICSPEC_SUFFIX);
  *music_list = std::move(*list);
  delete list;
```

client/tileset/tilespec.cpp
```diff
@@ -774,31 +774,29 @@ QEvent::Type TilesetChanged;
...
-    for (const auto &file : qAsConst(*list)) {
-      struct tileset *t =
-          tileset_read_toplevel(qUtf8Printable(file), false, topo);
-
-      if (t) {
-        tilesets[idx]->append(file);
-        tileset_free(t);
-      }
+  /* Note: this means you must restart the client after installing a new
+      tileset. */
```
lmoureaux, line 787: This comment is hopefully no longer valid :sweat_smile: 
```suggestion
```

client/tileset/tilespec.cpp
```diff
@@ -774,31 +774,29 @@ QEvent::Type TilesetChanged;
...
+  QVector<QString> *list = fileinfolist(get_data_dirs(), TILESPEC_SUFFIX);
+  tilesets[idx]->clear();
+  tilesets[idx] = new QVector<QString>;
```
lmoureaux, line 790: We don't need to allocate a new vector if we `clear`ed it before.
```suggestion
```

lmoureaux (changes requested):
With this change, I think it would make sense to forbid building the client without the mod installer. We should probably check our build options in CMake.

lmoureaux:
(We can do the CMake changes in another PR, but let's not introduce memory leaks)

client/audio/audio.cpp
jwrober: I figured a `clear()` before using it did the same. I'll fix.

lmoureaux (approved):
Nice work!

PR #2171, opened 2024-01-21
Remove "client" from the app name on Windows and in AppStream metadata

lmoureaux:
There is no need to use such a technical term in user-facing text. This also removes any need for translation.

### DIFF
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -199,7 +199,7 @@
   FunctionEnd
 
   Function RunFreeciv21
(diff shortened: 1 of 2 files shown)


jwrober (approved):
This looks good. I actually think we might want to back port this and fix flathub to match. I had to do some hacks in stable flathub to get it to work there. Reason for backport is so we don't have to fix it later when 3.1 becomes new stable.

lmoureaux:
Oh the `appstream-utils` calls? We could backport to remove those, yes.

Issue #1238, opened 2022-08-10
Economy View to give cities with redundant buildings


jwrober:
**Is your feature request related to a problem? Please describe.**
When you open the economy view (report) in the client, there is a column showing the number of an item that is redundant. However, the report does not tell you which cities have the redundant item(s). Right now you have to go find them yourself.

**Describe the solution you'd like**
Add a column that shows what city or cities the item is redundant in. Ideally a double-click of the city name would bring up the city view dialog box.

**Describe alternatives you've considered**
None

**Additional context**
The sell all redundant button also does not tell you what cities will be impacted by the sale or if some cities are skipped because something was sold there already in the turn.

jwrober:
Keeping open to see if we can add some link/button to open the city listed.

PR #2175, opened 2024-01-23
Add redundant city names

jwrober:
Part of #1238. List the names of cities where the redundant improvement resides. Does not offer a quick link at this time, but at least gives you the some more information. Units are never redundant, they are upgradeable, which is shown on the units view.


lmoureaux (approved):
Works well for a small number of redundant buildings

PR #2176, opened 2024-01-24
Add AUR to README

psampathkumar:


lmoureaux (commented):
Should we add a section next to Nix and FreeBSD?

Is there a way to make the label as tall as the other two?
image

psampathkumar:
I tried, but no. 

Regarding nix and bsd, i considered that, but i think its better to just give a badge and not repeat instructions which are already on arch wiki.

jwrober:
What about horizontal centering of the image/badge with the other two? Maybe put all of them into an html table?

blabber:
Too bad that AUR and FreeBSD use incompatible versioning schemes (the FreeBSD framework cannot handle the scheme that Freeciv uses). Otherwise you could also use a Repology badge:

https://repology.org/project/freeciv21/badges

lmoureaux (approved):

PR #2180, opened 2024-01-27
Bring docs up to date with some gui changes

jwrober:
No related issue.


lmoureaux (commented):
Are the window decorations on the new screenshot there to make the Flatpak team happy? If so, the screenshot should also include the window shadow and be significantly smaller so it gets displayed at 1:1 on their website.

jwrober:
I made this change before I was aware of the flathub business. I made it to make me happy, not them :)

lmoureaux:
I think it would make sense to make the screenshot smaller, I can't even display it at 1:1 on my screen. Let's take the Flathub dimensions as a guideline?

jwrober:
I'm fine to make the image smaller. I don't like the dropshadow requirement tho.

lmoureaux:
> I'm fine to make the image smaller. I don't like the dropshadow requirement tho.

Make it smaller without window decorations and if they're unhappy, explain that there is a conflict in their guidelines between "default configuration" (which is fullscreen) and "drop shadows". Since shadows are cosmetic, we choose to follow the configuration guideline.

jwrober:
I'll retake the screen shot in full screen mode and resize following their guidelines on size and push another commit.

lmoureaux (approved):

PR #2202, opened 2024-02-10
Remove gui theme dir

### LINKED ISSUE #2172: Drop `gui-qt` from themes path
A holdover from legacy freeciv that had different clients with different theme options, we have `data/themes/qui-qt/*`. Well we don't need that any more.

Change the theme code to not need that path and move the files up a level to get rid of `qui-qt`.

**Additional context**

jwrober:
Closes #2172


lmoureaux (commented):
The diff has many changes seemingly unrelated with the PR. While they look ok (reordering headers to be more consistent), it would be better to have them in a separate commit (and possibly PR, but I can live with it).

jwrober:
I cleaned lot of ref's to `//gui-qt` in the headers so that did create some noise.

PR #2208, opened 2024-02-16
fcmp: Fix bogus "query failed" message

blabber:
When the modpack installer updates an already installed modpack, it emits a bogus critical message that a query failed with error code 100.

Error code 100 is `SQLITE_ROW` and is not really an error, but signals that a data row has been successfully read. `SQLITE_ROW` is also an expected return value in `mpdb_update_modpack`.

Change the check for failed queries to ignore `SQLITE_ROW`.

### DIFF
--- a/tools/fcmp/download.cpp
+++ b/tools/fcmp/download.cpp
@@ -318,6 +318,7 @@ const char *download_modpack(const QUrl &url, const struct fcmp_params *fcmp,
           return msg;
         }
       }
+      delete[] inst_ver;
     }
   }
 
--- a/tools/fcmp/mpcli.cpp
+++ b/tools/fcmp/mpcli.cpp
@@ -73,6 +73,8 @@ static void setup_modpack_list(const QString &name, const QUrl &url,
   if (!notes.isEmpty()) {
     qInfo() << _("Comment=") << notes;
   }
+
+  delete[] tmp;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


lmoureaux (commented):
From the docs, `sqlite3_step` seems to return `SQLITE_ROW` in some cases, which is also not handled. It also seems we always need to call `sqlite3_finalize` (even when the query failed), or we have a memory leak.

blabber:
That's what I get for trying to get rid of a misleading message :D
I'll take a closer look at the sqlite handling next week.

lmoureaux:
I didn't require changes at this point, just want to understand better :joy: We can also open an issue to check

lmoureaux:
Thanks for the improved memory handling! Am I officially allowed to run a build with `-fsanitize=leak` ? :japanese_ogre:

blabber:
Feel free :) I don't think I've made anything worse, but I don't know what surprises might linger in the code.

And I'm not sure whether the `QString`s will be deleted at some point. I don't really know QT, but I read something about reference counting.

lmoureaux:
> I'm not sure whether the QStrings will be deleted at some point. I don't really know QT, but I read something about reference counting.

A `QString` created on the stack (so without a pointer) will automatically free its resources when going out of scope at the end of the function. There is some internal magic to make copying more efficient, but we don't need to care about this.

lmoureaux (approved):
ASan didn't detect any build (not that it detected any before - I suppose closing the sqlite session frees the resources as well)

PR #2210, opened 2024-02-24
Add refresh button for techleak

### LINKED ISSUE #1753: Make it possible to forcibly research a tech that is instantly researchable after TC due to tech leak
Currently, if RNGeezus wants, a player may end up in a situation where they have more bulbs stored than their current research target costs, due to tech leak. This happens due to player processing order, where a player might gain just-not-enough bulbs to research a tech, but a player processed later might have enough, causing the cost to drop due to tech leak. Sometimes you can then "force" the game to research the tech by switching to another one that's sufficiently expensive (so you don't insta-research it and lose your bulbs) and then switching back, forcing the server to process the research. This may not always be possible, forcing the player to wait until next TC.

Two potential solutions:
1. Make the server somehow find out if tech leak would affect players that have already been processed, and process their research again (this might do weird stuff with obsoletions and stuff, I don't know the exact sequence of events in the TC sequence)
2. Quite a bit simpler, when the player sets their tech goal to the one they're currently researching, the server accepts this and processes it as if the player switched to any other tech. I'm pretty sure this would mean just removing lines 1020-1022 in `techtools.cpp`, but this is only from a cursory search. It does appear that the client sends the "switch research" packet even when the player clicks on a tech they're currently researching.

Leaving it as-is...

**Additional context**
No additional context

jwrober:
Closes #1753


client/views/view_research.cpp
```diff
@@ -500,6 +512,12 @@ void science_report::update_report()
     researching_combo->setDisabled(false);
     goal_combo->setDisabled(false);
   }
+
+  // If tech leak happens we enable a button to force/push a refresh.
+  if (done >= total) {
+    refresh_but->setVisible(true);
```
blabber, line 518: Just out of curiosity, are you changing the visibility of the button here instead of disabling it for a reason?

I'm asking because I personally find it a bit confusing when controls are sometimes visible and sometimes not.

lmoureaux: I also think hiding a control is bad practice (the exception we make is when the thing it controls is disabled in the rules, as is done for controls related to nationality). I would prefer to keep it visible at all times. In fact, I don't think even disabling this button is needed: refreshing is harmless when it doesn't do anything.

lmoureaux (approved):

Issue #2177, opened 2024-01-27
Warn when declaring war would imply cancelling alliances


lmoureaux:
**Is your feature request related to a problem? Please describe.**

In games with complex alliance systems, declaring war (or cease-fires running out) can have unintended consequences.

**Describe the solution you'd like**

* When war would result in an alliance being cancelled, require the user to cancel the alliance first. 
* When a cease-fire running out would result in an alliance being cancelled, automatically lengthen the cease-fire until the alliance gets cancelled.

**Describe alternatives you've considered**

Message boxes asking for confirmation. It sounds more difficult to implement for no additional functionality.

**Additional context**

Happened twice in SG2 already...

PR #2214, opened 2024-03-15
Prevent user from declaring war if it would break alliances

blabber:
Part of #2177.


lmoureaux (changes requested):
From reading the rest of the function, it seems that part of the code under `Check fall-out of a war declaration` is now unreachable and should be removed. This section also hints that special handling could be needed for teams.

blabber:
Thanks for the hint, it would indeed be sensible to not only prevent the player from breaking his own alliances, but also the alliances of their team members. I'll look I to it.

lmoureaux:
> Thanks for the hint, it would indeed be sensible to not only prevent the player from breaking his own alliances, but also the alliances of their team members. I'll look into it.

Are we going towards diplomacy acting at the team level instead of the nation level? This sounds like a real possibility...

blabber:
> Are we going towards diplomacy acting at the team level instead of the nation level? This sounds like a real possibility...

I wouldn't go that far, and I'm not sure it would be worth the effort: most LT games seem to be teamless after all.

But where it makes sense to consider the diplomatic relations of your teammates, it certainly can't hurt.

blabber:
I converted this PR to draft, I need to test the change in a teams environment first.

blabber:
This seems to work as intended. I squashed the commits and switched the PR back to "ready for review".

team

lmoureaux:
Hmm, how can one cancel a pact on the first encounter?

blabber:
"Cancel treaty" in the nations view. I guess as the Americans were allied with our team mate we had an implicit treaty?

lmoureaux:
Oh it was a manual action. From the screenshot I thought the two messages were generated in sequence by the game

lmoureaux (approved):

Issue #2069, opened 2023-12-02
The macOS CI is failing


lmoureaux:
**Describe the bug**

We are affected by https://github.com/microsoft/vcpkg/issues/35332. This prevents macOS builds from completing.

**To Reproduce**
1. Trigger a CI build
2. Wait until vcpkg starts building fontconfig
3. See an error related to missing `distutils`

**Expected behavior**
The build completes.

**Platform and version (please complete the following information):**
 - OS: macOS
 - Freeciv21 version: `master` and `stable`

**Additional context**
`distutils` was removed in Python 3.12. Meson failed to follow suit until it was too late.

PR #2220, opened 2024-03-17
Fix the macOS CI build

lmoureaux:
The macOS CI stopped working, seemingly because multiple Python versions are installed simultaneously. We forced the Python version in response to #2069. There is now a proper fix for #2069 in upstream vcpkg. Update the vcpkg version so we can use it.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -235,7 +235,7 @@ jobs:
           brew link gettext --force
       - uses: actions/setup-python@v5  # Workaround for #2069
         with:
-        python-version: '3.11'
+          python-version: '3.11'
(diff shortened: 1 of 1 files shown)


blabber:
Is this still ready for review? This did not fix the build, so I guess work on this PR is ongoing.

lmoureaux:
> Is this still ready for review? This did not fix the build, so I guess work on this PR is ongoing.

Right, thanks. Switched to draft

lmoureaux:
Let's **squash and merge** if the pipeline succeeds

jwrober:
100%, so only one commit goes into master and stable branches.

jwrober (approved):

PR #2222, opened 2024-03-17
Minimal Qt6 port

lmoureaux:
This PR implements a minimal port to Qt 6:
* The code builds
* Tested basic client, server, fcmp things and fixed the bugs I encountered
* Tests pass on my local, not in the CI
* macOS, Nix, Snap, Flatpak integrations are broken. I have a patch for Nix
* There are many build warnings
* Documentation is outdated

If you have Qt6 and Kf6 on your platform, give this branch a try! I'm sure there are many more bugs.


client/CMakeLists.txt
```diff
@@ -199,11 +204,11 @@ if (AUDIO_SDL)
 endif()
 
 if (APPLE)
-  qt5_import_plugins(freeciv21-client INCLUDE Qt5::QCocoaIntegrationPlugin Qt5::QSvgPlugin)
+  qt6_import_plugins(freeciv21-client INCLUDE Qt6::QCocoaIntegrationPlugin Qt6::QSvgPlugin)
 endif()
 
 if(EMSCRIPTEN)
-  qt5_import_plugins(freeciv21-client INCLUDE Qt5::QWasmIntegrationPlugin)
```
jwrober, line 206: If we are no longer supporting wasm in the CI, do we need to check for EMSCRIPTEN either?

cmake/FreecivDependencies.cmake
```diff
@@ -21,7 +21,7 @@ check_function_exists(at_quick_exit HAVE_AT_QUICK_EXIT)
 find_package(Python3 REQUIRED)
 
 # Required as the main networking and utility library
-find_package(Qt5 5.15 COMPONENTS Core Gui Network REQUIRED)
+find_package(Qt6 6.7 COMPONENTS Core Gui Network REQUIRED)
```
jwrober, line 24: Qt 6.8 is LTS

utility/registry_ini.cpp
```diff
@@ -149,7 +149,8 @@
   - Now uses hash.c
  */
 // KArchive
-#include <KFilterDev>
+#include <KCompressionDevice>
+#include <kcompressiondevice.h>
```
jwrober, line 153: duplicate header

jwrober (commented):
some comments

cmake/FreecivDependencies.cmake
lmoureaux: I'll likely need to drop to 6.6 for Ubuntu 22.10

lmoureaux: There was a suggestion to go all the way back to 6.5 since it's an LTS release. However, no distro seems to ship it anymore.

lmoureaux:
I don't understand why the tests keep failing, they work fine on my system. It looks like the server always exits with an error code. I have a different Qt version (6.8), but apart from that the setup should be pretty similar. Need to investigate more.

client/CMakeLists.txt
lmoureaux: I'm not sure honestly. Removing it from the CI might have been a bold move, I just didn't want to spend time on getting it to work there.

lmoureaux: Will keep the code for now

client/fc_client.cpp
```diff
@@ -61,7 +60,7 @@ fc_client::fc_client() : QMainWindow(), current_file(QLatin1String(""))
   setWindowFlags(Qt::FramelessWindowHint);
   setWindowState(Qt::WindowFullScreen);
 #endif
-  QTextCodec::setCodecForLocale(QTextCodec::codecForName("UTF-8"));
+  // QTextCodec::setCodecForLocale(QTextCodec::codecForName("UTF-8"));
```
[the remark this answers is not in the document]
lmoureaux: Will see what breaks in real life.

blabber:
Foreign units without a name are still displaying garbage names in the tile info popup (middle-click). I have a fix that can be cherry-picked, the commit hash is fe8eb07.

PR #2226, opened 2024-03-22
adjust tab order (closes #2224)

hugomflavio:

PR #2229, opened 2024-03-22
allow using enter to select item on main menu (closes #1685)

### LINKED ISSUE #1685: Allow pressing <enter> key to select options on main screen
When the client starts, we can use the arrows to move up and down on the options ("tutorial", "start new game", etc). But pressing "Enter" doesn't select the option, so we end up having to use the mouse anyway. Would be nice if we could have a keyboard-only pathway to connect to a server game. You can press the spacebar and it works, but that does not feel intuitive any longer

Keep spacebar, but add the enter key as well for selecting the highlighted option.

**Additional context**
Discord - https://discord.com/channels/378908274113904641/1059497144874192906/1059497144874192906

hugomflavio:


lmoureaux:
This will likely clash with #2226 and need a rebase. Do them both in one PR?

lmoureaux (approved):

PR #2235, opened 2024-03-24
Allow tilesets to specify a default zoom level

lmoureaux:
This is used to render the tileset by default and when the user resets the zoom. Useful to emulate HiDpi.

Vectron will use this.

### DIFF
--- a/client/tileset/tilespec.cpp
+++ b/client/tileset/tilespec.cpp
@@ -326,6 +326,7 @@ struct tileset {
   int full_tile_width, full_tile_height;
   int unit_tile_width, unit_tile_height;
   int small_sprite_width, small_sprite_height;
+  double preferred_scale;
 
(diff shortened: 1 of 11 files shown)


lmoureaux:
I should probably do some effort to document this. Let me switch back to draft

lmoureaux:
Not sure I like the name of the parameter anymore. Suggestions?

jwrober:
Is this ready to review? Seems to work on my local.

lmoureaux:
> Is this ready to review? Seems to work on my local.

Yes it is. Any suggestion on the name of the tileset parameter?

jwrober:
I think `preferred_scale` works unless `preferred_zoom` is better, but I'm not thinking that it is better.

jwrober (approved):

PR #2239, opened 2024-03-26
Fix issues introduced with the prevention of war declarations if they would break alliances

blabber:
Commit ecd275d7a487c8859446fe53d0b94745930a95db introduced a change that prevents players from declaring war, if this would break existing alliances. There are two remaining issues:

* AI will  fail to declare war, if entangled in a web of alliances
* Cease-fires running out, will end in an "undefined" state if war can't be entered because of alliances

To fix this:
- [x] restore the old behavior for the AI
- [x] automatically extend cease-fires while the fragile diplomatic state is balanced by alliances 

Part of #2177

### DIFF
--- a/server/plrhand.cpp
+++ b/server/plrhand.cpp
@@ -746,22 +746,22 @@ void handle_diplomacy_cancel_pact(struct player *pplayer,
                                   int other_player_id,
                                   enum clause_type clause)
 {
-  handle_diplomacy_cancel_pact2(pplayer, other_player_id, clause,
-                                is_human(pplayer));
+  handle_diplomacy_cancel_pact_explicit(pplayer, other_player_id, clause,
+                                        is_human(pplayer));
 }
 
 /**
    A variant of handle_diplomacy_cancel_pact that allows the caller to
-   control if war declarations should be prevented to protect existing
-   alliances.
+   explicitely control if war declarations should be prevented to
+   protect existing alliances.
 
    protect_alliances is a flag, that, if set to true, will prevent war
      declarations if they would break existing alliances.
  */
-void handle_diplomacy_cancel_pact2(struct player *pplayer,
-                                   int other_player_id,
-                                   enum clause_type clause,
-                                   bool protect_alliances)
+void handle_diplomacy_cancel_pact_explicit(struct player *pplayer,
+                                           int other_player_id,
+                                           enum clause_type clause,
+                                           bool protect_alliances)
 {
   enum diplstate_type old_type;
   enum diplstate_type new_type;
@@ -954,8 +954,8 @@ void handle_diplomacy_cancel_pact2(struct player *pplayer,
                       player_name(pplayer), player_name(pplayer2));
         player_diplstate_get(other, pplayer)->has_reason_to_cancel = 1;
         player_update_last_war_action(other);
-        handle_diplomacy_cancel_pact2(other, player_number(pplayer),
-                                      CLAUSE_ALLIANCE, false);
+        handle_diplomacy_cancel_pact_explicit(other, player_number(pplayer),
+                                              CLAUSE_ALLIANCE, false);
       } else {
         /* We are in the same team as the agressor; we cannot break
          * alliance with him. We trust our team mate and break alliance
@@ -966,8 +966,8 @@ void handle_diplomacy_cancel_pact2(struct player *pplayer,
                       player_name(pplayer),
                       nation_plural_for_player(pplayer2),
                       player_name(pplayer2));
-        handle_diplomacy_cancel_pact2(other, player_number(pplayer2),
-                                      CLAUSE_ALLIANCE, false);
+        handle_diplomacy_cancel_pact_explicit(other, player_number(pplayer2),
+                                              CLAUSE_ALLIANCE, false);
       }
     }
   }
--- a/server/plrhand.cpp
+++ b/server/plrhand.cpp
@@ -88,11 +88,6 @@ static void send_player_diplstate_c_real(struct player *src,
 static void send_nation_availability_real(struct conn_list *dest,
                                           bool nationset_change);
 
-static void handle_diplomacy_cancel_pact_internal(struct player *pplayer,
-                                                  int other_player_id,
-                                                  enum clause_type clause,
-                                                  bool protect_alliances);
-
 // Used by shuffle_players() and shuffled_player().
 static int shuffled_order[MAX_NUM_PLAYER_SLOTS];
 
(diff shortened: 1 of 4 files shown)


server/plrhand.cpp
```diff
@@ -792,17 +806,22 @@ void handle_diplomacy_cancel_pact(struct player *pplayer,
...
     {
       if (pplayer3 != pplayer && pplayer3 != pplayer2
           && pplayers_allied(pplayer3, pplayer)
           && pplayers_allied(pplayer3, pplayer2)) {
-        blocking_alliances = true;
-
         if (players_on_same_team(pplayer, pplayer3)) {
+          // if pplayer3 is an AI, let them break their alliance as we
+          // can't coordinate with them in a way, that they would cancel
+          // alliances in preparation of war on team level.
```
lmoureaux, line 820: One more argument for diplomacy to be handled at the team level :sweat_smile: (and by the human when there's one on the team)

server/plrhand.cpp
```diff
@@ -736,15 +741,24 @@ void player_update_last_war_action(struct player *pplayer)
...
  */
 void handle_diplomacy_cancel_pact(struct player *pplayer,
                                   int other_player_id,
                                   enum clause_type clause)
+{
+  handle_diplomacy_cancel_pact_internal(pplayer, other_player_id, clause,
+                                        is_human(pplayer));
+}
+
+static void handle_diplomacy_cancel_pact_internal(struct player *pplayer,
```
lmoureaux, line 758: What's the reason for the new function? `protect_alliances` could just be deduced in the beginning

server/plrhand.cpp
blabber: It would definitely make things clearer, but it would also be a lot of work 😅

server/plrhand.cpp
blabber: We need the new function as once we start breaking alliances (at the `Check fall-out of a war declaration`), this function is called recursively with `protect_alliances` set to `false`.

server/plrhand.cpp
```diff
@@ -911,6 +932,41 @@ void handle_diplomacy_cancel_pact(struct player *pplayer,
                 player_name(pplayer), nation_plural_for_player(pplayer2),
                 nation_plural_for_player(pplayer),
                 diplstate_type_translated_name(new_type));
+
+  // Check fall-out of a war declaration.
```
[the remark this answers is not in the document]
blabber: Done.

blabber:
I think I figured this out, but I still need to do some test.

I extend cease-fires by four turns, if returning to war would break alliances. Any feedback if this should be changed is appreciated. Extending by only one turn, would cause a constant stream of notifications about cease-fires about to run out.

If anyone has a good name for `handle_diplomacy_cancel_pact2` (formerly known as `handle_diplomacy_cancel_internal`), this would be helpful :)

blabber:
I managed to test the automatic extension of cease-fires in a test game and as far as I can tell, this works now. I wasn't able to test the AI, but as we are falling back to the old behavior for AI, this should be okay.

Please feel free to test and comment on the fact, that cease-fires are extended by four turns.

lmoureaux:
> Please feel free to test and comment on the fact, that cease-fires are extended by four turns.

Four turns is a quarter of the normal cease-fire duration. It may be hard for the player to cancel it later if they are in Democracy (or planning to be)...
At the very least this should be a constant defined in `diplhand.h` next to the cease-fire duration (inappropriately called `TURNS_LEFT`).

blabber:
Defining the extension period in a header is file is a good idea, I will definitely do that.

Regarding the number of turns, what was your idea when you created the issue?

Extension by one turn will work, but you will get the according notification every turn. This might not be as big an issue as I thought, as there are lots of notifications every turn in the game I am currently playing.

lmoureaux:
I was thinking one turn but didn't consider notification spam. Maybe it's good to keep reminding the player, like when a city is in disorder.
Casus belli should also be extended I guess?

blabber:
> Casus belli should also be extended I guess?

Maybe, I don't have enough knowledge about the inner workings of diplomacy to comment on this issue. But I will look into it.

It would help, if you could elaborate a little why you think casual belli should be extended too 😅

blabber:
Successfully tested with cease-fire extension by one turn. Cancelling the blocking alliance drops you into war at TC.

I am still not sure what to do with casus belli. I am not sure how cease-fire and casus belli interact with each other.

blabber:
> At the very least this should be a constant defined in `diplhand.h` next to the cease-fire duration (inappropriately called `TURNS_LEFT`).

`TURNS_LEFT`  is not cease-fire specific, but also affects armistice. Therefore a neutral name makes a little more sense.

lmoureaux:
From a quick inspection of the code, it seems that being at war with one of your allies doesn't provide a casus belli (the Senate will object war declaration). This is a bit weird but understandable: if one player in an alliance isn't subject to the Senate restrictions, this player could declare war at will and their allies simply follow. This would make the senate moot.

Is my interpretation of the code correct? ie given the following players, A can't declare war on C?
| | A | B | C |
|:-:|:-:|:-:|:-:|
| A | Democracy | Alliance | Peace |
| B |  | (any gov) | War |
| C |  |  | (any gov) |

blabber:
I am a little confused. It seems you are reasoning about code I haven't touched? I kept all the checks regarding war declarations in place and added the alliance check after all other checks have returned without objecting the war declaration. The rules I implement are agnostic about the government type.

It follows a description of the implemented rules. I think they are not too complicated, but difficult to explain. If we are concerned that players are having problems with these rules, this functionality could be made configurable.

If we want to make this configurable, I would need a few more hints on how to implement this.

### War declaration 

| | A  | B | C  |
| --- | --- | ---| --- | 
| A | | Alliance | |
| B | Alliance | | Alliance |
| C | | Alliance | |

A can't declare war against C, as it would

1. break the alliance with B, if B is not As teammate
2. force B to break its alliance with C, if B is As teammate

Exception: if B is As teammate and an AI, it will break the alliance with C if this allows A to declare war. 

### Cease-Fire 

| | A  | B | C  |
| --- | --- | ---| --- | 
| A | | Cease-Fire | Alliance |
| B | Cease-Fire | | Alliance |
| C | Alliance | Alliance | |

When the cease-fire between A and B runs out, it will extended by another turn, to protect the alliances with C. As soon as A or B cancels the alliance with C (or vice-versa), the cease-fire will actually run out at TC and they'll drop into war. This applies only, if A and B are human players.

lmoureaux:
> I am a little confused. It seems you are reasoning about code I haven't touched?

I am a bit confused as well :D The level of entanglement in these situations doesn't make things easy. Casi belli are events that allow cancelling treaties despite the Senate. For instance spies poisoning cities. Maybe also launching a spaceship.

My concern is that the game might give a casus belli to one of the players when the cease-fire runs out (or someone declares war). If we forcefully keep the cease-fire running, the casus belli might not appear again. This might prevent a war declaration that would have been possible otherwise.

Now looking at the code I couldn't find a casus belli generated when someone declares war, which I find weird but it seems to be this way. So what I described is probably not happening.

blabber:
I tried to make sense out of the casus belli related code, but I guess I will need some more time to make sense out of it...

But the Freeciv help [documents](https://freeciv.fandom.com/wiki/Diplomacy#Senate_and_Casus_Belli) that a casus belli is provided if
> An alliance is broken due to a war between a player's allies (given until turn end to the side(s) you break alliance with). 

Breaking alliances due to a war is exactly what my code tries to prevent. On the other hand cancelling an alliance and declaring war afterwards seems not to provide a casus belli, therefore cautious players could already avoid this casus belli.

blabber:
After reverting the commit on which these changes were based, I will close this PR and open another on containing the whole changeset.

PR #2243, opened 2024-03-28
Remove duplicate sprite error in trident and hex2t tilesets

jwrober:
No related issue. Getting rid of some noise on the console.


lmoureaux (changes requested):
Remove instead of commenting out?

jwrober:
I can do that.

lmoureaux (approved):

PR #2250, opened 2024-03-30
Colorize (red) negative values on city dialog

### LINKED ISSUE #1511: Colorize/highlight the city growth if its negative
If your city is consuming more food that is needed, it will lose population when the granary is empty. The counter in the city dialog in this example says 7 turns, but that is 7 turns to a smaller city, not a larger city.

image

Colorize (e.g. make the text red) or highlight the value for `change in` value to bring more attention to it when growth is negative. Might consider doing the same thing for the negative food at the top.

**Additional context**
Came up as part of evaluating PR #1494

jwrober:
Closes #1511

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -664,7 +664,7 @@ city_info::city_info(QWidget *parent) : QWidget(parent)
   };
 
   QLabel *dummy;
-  std::tie(dummy, m_food) = create_labels(_("<B>Food:</B>"));
+  std::tie(dummy, m_food) = create_labels(_("<b>Food:</b>"));
   std::tie(dummy, m_granary) = create_labels(_("Granary:"));
   std::tie(dummy, m_size) = create_labels(_("Citizens:"));
   std::tie(dummy, m_growth) = create_labels(_("Change in:"));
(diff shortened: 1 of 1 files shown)


jwrober:
CodeFactor found an issue with something I didn't change. Interesting.

lmoureaux (commented):
> CodeFactor found an issue with something I didn't change. Interesting.

You did change the method it's pointing at :smiley:

I don't think trade, lux, or science can go negative. How much of this code could be factored out in a utility function?

jwrober:
> I don't think trade, lux, or science can go negative. How much of this code could be factored out in a utility function?

If they don't go negative, then why "surplus"? 

IDK on the second question. Why the need for another function when in-line IF does the trick?

lmoureaux:
> If they don't go negative, then why "surplus"?

Good question :rofl: 

> IDK on the second question. Why the need for another function when in-line IF does the trick?

To avoid repeating the same code many times (so when we update it we update it everywhere…)

Also: do we need a different color for the classic theme?

jwrober:
> Also: do we need a different color for the classic theme?

The hard code "white" is an issue with Classic. Is there a way to unset a stylesheet?

jwrober:
I guess I could do a bigger IF with a collection of OR in the evaluation to reduce the number of them.

lmoureaux:
Where do we have white? I see white-space and red

Issue #2037, opened 2023-09-20
Add a UI toggle to highlight polluted tiles


daavko:
**Is your feature request related to a problem? Please describe.**
Sometimes it's hard to see which tiles are or aren't polluted, mostly when there's lots of units and infrastructure around. Some people solve this by creating custom tilesets where they add a highlight to pollution sprites, which makes it clearer which tiles are polluted.

**Describe the solution you'd like**
A simple "Highlight pollution" toggle somewhere in the menu, which does a single thing: highlights polluted tiles.

**Describe alternatives you've considered**
None. We could maybe highlight pollution in the shipped tilesets, but not everyone might want that, so an option strikes me as better.

**Additional context**
https://discord.com/channels/378908274113904641/379424175623766027/1154075733598142514

jwrober:
Another convo on this: https://discord.com/channels/378908274113904641/1221099665462657075/1221099665462657075

lmoureaux:
This is a good candidate for a [tileset option](https://longturn.readthedocs.io/en/latest/Modding/Tilesets/options.html).

lmoureaux:
`hexemplio` done in #2498, all we need now is a sprite for amplio. There's already a sprite in RoundSquare that would work well for trident.

jwrober:
Is this feature complete? I think we have this for high-vis polluted tiles in our main tilesets.

PR #2251, opened 2024-03-31
Implement tileset options

lmoureaux:
This is a general solution for requests like #2037. With these changes, tilesets can specify arbitrary options that modify their appearance according to what the users prefers. This feature is documented in a new page under Modding/Tilesets.

I used this feature to remove `cimpletoon` and `toonhex`, as they were just copies of `amplio2` and `hexemplio` with the Cimpletoon units. This is now supported as an option to the main tilesets, reducing maintenance burden. So this PR adds a feature that reduces the total number of lines :D
Cimpletoon and Toonhex users will automatically be migrated to Amplio2 or Hexemplio with the appropriate configuration.

The ultimate test would be implementing #2037 for shipped tilesets :smiley:


client/menu.cpp
```diff
@@ -18,6 +18,7 @@
 #include <QMessageBox>
 #include <QStandardPaths>
 #include <QVBoxLayout>
+#include <qmenubar.h>
```
jwrober, line 21: Should this be `<QMenuBar>`?

client/tileset_options.cpp
```diff
@@ -0,0 +1,71 @@
+// SPDX-FileCopyrightText: 2024 Louis Moureaux
```
jwrober, line 1: I thought we were not using dates in SPDX, just the name.

jwrober (approved):
Other than the two small suggestions, rest looks good to me. Local test worked as well.

PR #2269, opened 2024-04-09
Always use angleDelta() for scrolling

### LINKED ISSUE #2236: Touchpad Scroll on Wayland not working
**Describe the bug**
The touchpad scroll only works left/right, up/down and not diagonally on Wayland. Is also laggy/slow. Works fine on X11

**To Reproduce**
1. Set your Linux to use Wayland instead of X11
2. Open a game
3. See error

**Expected behavior**
Mouse events work same on Wayland as on X11

**Screenshots**
N/A

 - OS: Linux
 - Freeciv21 version: 3.1-alpha.5

**Additional context**
https://discord.com/channels/378908274113904641/1109113177779732550/1215113440033636464

lmoureaux:
Using pixelDelta() resulted in seemingly meaningless values on KWin/Wayland. Qt code uses angleDelta() exclusively, so I see no reason not to do the same.

Closes #2236.

Suggest to backport (if we do another stable)

### DIFF
--- a/client/views/view_map.cpp
+++ b/client/views/view_map.cpp
@@ -376,10 +376,7 @@ void map_view::resizeEvent(QResizeEvent *event)
  */
 void map_view::wheelEvent(QWheelEvent *event)
 {
-  auto delta = event->pixelDelta();
-  if (delta.isNull()) {
-    delta = event->angleDelta();
-  }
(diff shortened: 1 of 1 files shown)


jwrober (approved):
LGTM

PR #2270, opened 2024-04-09
Improve wording for the "Show/hide workers" shortcut

### LINKED ISSUE #2249: Shortcut "Show/hide workers" does not show or hide workers
**Describe the bug**
The shortcut "Show/hide workers" does not show or hide workers, but toggles the city/unit color of the nearest city or settler, via `key_city_overlay`

https://github.com/longturn/freeciv21/blob/1e6a9a2d889bc5a7fab64559d4cb02bf150a0796/client/mapctrl.cpp#L318-L320

**To Reproduce**
1. Start game
2. Build city
3. Use the shortcut "Show/hide workers" on the city (default: Alt+Shift+Right Mouse)
4. Visibility of workers is not toggled, but the city area is colored

**Expected behavior**
1. Using the "Show/hide workers" shortcut should toggle visibility of workers.
2. There should be another, aptly named, shortcut to do what "Show/hide workers" currently does.

**Screenshots**
The map looks like this, after using "Show/hide workers" on a city:

city_overlay

 - OS: FreeBSD 14.0-RELEASE-p6
 - Freeciv21 version: 3.1-alpha.5

**Additional context**
Reported on Discord: https://discord.com/channels/378908274113904641/378908274680004619/1222791609192681523

lmoureaux:
This is the lowest-effort solution to address #2249. It doesn't follow the suggestion there, which is more complicated (and could become a feature request).

Closes #2249.


blabber (changes requested):
The documentation needs to be adapted accordingly. I think it would make sense to do this as part of this PR instead of creating a new issue.

https://github.com/longturn/freeciv21/blob/405bd0a66c77a8988cbb719a4a7f6b0ce07918c9/docs/Manuals/Game/shortcut-options.rst?plain=1#L137

PR #2278, opened 2024-04-13
Automate Snap builds - Take 1

### LINKED ISSUE #2174: Automate stable snap updates
When we do a release, we have CI that builds all of the installer packages for us automatically. We need something for snapcraft. Right now its a manual procedure.

Snapcraft.io has a way to integrate with Github - https://snapcraft.io/docs/build-from-github

jwrober:
Closes #2174

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -42,6 +42,7 @@ jobs:
             This PR updates `${{ steps.branch.outputs.branch }}`.
             Close and reopen this pull request to start the CI.
           delete-branch: true
+
   update-archive:
     name: "Update Source Archive"
     runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -72,3 +73,40 @@ jobs:
...
+      - uses: snapcore/action-build@v1
+        id: snapcraft
+        with:
+          path: build
+      - uses: snapcore/action-publish@v1
+        env:
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
+        with:
+          snap: ${{ steps.snapcraft.outputs.snap }}
+          release: beta
```
lmoureaux, line 112: How do you envision the move to a `stable` upload?

.github/workflows/release.yaml
```diff
@@ -72,3 +73,40 @@ jobs:
             Freeciv21-${{github.ref_name}}.tar.gz.sha256
             Freeciv21-${{github.ref_name}}.zip
             Freeciv21-${{github.ref_name}}.zip.sha256
+
+  snapcraft:
+    name: "Snap Package"
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release'
```
lmoureaux, line 80: Maybe we could avoid duplicating the code by moving this `if` to the upload steps?

lmoureaux (commented):
I have a few questions on how this will work in practice. In additions to the ones below, I'd like to make sure that the version number in the uploaded package will be correct.

.github/workflows/release.yaml
jwrober: Change it to stable and "edge" when we start up alphas again

lmoureaux: Should we document this procedure then?

jwrober:
@lmoureaux I think this is ready. Final review at your convenience.

.github/workflows/release.yaml
```diff
@@ -2,15 +2,12 @@ name: Release
 on:
   release:
     types: [published]
-  schedule:
-    - cron: "0 5 * * MON"
```
lmoureaux, line 6: Do not remove the weekly builds.

lmoureaux (changes requested):

Issue #2237, opened 2024-03-26
Interface Options Map Image doesn't do anything


jwrober:
**Describe the bug**
There is a tab in the Interface Options called Map Image.

![image](https://github.com/longturn/freeciv21/assets/23037938/2eb3d201-959f-4a2d-a8ba-b6a386c598cc)

This is actually a Game Options setting `/mapimg` to have the server save an image of the map (in minimap format) on a periodic basis (e.g. every turn).

There are two issues here:
1. The Interface Options tab should be part of the Game Options Dialog as this is a server setting, not a game/client setting.
2. The options, when set, should issue proper commands to the server to set `/mapimg`. On a local game, when I ask to show `/mapimg all`, I get nothing back.

```
jwrober: 'mapimg show all'
/mapimg: No map image definitions.
```

**Expected behavior**
We have options in the right spot that actually do something useful.

**Screenshots**
See above

**Platform and version (please complete the following information):**
 - OS: All
 - Freeciv21 version: 3.1-alpha.5
 - Ruleset/Longturn game (if applicable): N/A

**Additional context**
https://discord.com/channels/378908274113904641/378908274680004619/1219969376711741460

lmoureaux:
Looks like the client has support for most of this, apart from saving the images

jwrober:
My thinking on this one is to 'hide' the options and the tab for now. Then post Qt6 incorporate a full fix via #587 using kconfig.

jwrober:
I have it hidden on my local as an example
![image](https://github.com/longturn/freeciv21/assets/23037938/fd86082b-39ad-41ec-9454-8c1c0774f241)

lmoureaux:
I think a proper fix, preserving functionality while improving the UI, could be along these lines:
1. Remove these options from the general option dialog
2. Have the Game -> Save Map Image menu pop up a dialog with the options
3. In the new dialog, have options for the save path (should it be a folder?) and format (including the current full tileset-based format)
4. Persist options set in the new dialog

I would also like to get rid of the in-house PPM writer and use `QImage` instead. This would make the server depend on QtGui, which we don't necessarily want, so it should remain optional at configure time.

In any case the fix can be a feature for the next version. For now let's just remove the options.

jwrober:
Removing from the milestone so it stays open and we can fix post 3.1 release as per notes above.

PR #2279, opened 2024-04-26
Temporarily remove map image options

lmoureaux:
This PR removes map image options from the configuration dialog, as a temporary solution for #2237.
As a side-effect, any settings previously set by the user will be discarded when saving the config file. I don't consider this an issue since the options were basically useless; anyone changing them would have left them in a random state.


jwrober (approved):
This matches very closely to what I had in my local when testing things out.

PR #2287, opened 2024-05-20
Improve sizing and scrolling of the unit choice popup

### LINKED ISSUE #2285: Can't select units in the last row when there are more than 12 on a tile
**Describe the bug**
When there are more than 12 units on a tile, the units in the last row of the unit chooser popup cannot be selected

**To Reproduce**
1. Stack 20 units outside a city, or in an allied city
2. Click on the tile
3. Scroll down to the bottom
4. Try to click on some unit
5. See it's impossible because the scroll position isn't maintained

**Expected behavior**
It's possible to select any unit on a tile is possible.

 - OS: Linux
 - Freeciv21 version: b7959ddcffa1c9a8ef083694b59a3f7c9173d1fe
 - Ruleset/Longturn game (if applicable): LT82
 - Tileset: RoundSquare

lmoureaux:
The unit selection popup for stacks in the open (and in allied cities) didn't behave very well with many units: its width remained fixed to 4 and it was showing at most 12 units -- quite a small stack for a serious attack. In addition, the last row was in some cases inaccessible due to a scrolling bug (maybe Wayland-specific; #2285).

Increase the base width of the menu to 5 units (a loaded Galleon), and further to 6 when there are more than 25 units. Show up to 36 units (four fully loaded Transports) at once before resorting to scrolling.

Fix the scrolling bug, which was related to scroll events with deltaY=0 being interpreted as scrolling upwards.

Closes #2285.

---

This was likely already a bug in 3.0, so we should consider backporting the fix if there is another dot release.

### DIFF
--- a/client/unitselect.cpp
+++ b/client/unitselect.cpp
@@ -98,20 +98,16 @@ void units_select::create_pixmap()
     delete h_pix;
     h_pix = new QPixmap(item_size.width(), item_size.height());
     h_pix->fill(palette().color(QPalette::HighlightedText));
-    if (unit_count < 5) {
-      row_count = 1;
-      pix = new QPixmap((unit_list.size()) * item_size.width(),
-                        item_size.height());
-    } else if (unit_count < 9) {
-      row_count = 2;
-      pix = new QPixmap(4 * item_size.width(), 2 * item_size.height());
-    } else {
-      row_count = 3;
-      if (unit_count > 12) {
-        more = true;
-      }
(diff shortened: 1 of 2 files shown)


jwrober:
I never get more than 4 columns
image

There are 46 Howitzers on the tile. Other than that seems to work.

lmoureaux:
Hmm, this 12-units layout looks like the old behavior... are you sure the patch is applied?

jwrober:
No one saw anything... :)

image

jwrober (approved):

Issue #1838, opened 2023-02-23
Improve the units widget when clicking on allied/team cities


jwrober:
**Is your feature request related to a problem? Please describe.**
In team games or with allies in FFA games, you can click on a allied/team city to see what units are inside. The current widget that pops up makes it really hard to differentiate different types of units and impossible to determine which units are loaded on a particular transport ship. It seems random as to how the units are placed in the widget (I'm sure it's not random, but feels that way).

Also the widget acts like a "menu" and prevents easy screen shots.

**Describe the solution you'd like**
For one, get rid of the menu thing so its easy to get a screen shot for sharing in a team or with an ally.

I would next like to see the widget turned into a `QTreeWidget` or something so we can organize the units in there better.

1. Organize by Nation and Unit type, maybe alpha order or at least consistent with how units are organized in the city dialog
2. Organize by transporter and then add some indicator or arrangement that makes it super easy to determine which unit has been loaded on the ship. Could also follow from #1836 
3. If the units is "mine" and I want to move it to a different ship (also mine or one of my ally/team mates), I should be able to do it from this widget. Maybe via a drag and drop or a right-click menu option.

**Describe alternatives you've considered**
The usual of "leave it alone" comes to mind, but I think we can do better.

**Additional context**
Team games are FUN!

lmoureaux:
>  For one, get rid of the menu thing so its easy to get a screen shot for sharing in a team or with an ally.

Wayland solves this, not supporting dying tech like X11 if it takes code.

> 1. Organize by Nation and Unit type, maybe alpha order or at least consistent with how units are organized in the city dialog

This is done in #2288.

I don't like the idea of a tree widget as the main UI, because the grid we have is much easier to use when transports are not involved. I could imagine an "expand" button somewhere that would open a more complex dialog for such use cases.

The 3rd item (drag and drop for unit loading) would indeed be cool.

PR #2288, opened 2024-05-21
Sort unit lists shown to the user

### LINKED ISSUE #1836: Improve the UI for transport ships with loaded units in city view
If you have more than one transporter (e.g. a Caravel) inside of a city and that ship also has units loaded on it, it is impossible to tell what unit is on which ship. This is in the present units widget in the city dialog.

With the new city dialog using much more screen real estate, maybe we can expand the widget in some fashion. The first "row" is units stationed in the city, such as defenders. If this scenario arises, then have a second row that shows the transporter to the left and then the units loaded on it to the right with some visual effect so you can easily see who is loaded where. I don't envision a third row for a second transporter, just append it to the right in a pattern and only go to another row when the max width is reached

Could also leave things alone, but add some kind of embellishment/indicator on the transporter and the loaded units. Such as put a `1` on the Caravel and the loaded units for the first ship and a `2` on the second and so forth.

**Additional context**
Recent teams game is giving us lots of ideas :)

lmoureaux:
On top of #2287 - check the second commit.

This patch makes the user interface look more consistent by sorting
units in two contexts:

* In the city dialog
* In the unit picker

The sorting logic is an improved version of code formerly used to sort
units in the city dialog (but that had been lost for some reasion). It
tries to produce a sort order that looks natural and is stable in time.
To do so:

* Loaded units are always sorted right after their transport, so one can
  tell in which transport they are loaded;
* The best defensive units go to the beginning of the list. These tend
  to stay in cities and not move around too much;
* Tie breaker by unit type and nationality ensure similar units are
  grouped together;
* Finally, the unit id is used as a last resort.

I tested this logic in a few contexts and it seems to achieve its goal.
With respect to the previous sorting function, I removed an order based
on unit activities (fortified, fortifying, sentried) because it felt
wrong to have a unit move in the list just because I changed its
activity.

Closes #1836.
See #1838.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -124,6 +124,7 @@ target_sources(
   utils/colorizer.cpp
   utils/improvement_seller.cpp
   utils/unit_quick_menu.cpp
+  utils/unit_utils.cpp
   views/view_cities.cpp
   views/view_cities_data.cpp
   views/view_economics.cpp
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -26,6 +26,7 @@
 // utility
 #include "fc_types.h"
 #include "fcintl.h"
+#include "player.h"
 #include "support.h"
 // common
 #include "citizens.h"
@@ -54,6 +55,7 @@
 #include "unitlist.h"
 #include "utils/improvement_seller.h"
 #include "utils/unit_quick_menu.h"
+#include "utils/unit_utils.h"
 #include "views/view_cities.h" // hIcon
 #include "views/view_map.h"
 #include "views/view_map_common.h"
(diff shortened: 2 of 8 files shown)


jwrober (approved):
Overall this seems to work as described. Wondering if it makes sense to expand the units in city to two fully visible rows when applicable and then give a scroll bar for rows 3+

image

lmoureaux:
The order you show is not what I intended, the planes should come right after the carrier and the defending rifles should be first. This is strange...

I agree that more rows should be shown if possible, but right now I'm not sure how to do that (it's not the same widget as in the unit thing).

jwrober:
Ok. Loaded units "before" vs "after" is a thing. As long as we are consistent either is ok with me. Maybe sorting still needs some work. I'm also ok with tackling the city dialog scaling to two rows later.

jwrober:
Everything looks good except for the situation where a ship with loaded units is present in the city. The dialog shows the ship + loaded units first and then best defender.

lmoureaux:
Ok will check, thanks

lmoureaux:
> Everything looks good except for the situation where a ship with loaded units is present in the city. The dialog shows the ship + loaded units first and then best defender.

I can't reproduce this behavior

PR #2298, opened 2024-06-02
Parity with the original Freeciv nationlist.

Rorr4:
I added three nations that were recently added to the original Freeciv (Florida, Liège, and Palau) modified some nations to include the new ones in their civil war list, and modified some nation's city lists based on recent updates.


lmoureaux:
Superseded by #2320

PR #2305, opened 2024-06-21
Do not rebuild contents when UWT table is updated

### LINKED ISSUE #2296: UWT table on Units view doesn't allow you to scroll down
**Describe the bug**
If you have units waiting for UWT to expire and the list is bigger than the page, then you get a vertical scroll bar. If you try to scroll down, you can't. The game always drops you back to the top.

**To Reproduce**
1. Open a game with lots of units with UWT
2. Click on Units View
3. Try to scroll down to see the bottom one

**Expected behavior**
Scroll bars work as expected.

**Screenshots**

 - OS: Linux
 - Freeciv21 version: 3.1-beta.1
 - Ruleset/Longturn game (if applicable): 

**Additional context**
Came up as part of a conversation with @hugomflavio

blabber:
A timer triggers a periodic update of the UWT table in the unit view. Each update clears the current contents of the table and rebuilds it from scratch. As a result, the list loses its scrolling state and scrolls back up.

This PR changes the behavior to update existing entries and only selectively delete and add entries as needed.

To match units to their entries in the table, a hidden id column has been added to the UWT table.

Closes #2296

### DIFF
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -2,6 +2,7 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  * SPDX-FileCopyrightText: Freeciv21 and Freeciv contributors
  * SPDX-FileCopyrightText: James Robertson <jwrober@gmail.com>
+ * SPDX-FileCopyrightText: Tobias Rehbein <tobias.rehbein@web.de>
  */
 
 /**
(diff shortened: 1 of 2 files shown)


jwrober:
Interesting approach. I'll give it a test.

blabber:
Re-reading my diff, I noticed that I named the new variables using camelCase (the style we are using at my day job), this doesn't follow the style used in the existing code.

I will push an update later today to change these to snake_case, but this will be a purely cosmetic change.

hugomflavio:
Tested #2305. UWT table scrolls nicely without rushing back up.

image

jwrober (approved):

PR #2318, opened 2024-07-07
Fix the units report when there are many units

### LINKED ISSUE #2313: UWT table collapses to one row if enough unit types exist
The UWT table in the units tab is very hard to navigate if there are enough unit types present in the empire:

image

Make the distribution of vertical space between the two tables more even, or allow the user to change their relative height by moving the horizontal bar in the middle.

lmoureaux:
Split the window horizontally instead of vertically, which facilitates scrolling.

Closes #2313.

### DIFF
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -75,7 +75,7 @@ units_view::units_view() : QWidget()
   ui.uwt_widget->setHorizontalHeaderLabels(slist);
   ui.uwt_widget->setSortingEnabled(true);
   ui.uwt_widget->setAlternatingRowColors(true);
-  ui.uwt_label->setText("Units Waiting:");
+  ui.uwt_group->setTitle(_("Units Waiting:"));
 
   // Add shield icon for shield upkeep column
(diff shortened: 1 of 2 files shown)


blabber:
Works as intended, but the layout is a little inconsistent now:
before

I have hacked this together on `view_units.ui`:
after
 
What do you think?

lmoureaux:
I agree that your proposed layout looks better. Does it look ok when no units are waiting?

blabber:
The "units waiting" group is hidden in this case. Feel free to test my proposal (which is a good idea anyways): 38f5dea108bf696f7d0075fb00e84991ae4d8f09

jwrober (changes requested):
see note above

jwrober:
Another alternative idea is to change the view to a tabbed interface with units table on one tab and units waiting on a second. Then each table can use the whole screen.

lmoureaux:
A resizeable split screen can be done with a `QSplitter`. Closing this PR as the required changes go in a vastly different direction.

PR #2323, opened 2024-07-08
Automatically cancel the current activity in goto

### LINKED ISSUE #2272: allow starting a go-to order to automatically cancel previously set order
When the client is set to not clear orders when a unit is selected, pressing G to set a go-to results in an awkward behaviour. First, the cursor turns into the forbidden symbol (i.e. the unit can't move while e.g. sentried). But then, clearing the Sentry order with space does not update the previous go-to order, so the cursor will still show as the round forbidden symbol. Even worse, pressing G again will _stack_ the go-to orders (even if the sentry has been cleared in-between), causing a stacked forbidden symbol (i.e. one must press ESC as many times as one had pressed G to clear them all and reset).

Allow G to automatically cancel S or F. This would be consistent with how pressing S will immediately replace a Fortify command.

Keep the current initial behaviour (i.e. can't go-to if sentried or fortified), but clearing the previous order with space automatically releases the go-to order.

**Additional context**
Note: this also works for workers performing other tasks e.g. irrigating, roading.

image

lmoureaux:
It's common to select sentried or fortified units and initiate a goto. Currently this requires a click on the units to unset any activity they may have. Do this automatically in the path finder.

Closes #2272.

### DIFF
--- a/common/path_finder.cpp
+++ b/common/path_finder.cpp
@@ -101,6 +101,7 @@ bool vertex::comparable(const vertex &other) const
  */
 void vertex::fill_probe(unit &probe) const
 {
+  probe.activity = ACTIVITY_IDLE; // Else it doesn't want to move.
   probe.tile = location;
   probe.transporter = loaded;
   probe.client.transported_by = loaded ? loaded->id : -1;


jwrober (approved):
OMG. One line fixes a really crazy bug! Gotta love fc code.

PR #2328, opened 2024-07-24
Fix units view for many units

### LINKED ISSUE #2313: UWT table collapses to one row if enough unit types exist

blabber:
Put units-table and units-waiting-table into a splitter to facilitate adjusting the ui for many units.

If no units are waiting, the units-waiting-table is hidden and the units-table uses the full available screen estate.

Closes #2313.

### DIFF
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -78,10 +78,14 @@ units_view::units_view() : QWidget()
   ui.uwt_label->setText("Units Waiting:");
 
   // Configure the splitter
-  /* HACK: Why do we need to set the strech factor to 2:1
-   * in order to achieve a split that resembles 1:1? */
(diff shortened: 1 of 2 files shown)


blabber:
Still a draft a draft, until I find the reason for this:

```c
  // Configure the splitter
  /* HACK: Why do we need to set the strech factor to 2:1
   * in order to achieve a split that resembles 1:1? */
  ui.splitter->setStretchFactor(0, 2);
  ui.splitter->setStretchFactor(1, 1);
````

lmoureaux (approved):

jwrober (approved):

PR #2330, opened 2024-07-29
Keep scroll position when reinstantiating science_report

### LINKED ISSUE #2329: Research view loses scroll position
**Describe the bug**
Reported in Discord: 
> In the 3.1.beta.1 I notice that whenever I open the tech tab it starts at the far left/beginning of the tech screen even through if I had just opened it a sec ago and moved the bar over so far. As I recall it would remember where I moved the bar in the stable release?

I confirmed that the older versions did not lose the scroll position.

**To Reproduce**
1. Open research view
2. Scroll somewhere in the tree
3. Return to map and reopen research view
4. See that the research view is scrolled back to origin

**Expected behavior**
When the research view is reopened, the tree should be scrolled to the position where you left the view. 

 - OS: FreeBSD FreeBSD 14.1-RELEASE-p2
 - Freeciv21 version: 3.1-beta.1

blabber:
The `science_report` loses the scroll position when the view is closed. This is because the object is destroyed and therefore loses its state.

This commit adds code to track the scroll position in static fields and restore the state when the `science_report` is reinstantiated.

Closes #2329.

An alternative solution would be to prevent the destruction of `science_report` when the view is closed; this is the behavior in stable.

### DIFF
--- a/client/top_bar.cpp
+++ b/client/top_bar.cpp
@@ -38,9 +38,7 @@
 #include "ratesdlg_g.h"
 #include "tileset/sprite.h"
 #include "top_bar.h"
-#include "views/view_economics.h"
 #include "views/view_map.h"
-#include "views/view_nations.h"
 #include "views/view_research.h"
 #include "views/view_units.h"
 
@@ -449,11 +447,7 @@ void top_bar::addWidget(QWidget *fsw)
  */
 void top_bar_show_map()
 {
-  popdown_units_view();
   popdown_city_dialog();
-  popdown_players_report();
-  popdown_economy_report();
-  popdown_science_report();
   queen()->game_tab_widget->setCurrentIndex(0);
 }
 


blabber:
All views are losing their state (scroll positions etc.), tracking and restoring them is possible, but I think we should take our time to check if we could restore the old behavior from stable (not destroying the instances). I am converting this into a draft PR for now.

lmoureaux:
Notes:
* The position is lost when switching to the map view. Switching between views doesn't affect it
* The city list is magically not reset
* The default scroll location of the science report should probably focus on current research

blabber:
@jwrober, I updated this PR with a version I used to play LT85 until now. This change does not destroy reports/views when the player switches to the map and thus retains their state.

jwrober:
Looks like it needs a `clang-format` or a rebase

jwrober (approved):

blabber:
> * The city list is magically not reset

This is because `popdown_city_report` wasn't called in `top_bar_show_map`.

PR #2332, opened 2024-08-01
Don't access items in empty cache

blabber:
While debugging an issue reported in LT85, I found that in certain circumstances an empty specialist output cache (introduced in 5c44029) is accessed, what crashed the client.

An obvious fix, contained in this PR, is limiting access to non-empty caches. There may be better ways to handle this though.

While here, I applied the same safety measure to the waste level cache, introduced in ecabf5d.

### DIFF
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -2247,9 +2247,12 @@ void add_specialist_output(
     // This is more than just an optimization. For governors that forbid
     // specialists, the cache may not be filled.
     if (count > 0) {
+      // If there is a cache it must not be empty.
+      fc_assert(!pcsoutputs || !pcsoutputs->empty());
+
       output_type_iterate(stat_index)
       {
-        int amount = pcsoutputs && !pcsoutputs->empty()
+        int amount = pcsoutputs
                          ? pcsoutputs->at(sp)[stat_index]
                          : get_specialist_output(pcity, sp, stat_index);
 
@@ -2876,10 +2879,12 @@ void set_city_production(struct city *pcity,
    * on, so shield waste will include shield bonuses. */
   output_type_iterate(o)
   {
+    // If there is a cache it must not be empty.
+    fc_assert(!pcwaste || !pcwaste->empty());
+
     pcity->waste[o] =
         city_waste(pcity, o, pcity->prod[o] * pcity->bonus[o] / 100, nullptr,
-                   gov_centers,
-                   pcwaste && !pcwaste->empty() ? &pcwaste->at(o) : nullptr);
+                   gov_centers, pcwaste ? &pcwaste->at(o) : nullptr);
   }
   output_type_iterate_end;
 
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -2249,7 +2249,7 @@ void add_specialist_output(
     if (count > 0) {
       output_type_iterate(stat_index)
       {
-        int amount = pcsoutputs
+        int amount = pcsoutputs && !pcsoutputs->empty()
                          ? pcsoutputs->at(sp)[stat_index]
                          : get_specialist_output(pcity, sp, stat_index);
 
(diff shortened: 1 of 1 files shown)


jwrober (approved):

jwrober:
@lmoureaux Would you like to review this PR? It looks good to me and seems to fix the issue reported.

blabber:
In particular, I could imagine that a solution along these lines would be better, because it effectively continues to use the cache.

```cs
    output_type_iterate(stat_index)
    {
      int amount = 0;

      if (pcsoutputs &&  !pcsoutputs->empty())
      {
        amount = pcsoutputs->at(sp)[stat_index];
      }

      if (!pcsoutputs)
      {              
         amount = get_specialist_output(pcity, sp, stat_index);
      }

      output[stat_index] += count * amount;
    }
```

However, I was not sure how to interpret the empty cache.

lmoureaux:
Yes, I want to check where the empty cache comes from. Need to check the code again ^^

lmoureaux:
The cache is not created when the governor settings forbid specialists.

Stack trace of the crash:
```
#12 0x000055555585b2b3 in add_specialist_output(city const*, int*, std::vector<std::array<int, 6ul>, std::allocator<std::array<int, 6ul> > > const*) (pcity=0x55555c82c360, output=0x55555c82c484, pcsoutputs=0x55555a02ff10)
    at /home/louis/Personal/freeciv21/common/city.cpp:2249
#13 0x000055555585d4a6 in city_refresh_from_main_map(city*, bool*, std::vector<city*, std::allocator<city*> > const&, std::array<cached_waste, 6ul> const*, std::vector<std::array<int, 6ul>, std::allocator<std::array<int, 6ul> > > const*) (pcity=0x55555c82c360, workers_map=0x55555f476f70, gov_centers=std::vector of length 1, capacity 1 = {...}, pcwaste=0x55555a02ff40, pcsoutputs=0x55555a02ff10) at /home/louis/Personal/freeciv21/common/city.cpp:3050
#14 0x00005555559107d4 in apply_solution(cm_state*, partial_solution const*) (state=0x55555a02fe10, soln=0x55555a02fff8) at /home/louis/Personal/freeciv21/common/aicore/cm.cpp:704
```

lmoureaux:
#2355 addresses the root cause but in a less defensive way.

lmoureaux:
Even after #2355 we may want to keep checking that the cache isn't empty. I'm only mildly optimistic that the issue won't appear in a different code path.

common/city.cpp
```diff
@@ -2247,6 +2247,9 @@ void add_specialist_output(
     // This is more than just an optimization. For governors that forbid
     // specialists, the cache may not be filled.
     if (count > 0) {
+      // If there is a cache it must not be empty.
+      fc_assert(!pcsoutputs || !pcsoutputs->empty());
```
lmoureaux, line 2251: `fc_` asserts aren't fatal by default. Use `fc_assert_action(condition, continue)` to skip this specialist, which is probably an ok fallback behavior:
```suggestion
      fc_assert_action(!pcsoutputs || !pcsoutputs->empty(), continue);
```

common/city.cpp
```diff
@@ -2876,6 +2879,9 @@ void set_city_production(struct city *pcity,
    * on, so shield waste will include shield bonuses. */
   output_type_iterate(o)
   {
+    // If there is a cache it must not be empty.
+    fc_assert(!pcwaste || !pcwaste->empty());
```
lmoureaux, line 2883: This should probably go before the loop to avoid getting 6 messages every time. I think returning if the assert fails is fine, so use `fc_assert_ret(condition)`.

lmoureaux (changes requested):

common/city.cpp
blabber: I implemented another solution, falling back to the old behaviour by setting the cache to `nullptr` if the assertion triggers. This makes sure, that the game will behave correctly. What do you think?

common/city.cpp
blabber: Same here.

lmoureaux (approved):
That's a nicer solution!

PR #2334, opened 2024-08-06
Switch server-side mapimg to Qt

lmoureaux:
The server-side map image code was using a handcrafted ppm writer. This produces huge files. Since we have Qt with many built-in image writers, use that instead. This also changes the default to png since it's just a better format; and replaces the black background with transparent.


cmake/FreecivInstall.cmake
```diff
@@ -22,12 +22,10 @@ install(
   ${CMAKE_SOURCE_DIR}/dist/licenses/FTL.txt
   ${CMAKE_SOURCE_DIR}/dist/licenses/GPL2.txt
   ${CMAKE_SOURCE_DIR}/dist/licenses/GPL3.txt
-  ${CMAKE_SOURCE_DIR}/dist/licenses/IMAGEMAGICK.txt
```
jwrober, line 25: I think we should delete this file too.

jwrober (approved):

jwrober:
You also need to update the github action workflow files and readme to remove the dependency.

jwrober (changes requested):
See comments

cmake/FreecivInstall.cmake
lmoureaux: My bad, I missed that there were two licenses for ImageMagick

lmoureaux:
> You also need to update the github action workflow files and readme to remove the dependency.

I don't see it there - we never supported building with ImageMagick, I think the mention in the docs was a leftover from upstream.

jwrober:
> > You also need to update the github action workflow files and readme to remove the dependency.
> 
> I don't see it there - we never supported building with ImageMagick, I think the mention in the docs was a leftover from upstream.

I am referring to the installation of a library we no long need that is referred to in the README and the gh actions build files. You removed it from docs, but it still remains in other places.

lmoureaux:
I don't understand... this is the full list of packages we install on Ubuntu:
https://github.com/longturn/freeciv21/blob/572b15e8d00bc3446155026b4430a8651e0211a6/.github/workflows/build.yaml#L43-L55

And on Windows:
https://github.com/longturn/freeciv21/blob/572b15e8d00bc3446155026b4430a8651e0211a6/.github/workflows/build.yaml#L140-L151

What do you want me to remove?

jwrober:
Ok I stand corrected. I guess we had a dep install in docs that we didn't do anywhere else.

jwrober (approved):

PR #2336, opened 2024-08-06
In-depth map generator documentation

lmoureaux:
This is a big one, but hopefully an interesting read. I went through the code of the map generators (except FAIR, it's awful) and tried to explain how they work in words. Some of this information could make its way into other parts of the docs, but this is left for later.

@hugomflavio @wieder-fi you may find this interesting.

PR #2345, opened 2024-08-11
Split mapgen.cpp

lmoureaux:
Based on #2336, docs will need to be updated for renamed functions.

This PR splits out the very large `mapgen.cpp` file. It creates one file for ISLAND and one for FAIR, reducing the `mapgen.cpp` to 1500 lines from 4000 previously. I also cleaned some dust by renaming functions and increasing consistency.

My testing plan has been to generate maps with each generator while I was touching it.

CodeFactor is unhappy because I moved complex functions around.


lmoureaux:
CodeFactor is unhappy about splitting one complex file into two shorter but still complex files.

Other than that this should be ready.

jwrober (approved):
I tested a collection of map types based on my recipe docs. Seems to work for me.

PR #2359, opened 2024-09-04
improved battle messages (closes #2358)

### LINKED ISSUE #2358: Show reports about no-kill combat in Messages/Chat
Currently only combat that ends with a kill gets reported. When using limited combat rounds, probably more than half combat events don't end up with a kill, which means they don't get reported, which means the player can find out about them only if he kept track of all his units' health which is highly unlikely.

So, the request: report the results of no-kill combats the same way as with combats that end with a kill: unit #, location, lost HP for both attacker and defender.

hugomflavio:


common/featured_text.cpp
```diff
@@ -1208,3 +1207,21 @@ const char *unit_firepower_if_not_one(int firepower)
...
+   overwrites the previous.
+ */
+const char *unit_n_stack_kills(int unitcount)
+{
+  static char buf[MAX_LEN_LINK];
+
+  if (unitcount > 2) {
+    fc_snprintf(buf, sizeof(buf), _("%d units were"), unitcount = 1);
+  } else {
+    fc_snprintf(buf, sizeof(buf), _("unit was"));
```
lmoureaux, line 1224: You most likely do not want to implement [the rules for Arabic and French](https://www.unicode.org/cldr/charts/43/supplemental/language_plural_rules.html). Fortunately [the `PL_` macro](https://longturn.readthedocs.io/en/latest/Coding/internationalization.html) will do it for you.

common/featured_text.cpp
```diff
@@ -1208,3 +1207,21 @@ const char *unit_firepower_if_not_one(int firepower)
   }
   return buf;
 }
+
+/**
+   Get string of number of extra units killed by attacker
+   when a stack is taken down.
+   N.B.: The returned string is static, so every call to this function
+   overwrites the previous.
```
lmoureaux, line 1215: Per coding style:
```suggestion
 * Get string of number of extra units killed by attacker
 * when a stack is taken down.
 * \note The returned string is static, so every call to this function
 * overwrites the previous.
```

common/events.cpp
```diff
@@ -156,10 +156,12 @@ static struct {
     GEN_EV(E_TREATY_PEACE, E_S_TREATY, N_("Peace")),
     GEN_EV(E_TREATY_SHARED_VISION, E_S_TREATY, N_("Shared Vision")),
     GEN_EV(E_UNIT_LOST_ATT, E_S_UNIT, N_("Attack Failed")),
+    GEN_EV(E_UNIT_TIE_ATT, E_S_UNIT, N_("Attack Tied")),
     GEN_EV(E_UNIT_WIN_ATT, E_S_UNIT, N_("Attack Succeeded")),
     GEN_EV(E_UNIT_BUY, E_S_UNIT, N_("Bought")),
     GEN_EV(E_UNIT_BUILT, E_S_UNIT, N_("Built")),
     GEN_EV(E_UNIT_LOST_DEF, E_S_UNIT, N_("Defender Destroyed")),
+    GEN_EV(E_UNIT_TIE_DEF, E_S_UNIT, N_("Defender Tied")),
```
lmoureaux, line 164: Is the original table sorted in the same way as the list in `events.h`? If so, keep the same order for both.

common/events.h
```diff
@@ -165,6 +165,10 @@
 #define SPECENUM_VALUE133 E_UNIT_ACTION_TARGET_OTHER
 #define SPECENUM_VALUE134 E_UNIT_ACTION_TARGET_HOSTILE
 #define SPECENUM_VALUE135 E_UNIT_WAKE
+// --
```
lmoureaux, line 168:
```suggestion
// Combat without winner (Combat_Rounds)
```

common/featured_text.h
```diff
@@ -238,3 +238,5 @@ const char *unit_veteran_level_string(const struct unit *punit);
 const char *unit_achieved_rank_string(const struct unit *punit);
 const char *unit_tired_attack_string(const struct unit *punit);
 const char *unit_firepower_if_not_one(int firepower);
+const char *unit_firepower_if_not_one(int firepower);
```
lmoureaux, line 241: Remove duplicated line
```suggestion
```

server/unittools.cpp
```diff
@@ -2559,18 +2557,9 @@ void kill_unit(struct unit *pkiller, struct unit *punit, bool vet)
...
-                        unitcount - 1),
-                    pkiller_link, nation_adjective_for_player(pvictim),
-                    punit_link, unitcount - 1);
-    }
-
-    if (vet) {
-      notify_unit_experience(pkiller);
+                    /* TRANS: "Another unit was eliminated by..." */
+                    "Another %s eliminated by our attacking %s!",
+                    unit_n_stack_kills(unitcount), punit_link);
```
lmoureaux, line 2562: This is difficult to translate, better a `PL_()`:
* `unitcount==1`: Un autre unité a été éliminée dans l'attaque de notre Croiseur!
* `unitcount>1`: **2 autres** unité**s** **ont** été éliminée**s** dans l'attaque de notre Croiseur!

It is generally advisable not to break sentences like this because this tends to assume English grammar.

lmoureaux (commented):
Since this changes the network protocol, we need to check that the behavior of an old (3.0) client connecting to a new server. It is understood that the old client will not behave optimally, but it shouldn't crash or complain too loudly.

I also have comments on the code below. I haven't checked the new logic yet.

server/unittools.cpp
hugomflavio: I used PL_ elsewhere, but I couldn't quite figure out how to make it work here. Any ideas?

common/featured_text.cpp
hugomflavio: Happy to get rid of this if we have a way to _PL the line you mentioned later. My brain isn't running at very high capacity after all this cpp editing... :sweat_smile:

common/featured_text.cpp
lmoureaux: Indeed, most comments haven't been updated...

common/events.h
hugomflavio: Should these specenum values all pushed down by 1 (or 2) so the new TIE ones can live next to their WIN and LOST counterparts?

server/unittools.cpp
lmoureaux: Hmm I suppose the issue that the `%d` isn't always there. There's [a trick](https://stackoverflow.com/a/16765923) for that (first 3 lines, the rest is untranslatable), which is to move the optional parameter to the end:
```cpp
notify_player(...
    PL_("Another unit eliminated by our attacking %s!",
        "Another %2$d units eliminated by our attacking %1$s!",
         unitcount - 1),
    punit_link, unitcount - 1);
```

common/events.h
lmoureaux: This would screw up the icons for all changed events in old clients, so... no :sweat_smile:

common/events.cpp
hugomflavio: they are not in the same order as events.h

server/unithand.cpp
```diff
@@ -4022,145 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
+    if (pdefender->veteran == old_defender_vet ? 0 : 1) {
+      notify_unit_experience(pdefender);
+    }
+
+    wipe_unit(punit, ULR_KILLED, unit_owner(pdefender));
+    return true;
+  }
+  // this function should have not passed through this line of code
+  fc_assert_ret_val(false, true);
   return true;
```
lmoureaux, line 4230: You can remove the explicit `return true`, it's done in `fc_assert_ret_val`.
```suggestion
```

server/unittools.cpp
```diff
@@ -2586,53 +2578,41 @@ void kill_unit(struct unit *pkiller, struct unit *punit, bool vet)
...
-                punit_link, others, nation_adjective_for_player(pvictor),
-                pkiller_link);
-          }
+          notify_player(
+              player_by_number(i), ptile, E_UNIT_LOST_DEF, ftc_server,
+              /* TRANS: "Musketeers and 3 other units lost to
+               * an attack from the Polish Destroyer."
+               * (only happens with at least 2 other units) */
+              PL_("%d other unit lost when the %s %s defeated your %s.",
+                  "%d other units lost when the %s %s defeated your %s.",
```
lmoureaux, line 2604: "3 other units" or "3 more units"?

server/unittools.cpp
```diff
@@ -2586,53 +2578,41 @@ void kill_unit(struct unit *pkiller, struct unit *punit, bool vet)
...
+                  others),
+              others, nation_adjective_for_player(pvictor), pkiller_link,
+              punit_link);
```
lmoureaux, line 2607: Originally, you wanted to have a different message when only one unit is killed. The `%d` can be removed from the first message if it is the *last* argument of `notify_player`. The `%n$x` syntax can be used to reorder arguments when building the string:
```suggestion
              PL_("One other unit lost when the %s %s defeated your %s.",
                  "%4$d other units lost when the %s %s defeated your %s.",
                  others),
              punit_link, nation_adjective_for_player(pvictor), pkiller_link,
              others);
```
Now `punit_link` is number 1, `nation_adjective_for_player(pvictor)` is number 2, `pkiller_link` is number 3, and `others` is number 4.

When formatting the first string, arguments 1 to 3 will be used and 4 will be ignored. When formatting the second strings, all 4 arguments will be used.

server/unittools.cpp
hugomflavio: I think both work here? Other sounds better than more, to me. but I'm not native :)

server/unithand.cpp
```diff
@@ -4022,146 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
+        (float) def_power / POWER_FACTOR, def_hp_start - pdefender->hp,
+        pdefender->hp, nation_adjective_for_player(unit_owner(punit)),
+        attacker_vet, attacker_link, punit->id, attacker_fp,
+        (float) att_power / POWER_FACTOR, att_hp_start - punit->hp,
+        punit->hp,
+        (punit->veteran - old_unit_vet) ? unit_achieved_rank_string(punit)
+                                        : "");
+
+    // notify defender of promotion
+    if (pdefender->veteran == old_defender_vet ? 0 : 1) {
```
lmoureaux, line 4083:
```suggestion
    if (pdefender->veteran != old_defender_vet) {
```

server/unithand.cpp
```diff
@@ -4022,146 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
-        nation_adjective_for_player(unit_owner(ploser)), attacker_vet,
-        loser_link, punit->id, attacker_fp, (float) att_power / POWER_FACTOR,
-        att_hp_start);
+        pdefender->hp,
+        (pdefender->veteran - old_defender_vet)
+            ? unit_achieved_rank_string(pdefender)
+            : "");
+
+    // notify attacker of promotion
+    if (punit->veteran == old_unit_vet ? 0 : 1) {
```
lmoureaux, line 4058:
```suggestion
    if (punit->veteran != old_unit_vet) {
```

server/unithand.cpp
```diff
@@ -4022,146 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
+        (float) att_power / POWER_FACTOR, att_hp_start - punit->hp,
+        punit->hp, nation_adjective_for_player(unit_owner(pdefender)),
+        defender_vet, defender_link, pdefender->id, defender_fp,
         (float) def_power / POWER_FACTOR, def_hp_start - pdefender->hp,
-        pdefender->hp, attacker_tired,
-        nation_adjective_for_player(unit_owner(ploser)), attacker_vet,
-        loser_link, punit->id, attacker_fp, (float) att_power / POWER_FACTOR,
-        att_hp_start);
+        pdefender->hp,
+        (pdefender->veteran - old_defender_vet)
```
lmoureaux, line 4053:
```suggestion
        (pdefender->veteran != old_defender_vet)
```

server/unithand.cpp
```diff
@@ -4022,146 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
+        _("Your %s %s [id:%d %sA:%.1f, lost %d HP, %d HP remaining]"
+          " was attacked by the %s %s %s [id:%d %sD:%.1f, lost %d"
+          " HP, %d HP remaining%s]!"),
+        defender_vet, defender_link, pdefender->id, defender_fp,
+        (float) def_power / POWER_FACTOR, def_hp_start - pdefender->hp,
+        pdefender->hp, nation_adjective_for_player(unit_owner(punit)),
+        attacker_vet, attacker_link, punit->id, attacker_fp,
+        (float) att_power / POWER_FACTOR, att_hp_start - punit->hp,
+        punit->hp,
+        (punit->veteran - old_unit_vet) ? unit_achieved_rank_string(punit)
```
lmoureaux, line 4079:
```suggestion
        (punit->veteran != old_unit_vet) ? unit_achieved_rank_string(punit)
```

server/unithand.cpp
```diff
@@ -4022,146 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
-          "has %d remaining] succeeded against the %s %s %s "
-          "[id:%d HP:%d]."),
-        attacker_vet, winner_link, punit->id, attacker_fp, attacker_tired,
-        (float) att_power / POWER_FACTOR, att_hp_start - pwinner->hp,
-        pwinner->hp, nation_adjective_for_player(unit_owner(pdefender)),
-        defender_vet, loser_link, pdefender->id, def_hp_start);
+                  attacker_vet, attacker_link, punit->id, attacker_fp,
+                  attacker_tired, (float) att_power / POWER_FACTOR,
+                  att_hp_start - punit->hp, punit->hp,
+                  (punit->veteran - old_unit_vet)
```
lmoureaux, line 4137:
```suggestion
                  (punit->veteran != old_unit_vet)
```

server/unithand.cpp
```diff
@@ -4022,146 +4020,213 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
+       * air base with multiple defenders and unstacked combat). Note that
+       * this could mean capturing (or destroying) a city. */
+      if (fc_rand(100) < game.server.occupychance
+          && !is_non_allied_unit_tile(def_tile, pplayer)) {
+        occupy_move(punit, def_tile);
+      }
+      // The attacker may have died for many reasons
+      // 2024-09-04 HF: Not very sure why this is needed here
+      // but not in the other outcome cases? Is send_combat()
+      // doing this as well?
```
lmoureaux, line 4164: For instance if there is a hut on the target tile.

server/unittools.cpp
```diff
@@ -2586,53 +2578,41 @@ void kill_unit(struct unit *pkiller, struct unit *punit, bool vet)
           notify_player(player_by_number(i), ptile, E_UNIT_LOST_DEF,
                         ftc_server,
                         // TRANS: "Cannon ... the Polish Destroyer."
-                        _("%s lost to an attack by the %s %s."), punit_link,
-                        nation_adjective_for_player(pvictor), pkiller_link);
+                        _("%s lost when the %s %s defeated your %s."),
```
lmoureaux, line 2581: > Warriors lost when the Celtiberian Frigate defeated our Warriors.

--> this sounds like two different warriors.

server/unittools.cpp
```diff
@@ -2586,53 +2578,41 @@ void kill_unit(struct unit *pkiller, struct unit *punit, bool vet)
...
-                punit_link, others, nation_adjective_for_player(pvictor),
-                pkiller_link);
-          }
+          notify_player(
+              player_by_number(i), ptile, E_UNIT_LOST_DEF, ftc_server,
+              /* TRANS: "Musketeers and 3 other units lost to
+               * an attack from the Polish Destroyer."
+               * (only happens with at least 2 other units) */
+              PL_("One other unit lost when the %s %s defeated your %s.",
+                  "%4$d other units lost when the %s %s defeated your %s.",
```
lmoureaux, line 2604: Did you test this change? I'm not fully confident my proposal works :sweat:

server/unittools.cpp
```diff
@@ -33,6 +33,7 @@
 #include "city.h"
 #include "combat.h"
 #include "events.h"
+#include "featured_text.h"
```
lmoureaux, line 36: Is this `include` still needed?

server/unithand.cpp
hugomflavio: I really wrote that.. ouch... lol

server/unittools.cpp
hugomflavio: hm... Should this notify_player exist at all? If your unit loss the battle and you only have one unit on that tile, you already received a message about this at this stage. But I didn't see this behaviour in the test-game... I'll try again.

server/unittools.cpp
hugomflavio: nope

server/unittools.cpp
hugomflavio: tested. had to fix it :D

server/unithand.cpp
hugomflavio: updated comment

server/unithand.cpp
hugomflavio: fixed in latest commit

lmoureaux (commented):
One last suggestion

server/unithand.cpp
```diff
@@ -4022,146 +4020,324 @@ static bool do_attack(struct unit *punit, struct tile *def_tile,
...
+    // scorn the attacker if defender lost no hp
+    if (def_hp_start == pdefender->hp) {
+      // notify attacker of battle tie
+      notify_player(
+          unit_owner(punit), unit_tile(punit), E_UNIT_TIE_ATT, ftc_server,
+          /* TRANS: "Your green Warrior [id:100 ...A:1.0, lost 3 HP,
+           * 7 HP remaining] tried to attack the Greek Polish Archer [id:200
+           * ...D:8.0, lost 5 HP, 5 HP remaining]!"*/
+          _("Your %s %s [id:%d %sA:%.1f, lost %d HP, %d HP remaining]"
+            " tried to attack the %s %s %s [id:%d %sD:%.1f, lost %d"
```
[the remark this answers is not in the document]
hugomflavio: done :)

lmoureaux (approved):

PR #2360, opened 2024-09-04
Bump actions/upload-artifact to 4.1.7

lmoureaux:
This is needed for compatibility with the latest download-artifact.


lmoureaux:
Has to be in the default branch to work, but quite confident it will.

PR #2362, opened 2024-09-05
Bugfix/bombard info

hugomflavio:
Split from the previous PR.


server/unithand.cpp
```diff
@@ -3668,6 +3670,8 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
 {
   struct player *pplayer = unit_owner(punit);
   struct city *pcity = tile_city(ptile);
+  // create QMap before loop
+  std::set<struct player *> players_to_notify;
```
hugomflavio, line 3674: @lmoureaux confirm: this comment is wrong, right?

lmoureaux: Right, it's a `std::set`. And it's quite obvious that you're creating it so for me it's not needed.

hugomflavio: sounds good. removed

common/networking/packets.def
```diff
@@ -1083,6 +1083,7 @@ PACKET_UNIT_COMBAT_INFO = 65; sc, lsend
   HP defender_hp;
   BOOL make_att_veteran;
   BOOL make_def_veteran;
+  BOOL bombard;
```
lmoureaux, line 1086: Adding this field breaks compatibility between old clients (that won't have it) and new servers (that will), or old servers and new clients. It may technically work because of how boolean fields are encoded into the packets, but I'm not sure.

There is a mechanism to add or remove fields without breaking network compat. This is done via the `add-cap` or `remove-cap` flags and a capability string.

hugomflavio: Ok, I'll ask you about this once we conclude our current discord discussion :+1:

jwrober:
Discord convo starts around here - https://discord.com/channels/378908274113904641/912500712833974322/1282110071924461580

common/networking/packets.def
hugomflavio: Removing these changes as they are no longer needed

server/unithand.cpp
```diff
@@ -3772,6 +3761,24 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
   }
   unit_list_iterate_safe_end;
 
+  // Notify the players hit by the bombardment
+  for (auto player_to_notify : players_to_notify) {
+    notify_player(
+        player_to_notify, ptile, E_UNIT_BOMB_DEF, ftc_server,
+        /* TRANS: Your units in [tile] have been bombarded by the French
```
jwrober, line 3768: 'in' vs 'on'? Your units **on** [tile] have been bombarded by the [Nation] [Unit]. Unless the tile is a city, then "in" works, but [tile] would be a city name and not a map address.

server/unithand.cpp
```diff
@@ -3772,6 +3761,24 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
...
+           Bomber.*/
+        _("Your units in %s were bombarded by a %s %s %s [id:%d]."),
```
jwrober, line 3770: Same here - 'in' vs 'on'

server/unithand.cpp
```diff
@@ -3772,6 +3761,24 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
...
+        tile_link(ptile), nation_adjective_for_player(pplayer),
+        unit_veteran_level_string(punit), unit_name_translation(punit),
+        punit->id);
+  }
+  // Notify the player that bombarded
+  notify_player(pplayer, ptile, E_UNIT_BOMB_ATT, ftc_server,
+                /* TRANS: Your Bomber bombarded [coords?].*/
+                _("Your %s %s [id:%d] bombarded the units at %s."),
```
jwrober, line 3778: or... do the other two above as 'at' for the tile coords?

jwrober (commented):
Some grammar thoughts.

server/unithand.cpp
hugomflavio: changed to "at"

server/unithand.cpp
hugomflavio: changed to "at"

server/unithand.cpp
hugomflavio: changed to "at"

jwrober (approved):

PR #2364, opened 2024-09-07
Use @v4 for actions/{upload,download}-artifact

lmoureaux:
This will pick the latest available v4 for both actions without us needing to care about exact version numbers. Should fix the clang-tidy action.

The changes will only take effect once merged in `master`.

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -63,7 +63,7 @@ jobs:
             | ansifilter \
             | ./codacy-clang-tidy >report.json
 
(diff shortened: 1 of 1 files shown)


jwrober:
`clang-tidy` doesn't like this one it seems

jwrober:
Same for `stable` branch?

lmoureaux:
> clang-tidy doesn't like this one it seems

The `clang-tidy` action is always running from the `master` branch for security reasons (it has access to secrets). Since `master` hasn't been updated, it's still failing.

> Same for stable branch?

Sure, easy to backport

jwrober (approved):

PR #2378, opened 2024-09-09
Update compile installation for Debian/Ubuntu

jwrober:
No related issue, setting up a dev env for snapcraft helped me find the error

### DIFF
--- a/docs/Getting/compile.rst
+++ b/docs/Getting/compile.rst
@@ -164,6 +164,7 @@ following commands:
   $ sudo apt install git \
      cmake \
      ninja-build \
+     g++ \
      python3 \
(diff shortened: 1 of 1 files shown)


blabber (commented):
Just to make sure: do you really need both, g++ and clang? One is the GNU C++ compiler, the other  contains the clang C++ compiler. I'd expect that I e of them should be enough (probably clang).

blabber (approved):
Just to make sure: do you really need both, g++ and clang? One is the GNU C++ compiler, the other  contains the clang C++ compiler. I'd expect that I e of them should be enough (probably clang).

Ah, I see. It's clang-format, not clang...

PR #2384, opened 2024-09-10
Fix research refresh button

### LINKED ISSUE #1753: Make it possible to forcibly research a tech that is instantly researchable after TC due to tech leak

jwrober:
Closes #1753

Marking as draft for now. Need an event to test it.

### DIFF
--- a/client/views/view_research.cpp
+++ b/client/views/view_research.cpp
@@ -538,6 +538,7 @@ void science_report::current_tech_changed(int changed_index)
   if (researching_combo->hasFocus()) {
     if (can_client_issue_orders()) {
       dsend_packet_player_research(&client.conn, qvar.toInt());
+      dsend_packet_player_tech_goal(&client.conn, qvar.toInt());
     }
   }
 }
--- a/client/views/view_research.cpp
+++ b/client/views/view_research.cpp
@@ -538,7 +538,6 @@ void science_report::current_tech_changed(int changed_index)
   if (researching_combo->hasFocus()) {
     if (can_client_issue_orders()) {
       dsend_packet_player_research(&client.conn, qvar.toInt());
-      dsend_packet_player_tech_goal(&client.conn, qvar.toInt());
     }
   }
 }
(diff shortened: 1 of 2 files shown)


lmoureaux:
I doubt this will work, `handle_player_tech_goal` in `server/techtools.cpp` doesn't seem to set the current tech.

Moreover, this would wipe out any tech goal previously set by the user.

jwrober:
Ouch. Well glad I pushed this draft before I blew out a turn giving a test.  Ideas on what to try next?

lmoureaux:
See my comment on the issue. Try removing this line... but you need to set up a local test :)

jwrober:
That has been my biggest problem. I have not been able to get a local save with the "problem" in it to test locally.

lmoureaux:
[save.zip](https://github.com/user-attachments/files/16955683/save.zip)

jwrober:
I never thought about hacking a save file. Nice. The suggestion works great.

lmoureaux (approved):

Issue #2351, opened 2024-08-28
Game crashes when unit enters a city (Sim ruleset)


Corbeau-1:
To reproduce the bug: 

1. Load the savegame with the ruleset provided, both in the attachment below.
2. /take Fidel
3. Find the Raider unit # 1163, just west of main Cuban territory, sitting south of Venetian city of Famagosta
4. Attack/enter Famagosta (occupychance is set to 100)

UPDATE: it seems to happen in **some** instances where there is no city present, just combat. A wild guess: it may have something to do with limited combat rounds and entry into tile after victory.

Using Windows installation 3.1-beta.1.

(updated attachment below)

lmoureaux:
The ruleset doesn't load:
```
[critical] freeciv21-server (/home/louis/Personal/freeciv21/common/requirements.cpp:1086) - Invalid requirement DiplRel | Local |  | present | Is Foreign: bad type or name
[critical] freeciv21-server (/home/louis/Personal/freeciv21/server/ruleset.cpp:453) - "/tmp/abugabug/SIM30/game.ruleset" [actionenabler_investigate_city] has invalid or unknown req: "DiplRel" "Is Foreign".
```

lmoureaux:
Managed to start the game after upgrading the ruleset with `freeciv21-ruleup`. Couldn't reproduce the crash so far, the city gets wiped out and the raider gets promoted.

lmoureaux:
ASan sees it:
```
=================================================================
==2154558==ERROR: AddressSanitizer: heap-use-after-free on address 0x61e00006e4e8 at pc 0x650a0c4aa898 bp 0x7ffd46e4e000 sp 0x7ffd46e4dff0
READ of size 4 at 0x61e00006e4e8 thread T0
    #0 0x650a0c4aa897 in governor::run() /home/louis/Personal/freeciv21/client/governor.cpp:168
    #1 0x650a0c2a9930 in governor::unfreeze() /home/louis/Personal/freeciv21/client/governor.h:26
    #2 0x650a0c2a9930 in input_from_server(QIODevice*) /home/louis/Personal/freeciv21/client/clinet.cpp:325
    #3 0x650a0c445c0f in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (fc_client::*)()>::call(void (fc_client::*)(), fc_client*, void**) /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:152
[...]

PR #2385, opened 2024-09-10
Remove dangling city pointers from the governor code

lmoureaux:
The governor code was using pointers to keep track of cities to be removed or updated. These pointers would sometimes become invalid when the city was lost or otherwise conquered, causing use-after-free errors.

This fixes the first version of #2351.


lmoureaux:
Likely want to backport this.

jwrober (approved):

PR #2386, opened 2024-09-10
Cleanup two divide-by-zero issues from Coverity

jwrober:
No related issue.

CID 1435482, 1435595

### DIFF
--- a/ai/default/aiunit.cpp
+++ b/ai/default/aiunit.cpp
@@ -641,7 +641,8 @@ static void dai_military_bodyguard(struct ai_type *ait,
       int unit_mv_rate = unit_move_rate(punit);
       fc_assert_ret_msg(unit_mv_rate, "div by zero");
       if (me2goal < me2them
-          || (me2goal / unit_mv_rate < them2goal / unit_move_rate(aunit)
+          || (me2goal / unit_mv_rate
+                  < them2goal / std::min(1, unit_move_rate(aunit))
               && me2goal / unit_mv_rate < me2them / unit_mv_rate
               && unit_mv_rate > unit_move_rate(aunit))) {
         ptile = aunit->goto_tile;
@@ -794,7 +795,7 @@ int look_for_charge(struct ai_type *ait, struct player *pplayer,
         continue;
       }
       // Reduce want based on move cost.
-      def >>= move_cost / (2 * unit_move_rate(punit));
+      def >>= move_cost / (2 * std::min(1, unit_move_rate(punit)));
       if (def > best_def && ai_fuzzy(pplayer, true)) {
         *acity = pcity;
         *aunit = nullptr;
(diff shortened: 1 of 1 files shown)


ai/default/aiunit.cpp
```diff
@@ -641,7 +641,8 @@ static void dai_military_bodyguard(struct ai_type *ait,
       int unit_mv_rate = unit_move_rate(punit);
       fc_assert_ret_msg(unit_mv_rate, "div by zero");
       if (me2goal < me2them
-          || (me2goal / unit_mv_rate < them2goal / unit_move_rate(aunit)
+          || (me2goal / unit_mv_rate
+                  < them2goal / std::min(1, unit_move_rate(aunit))
               && me2goal / unit_mv_rate < me2them / unit_mv_rate
               && unit_mv_rate > unit_move_rate(aunit))) {
```
lmoureaux, line 647: We have two techniques to protect against zero `unit_move_rate` now... Can we keep using the same?
```cpp
      int punit_mv_rate = unit_move_rate(punit);
      int aunit_mv_rate = unit_move_rate(aunit);
      fc_assert_ret_msg(punit_mv_rate, "div by zero");
      fc_assert_ret_msg(aunit_mv_rate, "div by zero");
      if (me2goal < me2them
          || (me2goal / punit_mv_rate < them2goal / aunit_move_rate
              && me2goal / unit_mv_rate < me2them / punit_mv_rate
              && unit_mv_rate > aunit_move_rate)) {
```

ai/default/aiunit.cpp
```diff
@@ -640,6 +640,8 @@ static void dai_military_bodyguard(struct ai_type *ait,
       int them2goal = real_map_distance(unit_tile(aunit), aunit->goto_tile);
       int unit_mv_rate = unit_move_rate(punit);
       fc_assert_ret_msg(unit_mv_rate, "div by zero");
+      fc_assert_ret_msg(punit, "div by zero");
+      fc_assert_ret_msg(aunit, "div by zero");
```
lmoureaux, line 644: This is not sufficient, even existing units can have a move rate of 0 if `min_speed` is 0 in `units.ruleset`. Consider the following addition to the classic ruleset after setting `min_speed = 0` for Land units:
```ini
; Units get stuck in forests
[effect_elves_catch_units]
type    = "Move_Bonus"
value	= -1
reqs	=
    { "type", "name", "range"
      "Terrain", "Forest", "Local"
    }
```
Now try to play this from the beginning... (and notice the many bugs it exposes)

=> you need to check the move rate as is already done for `punit`

lmoureaux:
Pushed a commit using the same technique for `aunit` and `punit`. This makes the div by zero truly impossible by checking beforehand.

We could also have done some math and used `a/b < c/d  <=>  a*d < b*c` (for positive `b` and `d`), but this has different rounding which might be important.

jwrober:
Thanks for the help on this one.

Issue #2365, opened 2024-09-07
Autoscroll research tab to current tech


blabber:
**Is your feature request related to a problem? Please describe.**
When opening the research tab, the user is probably interested in the tech tree around the currently active tech. At the moment the user has to scroll the tech tree manually.

**Describe the solution you'd like**
When the research tab is opened the first time in a session, autoscroll the tech tree to make sure the current tech visible.

**Describe alternatives you've considered**
* Leave things as they are now
*  Autoscroll every time the research tab is opened; this might be annoying if you are interested in other parts of the tech tree

**Additional context**
This was brought up by @lmoureaux in #2329

blabber:
I implemented another alternative: I added small buttons next to the tech goal and current tech combo boxes. When one of these buttons is pressed, the corresponding tech is scrolled into view.

blabber:
Closed by #2391

PR #2391, opened 2024-09-17
Implement autoscroll for research tree

blabber:
Allows the user to conveniently scroll the technology tree to the currently researched technology or the technology goal. A button next to the corresponding combo box is available to the user for this purpose.

In the absence of a suitable icon, the buttons currently have the mnemonic label “L” for “Locate”.


blabber:
The functionality is done, but I still want to use an icon for the QToolButtons. If someone who can handle Inkscape better and is a bit more talented than me would like to provide an icon, I would be grateful :)

blabber:
I added a slightly modified `crosshair-light` icon from Phosphor Icons. I guess this should be ready for review now.

blabber:
Closes #2365

lmoureaux:
It works, but the icons are a bit blurry:
image

This is because the icon is 256x256px and the line width is 12px. Scaled down to 16x16, this give a line width of 0.75px.

I came up with the following as a more pixel-perfect version on a 16x16 baseline:
```svg

<?xml version="1.0" encoding="UTF-8"?>
<svg width="16" height="16" version="1.1" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
 <circle cx="8.5" cy="8.5" r="5.85" fill="none" stroke="#000" stroke-width="1.3"/>
 <g fill="#000">
  <circle cx="8.5" cy="8.5" r="2"/>
  <rect x="11.5" y="8" width="4.5" height="1"/>
  <rect x="1" y="8" width="4.5" height="1"/>
  <rect x="8" y="11.5" width="1" height="4.5"/>
  <rect x="8" y="1" width="1" height="4.5"/>
 </g>
</svg>
```

client/views/view_research.cpp
```diff
@@ -20,12 +20,16 @@
 #include <QPainter>
 #include <QPushButton>
 #include <QScrollArea>
+#include <QStringLiteral>
 #include <QTimer>
+#include <QToolButton>
```
jwrober, line 25: Why `QToolButton` when we already have `QPushButton` for the refresh button?

jwrober (commented):
Rest looks good to me.

client/views/view_research.cpp
blabber: Because Louis said:

> I would try a design with an icon-only `QToolButton`...

As QT is still uncharted territory for me, I can't say whether there is any objective reason to prefer one button class over the other.

lmoureaux: QPushButton is rectangular with text, QToolButton is square and icon-only.

blabber:
I updated the icon with the one @lmoureaux contributed; please check I got the license file right.

PR #2393, opened 2024-09-21
Fix drawing AllSeparate roads in hex mode

lmoureaux:
Some roads were missing in hexemplio because of confusion between two kinds of indices, directions valid for the tileset and all possible directions.

Reported by jockel, Corbeau, and others on Discord.


hugomflavio (approved):
Tested hexemplio and amplio. It's working nicely on both; and fixes the missing vertical road sprite.

PR #2407, opened 2024-10-18
Support ESC to close the production menu

### LINKED ISSUE #2383: Allow ESC to close build window
300168893-4d2d67c5-8d2a-4226-970b-7f778b221f3e

This window (widget?) is somewhat tricky to close. ESC doesn't close it, I need to click _somewhere else_, but there is a possibility that clicking _somewhere else_ may activate _something else_. So, basically, ESC would do.

lmoureaux:
There were already a Cancel button and right click to close, but ESC is more natural.

Closes #2383.


jwrober (approved):
Works for me!

PR #2409, opened 2024-10-21
Save less state in minimap_view

### LINKED ISSUE #1657: Minimap out of sync
Clicking on the minimap does not jump you to where you click, and the frame that represents the current view in the minimap also is not where you are.
It happened for me in LT75 and is like that from start in LT76.

lmoureaux:
The widget needs to compute horizontal and vertical scaling factors to map between the displayed minimap (in pixel units) and the computed one (essentially in tile units). It was computing them on resize events and storing them, which was causing issues when switching maps. Compute them when they are needed instead, it's just two float divisions.

Also remove a bunch of unused variables from the class.

Closes #1657.

### DIFF
--- a/client/minimap.cpp
+++ b/client/minimap.cpp
@@ -32,26 +32,10 @@ const auto always_visible_margin = 15;
 /**
    Constructor for minimap
  */
-minimap_view::minimap_view(QWidget *parent) : fcwidget()
+minimap_view::minimap_view(QWidget *parent) : fcwidget(parent)
 {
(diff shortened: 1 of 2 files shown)


hugomflavio (approved):
Tested switching between two games without closing the client. The mini-map resized correctly. Glad to see this one resolved! Also tested changing zoom levels; the mini-map also responded well.

jwrober (approved):
Works for me too.

PR #2416, opened 2024-10-25
Fix a crash in the units view

lmoureaux:
The units view was crashing when more than 250 units were on UWT. This was reported by @Corbeau-1 on Discord (Sim06).

The second commit isn't strictly necessary but keeps things consistent.


blabber (approved):
LGTM.

I can't reproduce the bug this is fixing, but I did a quick regression test. In my local game the units report, and especially the UWT table, works as expected.

PR #2420, opened 2024-10-27
Fix reports being shown out of screen

### LINKED ISSUE #1989: Historian report goes off the screen
**Describe the bug**
The historian report widget can appear off the screen

**To Reproduce**
Steps to reproduce the behavior: Not exactly sure how to re-create. I just logged into a running LT game and the widget shows up off the side of the screen.

**Expected behavior**
The widget stays in the confines of the game window.

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.1-alpha.2

**Additional context**
LT78 Turn 47

lmoureaux:
When connecting to a game and receiving an historian report, they would often be shown out-of-screen. This is ultimately caused by the resizing of the client window happening only after the game state is fully loaded. The placement algorithm has to provide sensible results when the widget doesn't fit inside the (small) base window size.

The complicated placement algorithm relied on an initial guess for the position and could result in the widget being shown off-screen when the window was too small. Replace it with a linear scan going left-to-right and top-to-bottom across the entire screen. Also avoid recursion, permitting more attempts by lifting the limitation due to the stack size.

This doesn't entirely solve the issue of widgets being shown in an unexpected location when connecting to a game: when the algorithm finds no solution, it now defaults to 0,0 to maximize visible content. This can in particular be the case for the small window size used right before connecting. I think a full revision of the pregame screens is needed, but this is way beyond the scope of this PR and won't make it to 3.1.

Closes #1989.

## Before Patch

image

## After Patch

image

The Demographics report gets placed correctly when invoked:

image

### DIFF
--- a/client/dialogs.cpp
+++ b/client/dialogs.cpp
@@ -987,6 +987,7 @@ void popup_notify_dialog(const char *caption, const char *headline,
 
   freeciv::report_widget *nd = new freeciv::report_widget(
       caption, headline, lines, queen()->mapview_wdg);
+  nd->move(queen()->mapview_wdg->find_place(nd->size()));
   nd->show();
 }
 
--- a/client/views/view_map.cpp
+++ b/client/views/view_map.cpp
@@ -60,14 +60,6 @@ extern int last_center_player_city;
 extern int last_center_enemy_city;
 
 /**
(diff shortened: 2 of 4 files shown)


jwrober (approved):
Works for me!

PR #2426, opened 2024-10-28
Action names: You're Fired -> Disband Unit

### LINKED ISSUE #2392: "You're Fired" action is confusing
We have a unit action called "You're Fired" that allows a player to disband a unit inside a city to recoup some shields. The term "You're Fired" is confusion.

Replace with "Disband Unit" or something more clear.

None

**Additional context**
Came up on Discord: https://discord.com/channels/378908274113904641/912500712833974322/1285687494247055422

lmoureaux:
"You're Fired" was difficult to understand and was on a totally different tone than other action names.

Closes #2392.

### DIFF
--- a/common/actions.cpp
+++ b/common/actions.cpp
@@ -6219,8 +6219,8 @@ const char *action_ui_name_default(int act)
     // TRANS: Rec_ycle Unit (100% chance of success).
     return N_("Rec%sycle Unit%s");
   case ACTION_DISBAND_UNIT:
-    // TRANS: _You're Fired (100% chance of success).
-    return N_("%sYou're Fired%s");
+    // TRANS: _Disband Unit (100% chance of success).
+    return N_("%sDisband Unit%s");
   case ACTION_HOME_CITY:
     // TRANS: Set _Home City (100% chance of success).
     return N_("Set %sHome City%s");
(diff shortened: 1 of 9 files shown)


jwrober:
Does this change break the other LT rulesets?

lmoureaux:
It shouldn't affect them at all -- they'll keep showing `You're Fired`.

common/actions.cpp
```diff
@@ -6219,8 +6219,8 @@ const char *action_ui_name_default(int act)
     // TRANS: Rec_ycle Unit (100% chance of success).
     return N_("Rec%sycle Unit%s");
   case ACTION_DISBAND_UNIT:
-    // TRANS: _You're Fired (100% chance of success).
-    return N_("%sYou're Fired%s");
+    // TRANS: _Disband Unit (100% chance of success).
+    return N_("%sDisband Unit%s");
```
jwrober, line 6223: The reason I'm asking is this change to `actions.cpp`. So this won't effect rulesets that use alternate language? Do we need to keep both in the `cpp` file?

jwrober (commented):
.

common/actions.cpp
lmoureaux: I see, good call! This is used as a default when no action name is specified in the ruleset and to avoid saving the default names in the ruleset editor.

jwrober (approved):
Ok I tested with built-in rules via Tutorial and with a copy of LTX with old action names and had no issues.

PR #2428, opened 2024-10-30
City Dialog: ExpCol present units list

### LINKED ISSUE #2295: Add "expand" button to units panel in new City Dialog
When you have a large number of units in a city, it can be difficult to see them and even harder to find and pick the one you want to engage the enemy with.

Reduce the units panel to only show a single row of units (not the 1.5 we have now) and then when there is more than one row enable an "up arrow" expand button that would enlarge the panel to something like 4 or 5 rows max to make it easier to see/work with the units.

None other than the status quo. We do have grand plans to make more changes to the city dialog, but that is probably a long way off.

**Additional context**
Came up in a conversation with @hugomflavio

jwrober:
Closes #2295

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -904,6 +904,8 @@ city_dialog::city_dialog(QWidget *parent) : QWidget(parent)
   connect(ui.p_table_p->selectionModel(),
           &QItemSelectionModel::selectionChanged, this,
           &city_dialog::item_selected);
+  connect(ui.present_units_exp_col_but, &QAbstractButton::clicked, this,
+          &city_dialog::present_units_exp_col);
 
   // governor tab
   ui.qgbox->setTitle(_("Presets:"));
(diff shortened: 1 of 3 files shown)


client/citydlg.cpp
```diff
@@ -1789,11 +1791,37 @@ void city_dialog::update_units()
     units = pcity->tile->units;
   }
 
+  // set direction of the expand/collapse arrow
+  if (present_units_exp == false) {
```
lmoureaux, line 1795:
```suggestion
  if (!present_units_exp) {
```

client/citydlg.cpp
```diff
@@ -1789,11 +1791,37 @@ void city_dialog::update_units()
...
-  ui.curr_units->setText(QString(buf));
+  ui.present_units_label->setText(QString(buf));
+}
+
+/**
+ * Slot to expand or collapse the present units list
+ */
+void city_dialog::present_units_exp_col()
+{
+  if (present_units_exp == false) {
```
lmoureaux, line 1814:
```suggestion
  if (!present_units_exp) {
```

lmoureaux (commented):
Small style comments. IMNSHO the conditions would be easier to read if the `if` and `else` cases were inverted

jwrober:
> Small style comments. IMNSHO the conditions would be easier to read if the `if` and `else` cases were inverted

so do the true condition first with else being false?

lmoureaux:
> > Small style comments. IMNSHO the conditions would be easier to read if the `if` and `else` cases were inverted
> 
> so do the true condition first with else being false?

Yes, to avoid having to think about the `!` or `== false`. But it's eventually a matter of taste.

hugomflavio (commented):
tested, looks pretty cool!

PR #2440, opened 2024-11-07
Bugfix release.yaml for location of freeciv21-server.cmd

jwrober:
Found releasing 3.1-rc.1


lmoureaux (approved):
While at it, should we add the `-f` flag so it doesn't fail if the file isn't there?

jwrober:
> While at it, should we add the `-f` flag so it doesn't fail if the file isn't there?

good call

PR #2445, opened 2024-11-18
Minor fixes to battle messages

hugomflavio:
In some places, "A" and "D" were switched. Also in a case the defence value was not displayed at all; which was not consistent with the rest.


jwrober (approved):

jwrober:
Looks good to me

PR #2463, opened 2024-12-23
Allow setting landmass to 100

### LINKED ISSUE #2444: Remove limitation [15, 85] for landmass
Currently it is impossible to generate map that would contain 0% or 100% land. The limit is set to 15% and 85%. Please remove it.

I presume there may be generator crashes, however, this is up to the ruleset maker to solve, that is, to not set up such numbers to cause the crash.

Reason: creating some exotic tile/rulesets that require some more flexibility.

lmoureaux:
Some generators can generate maps without water. Allow game admins to request this.

Closes #2444.


jwrober:
You need to update docs - Manuals/Server/options

jwrober (approved):

PR #2466, opened 2024-12-23
Climate change: Check unit activities on tiles surrounding changed one

### LINKED ISSUE #1064: Check legality of unit actions on adjacent tiles on tile change
**Describe the bug**
There's a long standing issue with typical rulesets where legality of the irrigating activity can depend on the irrigation of the adjacent tile. If that adjacent tile loses the irrigation for some reason, we later get an error about unit doing illegal activity. 

**To Reproduce**
1. Irrigate a tile
2. If neighbouring tile loses irrigation during TC, due to some reason you get an error for illegal behaviour

**Expected behavior**
Check in advance, so there is no illegal behaviour

**Screenshots**

master

**Additional context**
Port https://github.com/freeciv/freeciv/commit/15d274689daa21a1361491f5348abf37dd56381f

lmoureaux:
See osdn #44155
Closes #1064.

@XHawk87

### DIFF
--- a/server/maphand.cpp
+++ b/server/maphand.cpp
@@ -199,13 +199,7 @@ void climate_change(bool warming, int effect)
       update_tile_knowledge(ptile);
 
       // Check the unit activities.
-      unit_list_iterate(ptile->units, punit)
-      {
-        if (!can_unit_continue_current_activity(punit)) {
-          unit_activity_handling(punit, ACTIVITY_IDLE);
-        }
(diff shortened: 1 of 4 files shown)


jwrober (approved):

jwrober:
CodeFactor no happy, but that's an old function.

lmoureaux:
If anything I'm making it simpler :)

PR #2474, opened 2024-12-28
Implement units cycling

blabber:
Since the new Units Report was introduced in 3.1, there was no longer a way to cycle all units of a given unit type on the map.

With this commit scrolling the mouse wheel on the selected unit icon in the "Unit Controls" widget, cycles through idle or sentried units of the same unit type.

### DIFF
--- a/client/hudwidget.cpp
+++ b/client/hudwidget.cpp
@@ -44,6 +44,12 @@
 #include "widgets/decorations.h"
 
 static QString popup_terrain_info(struct tile *ptile);
+static bool cycle_units_predicate(const struct unit *current_unit,
+                                  const struct unit *candidate);
+static std::tuple<int, int>
+cycle_units_get_count_and_index(const struct unit *current_unit);
+static void cycle_units_select_index(const struct unit *current_unit,
+                                     int index);
 
 /**
    Returns true if player has any unit of unit_type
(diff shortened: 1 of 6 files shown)


blabber:
This is something I was missing in Sim06. The be honest the UX is obscure, but it is similar to the way it worked in the units report pre-3.1.  And it is easier than working with the advanced unit selection for this use case.

blabber:
CodeFactor is not happy, although I don't think the method it is complaining about is overly complex. Will check later.

jwrober:
I had an idea to change the units view to allow for expanding the table to show rows for each unit type and you could double-click to got to the specific unit. This is a nice side idea too. I'll have a look for review.

jwrober:
I can't seem to get this to work. If I select a unit on the map, hover my mouse over it and scroll the wheel, the map moves up/down.

jwrober:
https://github.com/user-attachments/assets/1771cabe-e49c-4f25-8e95-0adffc7b997e

blabber:
This is a misunderstanding. Try the icon in the units control widget.

20241230_17h42m20s_grim

blabber:
Fixed the CodeFactor warning.

client/hudwidget.cpp
```diff
@@ -1731,3 +1770,135 @@ void hud_battle_log::showEvent(QShowEvent *event)
...
+{
+  if (get_num_units_in_focus() != 1) {
+    return;
+  }
+
+  struct unit *current_unit = head_of_units_in_focus();
+
+  int unit_count = 0;
+  int current_unit_index = 0;
+  std::tie(unit_count, current_unit_index) =
```
lmoureaux, line 1882: Structured binding (we have C++17, let's make use of it!)
```suggestion
  auto [unit_count, current_unit_index] =
```

client/hudwidget.cpp
```diff
@@ -1731,3 +1770,135 @@ void hud_battle_log::showEvent(QShowEvent *event)
...
+/**
+   Cycle idle or sentried units for the unittype of the currently
+   selected unit.
+
+   The sign of direction determines in which direction the units will
+   be cycled.
+ */
+void cycle_units(const int direction)
+{
+  if (get_num_units_in_focus() != 1) {
```
lmoureaux, line 1874: I think you were right not to consider the case `>1`. Would you have a suggestion of what to do in that case?

docs/Manuals/Game/unit-controls.rst
```diff
@@ -42,3 +42,6 @@ appear in quotes after the Unit ID value.
 If the unit selected is not in your field of vision on the map, then you can click on the icon for the unit on
 the left side and the game map will center on the unit for you. As with other widgets in Freeciv21, you can
 click+drag the widget to move it by using the plus symbol in the upper left corner.
+
+If you scroll your mouse wheel over the unit icon on the left side of the widget, the game will cycle through
+idle or units of the same unit type.
```
lmoureaux, line 47:
```suggestion
idle or sentried units of the same unit type.
```

client/hudwidget.cpp
```diff
@@ -1731,3 +1770,135 @@ void hud_battle_log::showEvent(QShowEvent *event)
...
+
+  if (ACTIVITY_IDLE != candidate->activity
+      && ACTIVITY_SENTRY != candidate->activity) {
+    return false;
+  }
+
+  if (!can_unit_do_activity(candidate, ACTIVITY_IDLE)) {
+    return false;
+  }
+
```
lmoureaux, line 1798: Should we exclude sentried units in boats?

client/hudwidget.cpp
blabber: No, I couldn't think of a sensible way of dealing with the case of multiple selected units. That's why I decided to exit early in this case and make this  a no-op.

client/hudwidget.cpp
blabber: The checks in `cycle_units_predicate` are based on the `find nearest` algorithm used by the units report: 

https://github.com/longturn/freeciv21/blob/df33c26984b67cea7f86c2bfd99f1dc6c97ac421/client/views/view_units.cpp#L553-L558

I think it makes sense not to introduce a changed behaviour for this functionality in order to keep it consistent.

lmoureaux: Two things here:
* Can we avoid duplicating the logic?
* I'm not convinced by the `can_unit_do_activity` check, we can select units without cancelling their orders. The [original commit](https://github.com/freeciv/freeciv/commit/e839b2595f7d5234338f091c49ac4e825090c233) doesn't provide much information. We should probably open a new issue to look into this further.

blabber: * I factored out the logic into a function in `controls.cpp`. Looking at the other functions in this file, I figured it is fitting.
* The `can_unit_do_activity` check was also in the original cycle units code in the old units report. I guess the new code was based on this and inherited the check.

https://github.com/longturn/freeciv21/blob/3003835fa85260bafdd48f4895c0da86b0cdf149/client/unitreport.cpp#L362-L367

lmoureaux (approved):
I'll let @jwrober check that it works but I'm happy from the code perspective :)

jwrober (approved):

Issue #2477, opened 2024-12-30
Segfault when city removal causes boat displacement and unit removed with Lua during unit_moved event


XHawk87:
**Describe the bug**
The server crashes if a city is disbanded, for example by migration, and the city contained a boat which was forced to move off of the city tile, and during the `unit_moved` Lua event the boat is destroyed.

**To Reproduce**
Steps to reproduce the behaviour:
1. Load the saved game
2. Start the game
3. See crash

**Expected behaviour**
No crash.

**Screenshots**
![image](https://github.com/user-attachments/assets/d6d8dd38-825b-41c2-b7b4-e2d04c138736)

**Platform and version (please complete the following information):**
 - OS: Linux
 - Freeciv21 version: 3.1-rc.1
 - Ruleset/Longturn game (if applicable): Chaos

**Additional context**
[Discord discussion](https://discord.com/channels/378908274113904641/912500712833974322/1323107746811347027)

XHawk87:
I'm looking into this bug. I found that it's possible to use `bounce_unit()` for all but one feature of the custom bounce code, displaying a city tile link for the disbanded city.

https://github.com/longturn/freeciv21/blob/74984fadae58b425e15eedad818b3e68eeeb0b58/server/citytools.cpp#L1688

https://github.com/longturn/freeciv21/blob/74984fadae58b425e15eedad818b3e68eeeb0b58/server/citytools.cpp#L1748-L1753

How would you go about solving this? Should I pass in some extra data with the `bounce_reason` somehow?

https://github.com/longturn/freeciv21/blob/74984fadae58b425e15eedad818b3e68eeeb0b58/server/unittools.h#L108-L115

XHawk87:
Actually, single-responsibility. I don't think `bounce_unit()` should also be trying to communicate, it should just be bouncing the unit and reporting success or failure. We should have something else that's solely responsible for communicating the outcome.

XHawk87:
See related refactor #2478

XHawk87:
See another related refactor #2481

PR #2480, opened 2025-01-02
Refactored bounce_unit to allow reporting using function callbacks

### LINKED ISSUE #2478: Refactor `bounce_unit` to split out additional responsibilities into separate functions
Currently `bounce_unit` doesn't adhere to the single responsibility principle as it also attempts to communicate the outcome of the bounce. This is dependent on the reason for the bounce, which is getting passed in as a parameter. This will only get more complex over time as more reasons are accumulated. 

Split out communication of the outcome into separate functions which can be called from the call site where needed.

We could pass in references to callback functions to use as required depending on the outcome of the bounce. Though this might be adding unnecessary complexity.

**Additional context**
https://github.com/longturn/freeciv21/blob/74984fadae58b425e15eedad818b3e68eeeb0b58/server/unittools.h#L108-L115

XHawk87:
Closes #2478

### DIFF
--- a/server/maphand.cpp
+++ b/server/maphand.cpp
@@ -1788,7 +1788,20 @@ static void check_units_single_tile(struct tile *ptile)
   {
     if (unit_tile(punit) == ptile && !unit_transported(punit)
         && !can_unit_exist_at_tile(&(wld.map), punit, ptile)) {
-      bounce_unit(punit, true, bounce_reason::terrain_change, 1);
+      bounce_unit(
+          punit, 1,
+          [](struct bounce_event bevent) -> void {
+            notify_player(unit_owner(bevent.bunit), bevent.to_tile,
+                          E_UNIT_RELOCATED, ftc_server,
+                          _("Moved your %s due to changing terrain."),
+                          unit_link(bevent.bunit));
+          },
+          [](struct bounce_event bevent) -> void {
+            notify_player(unit_owner(bevent.bunit), unit_tile(bevent.bunit),
+                          E_UNIT_LOST_MISC, ftc_server,
+                          _("Disbanded your %s due to changing terrain."),
+                          unit_tile_link(bevent.bunit));
+          });
     }
   }
   unit_list_iterate_safe_end;
--- a/server/maphand.cpp
+++ b/server/maphand.cpp
@@ -1796,7 +1796,7 @@ static void check_units_single_tile(struct tile *ptile)
                           _("Moved your %s due to changing terrain."),
                           unit_link(bevent.bunit));
           },
-          [](struct bounce_event bevent) -> void {
+          [](struct bounce_disband_event bevent) -> void {
             notify_player(unit_owner(bevent.bunit), unit_tile(bevent.bunit),
                           E_UNIT_LOST_MISC, ftc_server,
                           _("Disbanded your %s due to changing terrain."),
--- a/server/unittools.cpp
+++ b/server/unittools.cpp
@@ -1331,18 +1331,22 @@ bool bounce_path_constraint::is_allowed(
  * See `bounce_unit_silently` to attempt a bounce without reporting it to the
  * unit's owner.
  *
- * Note: The unit may have died even if the bounce was successful.
- * Note: If a transport fails to bounce, all units in the transport will be
+ * \note
+ * The unit may have died even if the bounce was successful.
+ *
+ * \note
+ * If a transport fails to bounce, all units in the transport will be
  * bounced, triggering on_success or on_fail for each.
  */
 void bounce_unit(struct unit *punit, int max_distance,
                  std::function<void(struct bounce_event)> on_success,
-                 std::function<void(struct bounce_event)> on_failure)
+                 std::function<void(struct bounce_disband_event)> on_failure)
 {
   if (!punit) {
     return;
   }
 
+  const int unit_id = punit->id;
   const auto pplayer = unit_owner(punit);
   const auto punit_tile = unit_tile(punit);
 
@@ -1394,7 +1398,7 @@ void bounce_unit(struct unit *punit, int max_distance,
 
       handle_unit_orders(pplayer, &packet);
 
-      if (!punit) {
+      if (!unit_exists(unit_id)) {
         return; // Unit died while executing orders
       }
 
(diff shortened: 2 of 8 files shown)


server/unittools.cpp
```diff
@@ -1394,6 +1394,10 @@ void bounce_unit(struct unit *punit, bool verbose, bounce_reason reason,
 
       handle_unit_orders(pplayer, &packet);
 
+      if (!punit) {
```
lmoureaux, line 1397: This will not work. The pointer is set at the beginning of the function and is not modified when killing the unit; only the associated memory is freed, resulting in a dangling pointer (which may or may not crash depending on [your compiler](https://feross.org/gcc-ownage/) and runtime).

The usual solution is to store the unit ID, then query the unit again after executing orders.

server/unittools.cpp
```diff
@@ -1409,27 +1413,43 @@ void bounce_unit(struct unit *punit, bool verbose, bounce_reason reason,
...
+    struct bounce_event bevent)
+{
+  notify_player(unit_owner(bevent.bunit), bevent.to_tile, E_UNIT_RELOCATED,
+                ftc_server,
+                // TRANS: A unit is moved to resolve stack conflicts.
+                _("Moved your %s."), unit_link(bevent.bunit));
+}
+
+void report_unit_disbanded_to_resolve_stack_conflicts(
+    struct bounce_event bevent)
```
lmoureaux, line 1445: Let's add some documentation, even if minimal.

.gitignore
```diff
@@ -46,6 +46,9 @@ build.ninja
 *_autogen
 /Testing
 
+# Custom user scripts
+*.sh
```
lmoureaux, line 50: I think this will hide the two scripts we have in the tree, but I may be wrong:
```
./translations/generate_skips.sh
./translations/stats.sh
```
I suggest to restrict it to the root folder,
```suggestion
/*.sh
```

server/maphand.cpp
```diff
@@ -1788,7 +1788,20 @@ static void check_units_single_tile(struct tile *ptile)
...
+      bounce_unit(
+          punit, 1,
+          [](struct bounce_event bevent) -> void {
+            notify_player(unit_owner(bevent.bunit), bevent.to_tile,
+                          E_UNIT_RELOCATED, ftc_server,
+                          _("Moved your %s due to changing terrain."),
+                          unit_link(bevent.bunit));
+          },
+          [](struct bounce_event bevent) -> void {
+            notify_player(unit_owner(bevent.bunit), unit_tile(bevent.bunit),
```
lmoureaux, line 1800: Why go through `unit_tile()` here instead of using `to_tile`?

server/unittools.cpp
```diff
@@ -1319,13 +1320,24 @@ bool bounce_path_constraint::is_allowed(
...
+ * before the unit is moved. A function reference may also be provided to
+ * handle failing to find a bounce location. By default this will report the
+ * failed bounce to the unit's owner before the unit is disbanded.
+ *
+ * See `bounce_unit_silently` to attempt a bounce without reporting it to the
+ * unit's owner.
+ *
+ * Note: The unit may have died even if the bounce was successful.
+ * Note: If a transport fails to bounce, all units in the transport will be
+ * bounced, triggering on_success or on_fail for each.
```
lmoureaux, line 1336: Note: Doxygen has a special command for this, called [`\note`](https://www.doxygen.nl/manual/commands.html#cmdnote) (or `@note` in JavaDoc style). It renders a bit nicer (if one can call the Doxygen output *nice*).

server/unittools.cpp
```diff
@@ -1409,27 +1413,43 @@ void bounce_unit(struct unit *punit, bool verbose, bounce_reason reason,
    * Try to bounce transported units. */
   if (0 < get_transporter_occupancy(punit)) {
     const auto pcargo_units = unit_transport_cargo(punit);
-    unit_list_iterate(pcargo_units, pcargo) { bounce_unit(pcargo, verbose); }
+    unit_list_iterate(pcargo_units, pcargo)
+    {
+      bounce_unit(pcargo, max_distance, on_success, on_failure);
```
lmoureaux, line 1418: Here we are in a tricky situation, because the script running in the move handler might have killed everything on the tile: "unloading a unit may trigger a nuclear explosion" sounds fitting for the Chaos ruleset. I think we need `unit_list_iterate_safe` and check that the cargo still exists afterwards.

server/unittools.h
```diff
@@ -105,14 +107,21 @@ int get_unit_vision_at(struct unit *punit, const struct tile *ptile,
...
+struct bounce_event {
+  struct unit *bunit;
+  struct tile *to_tile;
 };
-void bounce_unit(struct unit *punit, bool verbose,
-                 bounce_reason reason = bounce_reason::generic,
-                 int max_distance = 2);
+void report_unit_bounced_to_resolve_stack_conflicts(
+    struct bounce_event bevent);
+void report_unit_disbanded_to_resolve_stack_conflicts(
```
lmoureaux, line 116: Are they really only used to solve stack conflicts?

lmoureaux (commented):
I like the changes, but this included a fix for the equivalent of #2477 in `bounce_unit` that won't work (we can also do this part later).

server/unittools.h
XHawk87: That's how the translation note is worded. If it's not, we need separate messages for different usages. (Which I believe we definitely should, but that's outside the scope of a refactor.)

server/unittools.cpp
lmoureaux: I suspect that `pcargo_units` will be freed if `pcargo` gets killed in an event handler (and hopefully the loaded units will have been deleted before). Also, you need to check if `pcargo` still exists before calling `on_failure` or `wipe_unit` with it.

lmoureaux (changes requested):

server/unittools.cpp
lmoureaux: Sorry, I thought the transported unit list being deleted would also create issues, but on reading  `unit_list_iterate_safe` I see that it is copied before iterating. So we're safe from that.

lmoureaux (approved):

PR #2484, opened 2025-01-05
Fix selling improvements for multiple cities at once

### LINKED ISSUE #2483: Selling improvements from several cities at once does not work
**Describe the bug**
The cities view implies that you can sell improvements from multiple cities at once, but it is not working.

**To Reproduce**
1. Go to 'Cities' view (F4)
2. Select multiple cities
3. Sell an improvement using the context menu
4. The improvement is not sold from the selected cities.

**Expected behavior**
The improvement should be sold from all selected cities.

 - OS: FreeBSD 14.2-RELEASE
 - Freeciv21 version: 3.1-rc.1

**Additional context**
Related [discussion](https://discord.com/channels/378908274113904641/1011029409572077699/1324503132889284609) on Discord.

blabber:
Closes #2483

### DIFF
--- a/client/views/view_cities.cpp
+++ b/client/views/view_cities.cpp
@@ -21,6 +21,7 @@
 #include "fcintl.h"
 // common
 #include "citydlg_common.h"
+#include "game.h"
 #include "global_worklist.h"
 // client
 #include "cityrep_g.h"
@@ -565,10 +566,7 @@ void city_widget::display_list_menu(const QPoint)
     cid id;
     struct universal target;
     QString imprname;
-    const struct impr_type *building;
-    Impr_type_id impr_id;
     bool need_clear = true;
-    bool sell_ask = true;
 
     if (!act) {
       return;
@@ -579,6 +577,11 @@ void city_widget::display_list_menu(const QPoint)
     qvar = act->data();
     id = qvar.toInt();
     target = cid_decode(id);
+
+    if (m_state == SELL) {
+      sell(target.value.building);
+      return;
+    }
 
     city_list_iterate(client_player()->cities, iter_city)
     {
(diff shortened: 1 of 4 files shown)


blabber:
I will keep this as a draft for now. I aim to  factor out the selling of improvements into it's own function/method before merging it.

client/views/view_cities.cpp
```diff
@@ -732,6 +705,47 @@ void city_widget::display_list_menu(const QPoint)
...
+  ask->set_text_title(buf, _("Sell?"));
+  ask->setAttribute(Qt::WA_DeleteOnClose);
+  connect(ask, &hud_message_box::accepted, this, [=]() {
+    const auto saved_selection =
+        selected_cities; // Copy to avoid invalidations
+
+    for (auto *pcity : saved_selection) {
+      int city_id = pcity->id;
+      if (nullptr == game_city_by_number(city_id)) {
+        return;
```
lmoureaux, line 737: I think skipping the lost city makes more sense in this case.
```suggestion
        continue;
```

client/views/view_cities.cpp
```diff
@@ -732,6 +705,47 @@ void city_widget::display_list_menu(const QPoint)
...
+  QString buf =
+      QString(_("Are you sure you want to sell the %1?")).arg(imprname);
+  ask->setStandardButtons(QMessageBox::Cancel | QMessageBox::Yes);
+  ask->setDefaultButton(QMessageBox::No);
+  ask->button(QMessageBox::Yes)->setText(_("Yes Sell"));
+  ask->set_text_title(buf, _("Sell?"));
+  ask->setAttribute(Qt::WA_DeleteOnClose);
+  connect(ask, &hud_message_box::accepted, this, [=]() {
+    const auto saved_selection =
+        selected_cities; // Copy to avoid invalidations
```
lmoureaux, line 732: I'm not sure this line is correct. It is present in other methods because `selected_cities` is a class member that may be updated when a city changes. But here, doesn't `[=]` already copy the list? (Or does it copy the whole widget?)

lmoureaux (commented):
I think it would make sense to unify this with `eco_report::sell_buildings` in `view_economics.{h,cpp}`.

client/views/view_cities.cpp
blabber: `[=]` indeed captures the environment by value, if I understand correctly.

blabber:
> I think it would make sense to unify this with `eco_report::sell_buildings` in `view_economics.{h,cpp}`.

Unifying the behavior?  Showing a sell results window after selling the improvements from the cities view seems like a good idea, and I can do that. Unifying the code on the other hand...

lmoureaux:
Unifying code is probably better. The function `sell_all_improvements` in `repodlgs_common.cpp` could take a city list parameter (and be renamed accordingly), and maybe boilerplate for the "sell results" dialog could be shared as well. Not sure about the confirmation dialog since the message is slightly different: `sell every Temple` vs `sell the Temple`.

If you don't see how to do it we can open a refactoring issue.

blabber:
I followed your advice and unified the code used by both views to sell city improvements.

We could also generate the question and result dialog in this central function, but I am not sure if this is considered good practice, so I left it out for now. Let me know what you think.

lmoureaux (approved):
Thanks! I think we can remove the draft flag (and then I'll want to check the functionality one last time).

client/views/view_cities.cpp
```diff
@@ -732,6 +705,48 @@ void city_widget::display_list_menu(const QPoint)
...
+    return;
+  }
+
+  hud_message_box *ask = new hud_message_box(king()->central_wdg);
+  QString imprname = improvement_name_translation(building);
+  QString buf;
+  if (selected_cities.size() == 1) {
+    buf = QString(_("Are you sure you want to sell the %1?")).arg(imprname);
+  } else {
+    buf = QString(_("Do you really wish to sell every %1?")).arg(imprname);
```
lmoureaux, line 727: I think "every" can be misinterpreted as "every city you own".
```suggestion
    buf = QString(_("Do you really wish to sell all selected %1?")).arg(imprname);
```

lmoureaux (approved):
Small nitpick on the working, else ok!

PR #2494, opened 2025-01-13
Version 3.2 is alive!

jwrober:
You betcha it is :)


jwrober:
Suggesting a squash merge on this one.

jwrober:
I should add `patch` to the list of values in the regex and we can back port to stable. I'll push another commit

Issue #487, opened 2021-06-23
Create LTT ruleset manual


jwrober:
Part of Documentation project. First manual to add to `docs`

jwrober:
Draft is available on google docs for review -- https://docs.google.com/document/d/1nFgztVP7d1sYXYLJ343wcOcOYMI0CvNY9pBkgSmBqJQ/edit?usp=sharing

pranavtifr:
I'd say we close this and create this at https://github.com/longturn/LTT-LTX

jwrober:
This issue was not for a standard python extracted documentation of the ruleset. This issue is more about how to play the rulesets as opposed to just an extraction of the ruleset files to rtd.

pranavtifr:
Ya, thats why it needs to go to the ruleset repo.
Python/extraction stuff can stay here since they are more general (or maybe pyciv repo, not sure yet)

lmoureaux:
Relevant material: https://longturn21.blogspot.com/p/freeciv-for-absolute-beginners-short.html

jwrober:
> Relevant material: https://longturn21.blogspot.com/p/freeciv-for-absolute-beginners-short.html

This link is included in first PR

PR #2502, opened 2025-01-15
New LT Gamer's Guide: Part 1

jwrober:
This is step 1 of many to keep PR reviews short.

Part of #487

### DIFF
--- a/docs/Playing/LT-Guide/intro.rst
+++ b/docs/Playing/LT-Guide/intro.rst
@@ -1,0 +1,90 @@
+.. SPDX-License-Identifier: GPL-3.0-or-later
+.. SPDX-FileCopyrightText: James Robertson <jwrober@gmail.com>
+
+.. Custom Interpretive Text Roles for longturn.net/Freeciv21
+.. role:: unit
+.. role:: improvement
+.. role:: wonder
+.. role:: advance
+
+Introduction
+************
+
+The Longturn community maintains and plays three specialized rulesets: Longturn Tradidional (LTT), Longturn
+Experimental (LTX), and Royale. The LTT and LTX rulesets are losely based on the Civ2Civ3 ruleset as it existed
+"way back" in legacy Freeciv v2.0 days. Royale is new as of 2023.
+
+Here are some generalized notes related to the three rulesets:
+
+* :strong:`LTT`: There are actually two versions: LTT and LTT2. LTT is the original standard ruleset played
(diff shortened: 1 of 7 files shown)


docs/Playing/index.rst
```diff
@@ -6,6 +6,7 @@ Freeciv21 and any of the varying rulesets that the Longturn.net community likes
 
 
 .. toctree::
+  LT-Guide/index.rst
   faq.rst
   how-to-play.rst
```
blabber, line 11: In the LT Guide there is a note pointing to the How To Play, as a more general guide. I think we should add a corresponding note to the How To Play, pointing to the LT guide, as a guide more focused on Longturn games.

docs/Playing/index.rst
```diff
@@ -6,6 +6,7 @@ Freeciv21 and any of the varying rulesets that the Longturn.net community likes
 
 
 .. toctree::
+  LT-Guide/index.rst
```
blabber, line 9: I consider the LT Guide and the How To Play to be closely related. Therefore I would not separate the both guides withe FAQ entry. Additionally I would propose to let the general Guide come first, followed by the more specific guide.

1. How To Play
2. LT Guide
3. FAQ
4. ...

blabber (changes requested):

docs/Playing/LT-Guide/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Longturn games will generally use one of two map topologies: squares or hexes. Some players love squares, and
+hate hexes. Some players are the opposite! Longturn game admins will annouce up front what map topology a game
+will use.
+
+For squares: The shipped Amplio2 tileset is the default. It is in an isometric direction. There are two
+overhead tilesets available in the modpack installer --- RoundSquare and Sextant --- that you might prefer.
+
+For hexes: The shipped Hexemplio tileset is the default. It is in an isometric direction. There is an overhead
+tileset available in the modpack installer --- RoundHex --- that you might prefer.
```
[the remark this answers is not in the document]
jwrober: I'll add the install URL to the modpack installer. Thanks for checking. I made an assumption there.

docs/Playing/index.rst
jwrober: Good call out. Thanks

docs/Playing/index.rst
jwrober: Another good call.

Issue #2473, opened 2024-12-28
Lua: Support notifications with featured text


XHawk87:
**Is your feature request related to a problem? Please describe.**
I want to be able to send notifications to players including tile, unit, and city links in them using lua.

**Describe the solution you'd like**
As documented in the [Lua Reference Manual](https://freeciv.fandom.com/wiki/Lua_reference_manual):
- Add `link_text` method to City type
- Add `tile_link_text` method to City type
- Add `link_text` method to Unit type
- Add `tile_link_text` method to Unit type
- Add `link_text` method to Tile type

**Describe alternatives you've considered**
It might technically be possible to manually duplicate the format for featured text in message strings but that would be ugly and messy.

**Additional context**
This was already been done in: https://github.com/freeciv/freeciv/blob/64f64a14ea220185058d8b57de2bf5c809169a18/common/scriptcore/api_game_methods.h#L220

lmoureaux:
Can you pinpoint the commit on the fc side?

XHawk87:
This looks like it: https://github.com/freeciv/freeciv/commit/8bf3a070c308c2828a93b6434f7a5076bef376cf
Relating to: https://osdn.net/projects/freeciv/lists/archive/tickets/2021-December/002049.html

lmoureaux:
Thanks! Doesn't seem difficult to port. Change function headers and ship it.

PR #2508, opened 2025-01-18
Moved city looting to default.lua

### LINKED ISSUE #2307: Ruleset or server setting for disabling gold looting when a city is conquered
If a city is lost the defender is already having trouble. If no gold could be looted the defending party would have a better chance for counter attacking at a later time and this reduces the chance to lose most of the nation in a few turns. Another advantage for removing the gold looting is the reduced need to be online at the tc. If a nation is going to be attacked the defender greatly benefits if it's possible to spend the gold as fast as possible. This kind of RTS gold spending doesn't belong to all turn based games even while it might be a good idea for a short turn game or for some classic rulesets. 

The alternative on the Discord talks was defining different kind of setups and conditions based on the government the defender has. This however might be way too complicated for most cases. Keeping it simple would probably be enough for most of the use cases.

XHawk87:
Closes #2307 

Requires #2473

### DIFF
--- a/data/default/default.lua
+++ b/data/default/default.lua
@@ -227,3 +227,37 @@ function _deflua_harmless_disaster_message(disaster, city, had_internal_effect)
 end
 
 signal.connect("disaster_occurred", "_deflua_harmless_disaster_message")
+
+function _deflua_city_conquer_gold_loot(city, looterunit)
+  local treasury = city.owner:gold()
+  local loot = math.min(
+    treasury, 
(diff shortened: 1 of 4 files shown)


jwrober:
Code looks ok to me. How does one test this?

XHawk87:
I started an LTX local game, let the AI play a few turns, then took over one of them and spawned myself some units using lua and took some cities.

jwrober (approved):

PR #2510, opened 2025-01-19
New LT Gamer's Guide: Part 2

jwrober:
This is step 2 of many to keep PR reviews short.

Part of https://github.com/longturn/freeciv21/issues/487


docs/Playing/LT-Guide/twenty-turns.rst
```diff
@@ -0,0 +1,604 @@
...
+   :ref:`Nations view <game-manual-nations-and-diplomacy-view>` and ask to swap embassies. Many players will
+   agree to the sharing of embassies as it gives important information that both sides can equally use. This
+   is why it is important to do #1 first, so you have already started a dialog with your neighbor.
+
+At a minimum, find out where your neighbor’s capital city and major borders are. It goes without saying that
+you should not overstay your welcome. There is also nothing wrong with placing an :unit:`Explorers` on a
+shared border, on a tile inside your border, to act as a sentry.
+
+.. note::
+  As opposed to typical single-player :term:`AI` games, “peace” is the default state between nations in
```
blabber, line 51: Does it still hold true, that peace is the default state? I got the impression, that we are moving towards "war" as the default, because Armistice is a "newbie trap".

jwrober: Good catch. In LTT its Armistice->Peace and in LTX its War. I'll update the other branch with a change.

Issue #1612, opened 2022-12-17
Enable better F6-12 key support


jwrober:
**Is your feature request related to a problem? Please describe.**
Right now the following F# keys work in a toggle format:
* F2 Units widget
* F3 Nations view
* F4 Cities view
* F5 Economy view

If you press the key the item opens and if you press the key again it closes (e.g. a toggle)

**Describe the solution you'd like**
Do the same thing for the other F# keys. In these cases you can press the button once to open the item, but a subsequent press does not close it.

* F6 Research view
* F7 Wonders of the World widget
* F8 Top 5 cities widget
* F9 - does not have something set, maybe Messages window
* F10 - does not have anything set
* F11 Demographics widget
* F12 Spaceship view

**Describe alternatives you've considered**

**Additional context**
Came up as part of a Discord discussion - https://discord.com/channels/378908274113904641/912500712833974322/1053770640672444498

blabber:
F6 and F12 have been fixed in #2491.

blabber:
Completed with #2518 .

PR #2518, opened 2025-01-22
Toggle report notifications

blabber:
This addresses the missing parts from issue #1612. While I already made F6 (research view) and F12 (spaceship view) work as a toggle, making F7 (wonders of the world), F8 (top 5 cities) and F11 (demographics) work like that is a bit special.

These reports are generic notification popups, which the server triggers to be displayed. The flow is

1. the user presses F7
2. the client asks the server for the research report
3. the server sends the research report
4. the client displays the report in a notification popup by calling `popup_notify_dialog`

This PR changes the behavior of subsequent calls to `popup_notify_dialog`. The former behavior was to refresh an existing notification, the new behavior is to close the existing notification. The notification to be manipulated is, in both cases, identified by caption and headline.

A drawback of this implementation is, that toggling a report still needs a round-trip to the server:

1. the user presses F7, while the wonder of the worlds report is open
2. the client asks the server for the research report
3. the server sends the research report
4. the client registers that there is already an open report, identified by the caption and headline provided from the server, and closes it
5. the just received report is discarded.

Keeping state in the client and closing the notifications without involving the server would require the usage of magic strings to identify relevant notification (this implementation does use string comparisons too, but these strings are provided by the server and, at least from the clients perspective, less magic).

Giving the notifications an "identity" would require either changes in the server and the protocol, or the client would need to deduce it by using magic strings.

### DIFF
--- a/client/dialogs.cpp
+++ b/client/dialogs.cpp
@@ -981,6 +981,7 @@ void popup_notify_dialog(const char *caption, const char *headline,
   for (auto report : list) {
     if (report->caption() == caption && report->headline() == headline) {
       report->close();
+      return;
     }
   }
 
--- a/client/dialogs.cpp
+++ b/client/dialogs.cpp
@@ -40,7 +40,6 @@
 #include "client_main.h"
 #include "control.h"
 #include "fc_client.h"
-#include "fonts.h"
 #include "helpdlg.h"
 #include "hudwidget.h"
 #include "icons.h"


blabber:
Marked this as draft to gather feedback about the code change.

blabber:
I have played the last days with this change. It is working and rather convenient, although I still have doubts the implementation will be accepted.

I remove the draft flag and await the verdict of the High Council.

lmoureaux (approved):
I guess this is the best possible implementation given protocol restrictions.

I checked out the branch and will try it out in the next turns, then merge if everything is alright.

PR #2522, opened 2025-02-03
Update Contributing/dev-env.rst

jwrober:
No related issue. Found while re-building a MSYS2 environment to test a new cmake bug.

### DIFF
--- a/docs/Contributing/dev-env.rst
+++ b/docs/Contributing/dev-env.rst
@@ -91,6 +91,8 @@ your computer:
 .. code-block:: sh
 
   ~/GitHub/freeciv21$ git remote add upstream https://github.com/longturn/freeciv21.git
+  ~/GitHub/freeciv21$ git fetch upstream
+  ~/GitHub/freeciv21$ git pull upstream master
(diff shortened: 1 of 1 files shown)


docs/Contributing/dev-env.rst
```diff
@@ -91,6 +91,8 @@ your computer:
 .. code-block:: sh
 
   ~/GitHub/freeciv21$ git remote add upstream https://github.com/longturn/freeciv21.git
+  ~/GitHub/freeciv21$ git fetch upstream
```
blabber, line 94: `git pull upstream master` includes the `fetch`, so this step could be dropped.

jwrober: I don't think so. You have to `fetch` to get any new tags and such. `pull` doesn't do that on its own.

blabber: Ah right, the `pull` probably fetches only the `master` branch and related tags in this example. I stand corrected :)

blabber (approved):

PR #2526, opened 2025-02-16
Intermediate cost in bribe calculations changed to float

JustinBerke0406:
The _cost_ variable used to calculate the bribery cost was an integer, which resulted in the cost being truncated in the middle of calculations.

Furthermore, the cost was multiplied by the shield build cost of a unit (an integer) divided by 10. Since these were both integers, any build cost less than 10 resulted in the cost being set equal to 0.


jwrober (approved):
Thanks @JustinBerke0406

PR #2537, opened 2025-03-01
Add notes about security for macOS install

jwrober:
No related issue. Came up on Discord from a new player.


blabber (commented):
Something about the way RTD renders the note is bugging me, it might be more pleasing if the links are listed in a bullet list. But in the end, this is a matter of taste.

20250301_23h37m05s_grim

blabber (approved):
Looks better. Thanks.

Issue #1902, opened 2023-04-06
Middle-click popup needs a refactor


jwrober:
**Is your feature request related to a problem? Please describe.**
There are a number of "issues" with the middle-click pop-up widget on the Map View...

1. If a unit has a goto order set, the widget can cover it and you can't see the goto line.
2. The widget only gives details for a single unit on the tile, not all of them.
3. The widget is not moveable, nor sticky (only active while middle-click is being done).
4. You cannot copy and paste text from it.
5. (bug) #1901

**Describe the solution you'd like**
Convert the notify popup to an actual widget that is unique, sticky and can be moved and closed. Allow CnP and show all unit details. 

If the unit has goto set, the line still shows while the widget is opened so when you move it you can still see what is going on there.

**Describe alternatives you've considered**

**Additional context**

blabber:
> 3. The widget is not moveable, nor sticky (only active while middle-click is being done).

The second part is no longer true. Since #2560 the popup stays open until the user presses `Esc`, clicks outside of the popup or clicks one of the links in the popup.

> 4. You cannot copy and paste text from it.

This has also changed with #2560. You can now select text in the popup and use copy and paste.

PR #2540, opened 2025-03-05
Proof of Concept: Links from tile info popup into help

blabber:
Current state: If you open a tile info popup and move your mouse over the popup while holding the trigger, the popup will stay open when you release the trigger. You can then click the various info links and the corresponding help entry is opened.

### DIFF
--- a/client/helpdlg.cpp
+++ b/client/helpdlg.cpp
@@ -91,6 +91,38 @@ void update_help_fonts()
   }
 }
 
+QString create_help_link(const char *name, help_page_type hpt)
+{
+  QString s;
+  s = QString(name).toHtmlEscaped().replace(QLatin1String(" "),
+                                            QLatin1String("&nbsp;"));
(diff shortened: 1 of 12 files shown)


jwrober:
Since you are making changes to the middle click widget, you might consider some of the things here - https://github.com/longturn/freeciv21/issues/1902

jwrober:
I've done some playing with this and really like it. You are also adding some things to tile.cpp that I was thinking were needed for the map editor so I'm really looking forward for this to be merge ready.

blabber:
> I've done some playing with this and really like it. You are also adding some things to tile.cpp that I was thinking were needed for the map editor so I'm really looking forward for this to be merge ready.

Thanks for testing. I submitted the first PR (#2560) that brings the basic infrastructure and a first link to master. I will keep working on this, but I close this draft to avoid irritation.

PR #2542, opened 2025-03-06
Reduce the width of the governor tab

lmoureaux:
This is done by removing needless frames and moving the Optimize: button to its own line. The gain is about 100px on my system.

Qt Designer tried to update enumerator names in the UI file and made many mistakes. I tried to fix them but there might still be some.

We may want to backport to 3.1 but this is also a bit risky given the mess created by Designer.

### DIFF
--- a/client/citydlg.cpp
+++ b/client/citydlg.cpp
@@ -909,8 +909,6 @@ city_dialog::city_dialog(QWidget *parent) : QWidget(parent)
           &city_dialog::present_units_exp_col);
 
   // governor tab
-  ui.qgbox->setTitle(_("Presets:"));
-
   ui.cma_table->horizontalHeader()->setSectionResizeMode(
       QHeaderView::Stretch);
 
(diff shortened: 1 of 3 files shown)


jwrober:
This "feels" like a big change, so I'm not sure about a back port, especially with all the changes Designer did.

Did you mean to change the background to the darker color?
image

vs what we have on the general tab?
image

lmoureaux:
We agree on no backport then.

The background in the general tab is lighter because of a scroll area for small screen heights. Adding a scroll area in the governor tab leads to size artifacts (somehow my Qt doesn't deal well with Libertine), and the light background is the same as what's used for the presets list. I could change the widget background or add frames like in the citizens tab (they'd use about 25px per side it seems).

jwrober:
> 
> The background in the general tab is lighter because of a scroll area for small screen heights. Adding a scroll area in the governor tab leads to size artifacts (somehow my Qt doesn't deal well with Libertine), and the light background is the same as what's used for the presets list. I could change the widget background or add frames like in the citizens tab (they'd use about 25px per side it seems).

ok that makes sense. I'll approve.

jwrober (approved):

PR #2543, opened 2025-03-06
Ask for confirmation before recycling units

lmoureaux:
Since this is a destructive action, better be sure.


jwrober (approved):
Works for me

PR #2544, opened 2025-03-08
Update fc21 links page

jwrober:
No related issue. Came up on Discord.

### DIFF
--- a/docs/README
+++ b/docs/README
@@ -1,12 +1,12 @@
 Root of sphinx-build files here for the Longturn Documentation Project
     https://github.com/orgs/longturn/projects/1
 
+To create a local test build run these commands from the source directory.
+
 LINUX::
 
-To create a local test build run this command from this same directory
-
     cd build
-    (once) pip install virtualenv
+    sudo apt-get install virtualenv     # Only required the first time
     python -m virtualenv venv
     source venv/bin/activate
     pip install -r ../docs/requirements.txt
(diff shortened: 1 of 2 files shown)


blabber (changes requested):
Following the issue report on discord I did a quick test, and at least a `/` in your password seems to break the parser. The client exits unexpectedly directly with a message like: `[critical] freeciv21-client - Error contacting server "<a garbled url>"`. [Url-encoding](https://en.wikipedia.org/wiki/Percent-encoding) the password fixes the issue. I think this is something that we should add to the docs too. 

Maybe something along the lines:

> Passwords containing special characters, e.g. “/”, can cause the client to exit unexpectedly without establishing a connection to the server. In this case, you must use your password (url-encoded)[https://en.wikipedia.org/wiki/Percent-encoding].

docs/README
```diff
@@ -1,13 +1,28 @@
 Root of sphinx-build files here for the Longturn Documentation Project
     https://github.com/orgs/longturn/projects/1
 
+LINUX::
+
 To create a local test build run this command from this same directory
 
     cd build
+    (once) pip install virtualenv
```
lmoureaux, line 9: I can already see people typing `(once)` in the terminal.
```suggestion
    pip install virtualenv  # Only the first time
```
And I wouldn't recommend installing anything via `pip` if you can install it with `apt`...

docs/README
```diff
@@ -1,13 +1,28 @@
...
 
-and then open build/docs/index.html in your browser of choice
+    and then open build/docs/index.html in your browser of choice
+
+WINDOWS::
+
+    cd build-vs
+    (once) pip install virtualenv 
+    python -m virtualenv venv
+    venv\Scripts\Activate.ps1 (or .bat for command prompt)
```
lmoureaux, line 23: Same as above:
```suggestion
    pip install virtualenv  # Only the first time
    python -m virtualenv venv
    venv\Scripts\Activate.ps1  # Or .bat for command prompt
```

PR #2545, opened 2025-03-08
Feature: Map Editor 3

jwrober:
Third commit of the map editor project.

This PR adds:
- A combo box with the list of players in the game/scenario. It doesn't do any work yet, just populated with values and flags.
- A new Tile Tool widget. It also does not really do anything yet.


client/editor/map_editor.cpp
```diff
@@ -111,3 +121,40 @@ void map_editor::close()
...
+ */
+void map_editor::select_tool_tile()
+{
+  if (ett_wdg_active) {
+    ui.tbut_edit_tile->setToolTip(_("Open Tile Tool"));
+    ett_wdg->hide();
+    ett_wdg_active = false;
+  } else {
+    ui.tbut_edit_tile->setToolTip(_("Close Tile Tool"));
+    ett_wdg = new editor_tool_tile(0);
```
lmoureaux, line 154:
```suggestion
    ett_wdg = new editor_tool_tile;
```
Background: by convention, the parent is optional and the default is `nullptr`.

client/editor/map_editor.h
```diff
@@ -4,19 +4,17 @@
...
-
-class QEvent;
+#include "editor/tool_tile.h"
 
 #include "ui_map_editor.h"
 /**
  * Map Editor Dialog Class
  */
-class map_editor : public QWidget {
+class map_editor : public QStackedWidget {
```
lmoureaux, line 17: Keep a `QWidget` if you're not using `QStackedWidget`'s capabilities

client/editor/map_editor.h
```diff
@@ -29,6 +27,12 @@ private slots:
   map_editor(QWidget *parent = 0);
   ~map_editor() override;
   void check_open();
+  void update_players();
+  bool players_done = false;
```
lmoureaux, line 31: Can this be kept `private`?

client/editor/map_editor.h
```diff
@@ -29,6 +27,12 @@ private slots:
...
+  void select_tool_tile();
+
+  editor_tool_tile *ett_wdg;
+  bool ett_wdg_active = false;
```
lmoureaux, line 35: Hopefully these can be `private` as well

client/editor/tool_tile.h
```diff
@@ -0,0 +1,27 @@
...
+ * Tile tool editor class
+ */
+class editor_tool_tile : public QWidget {
+  Q_OBJECT
+
+private:
+  Ui::FormToolTile ui;
+
+public:
+  editor_tool_tile(QWidget *parent = 0);
```
lmoureaux, line 20:
```suggestion
  editor_tool_tile(QWidget *parent = nullptr);
```

client/editor/tool_tile.cpp
```diff
@@ -0,0 +1,80 @@
...
+ *  \brief Show event for editor_tool_tile
+ */
+void editor_tool_tile::showEvent(QShowEvent *event) { setVisible(true); }
+
+/**
+ *  \brief Paint event for editor_tool_tile
+ */
+void editor_tool_tile::paintEvent(QPaintEvent *event)
+{
+  // Draw a rounded rectangle border around the Tile Tool widget
```
lmoureaux, line 67: Can this be moved to CSS? Then we don't need the function at all.
```css
editor_tool_tile {
  border: 1px solid white;
  border-radius: 5px;
}
```

client/editor/tool_tile.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ *  \brief Destructor for editor_tool_tile
+ */
+editor_tool_tile::~editor_tool_tile() {}
+
+/**
+ *  \brief Show event for editor_tool_tile
+ */
+void editor_tool_tile::showEvent(QShowEvent *event) { setVisible(true); }
```
lmoureaux, line 60: Not needed. `showEvent` notifies the widget that it just became visible, so `setVisible(true)` is a no-op.
```suggestion
```

client/editor/tool_tile.h
```diff
@@ -0,0 +1,27 @@
...
+  ~editor_tool_tile() override;
+  void select_tile();
+
+protected:
+  void showEvent(QShowEvent *event) override;
```
lmoureaux, line 25: Not needed, see above.
```suggestion
```

client/editor/map_editor.cpp
```diff
@@ -1,15 +1,16 @@
...
-#include "minimap_panel.h"
-#include "page_game.h"
-#include "views/view_map.h"
+// Client
+#include "citydlg.h"        //showEvent
+#include "client_main.h"    //check_open
+#include "fc_client.h"      //check_open
+#include "minimap_panel.h"  //showEvent
+#include "page_game.h"      //showEvent
+#include "views/view_map.h" //showEvent
```
lmoureaux, line 13: Not sure what these are for... I would interpret them as e.g. `minimap_panel::showEvent` being used, but this is not the case.

lmoureaux (changes requested):

client/editor/tool_tile.cpp
jwrober: I can't seem to find the right object. Its really the VBoxLayout on the map_editor widget we want to draw a line around. I'll open an issue to change from a paint event to css.

client/editor/map_editor.cpp
jwrober: This is to help remind me why I include the file. I've been hacking by adding varioud `.h` files to try different things so this helps me know which ones are vetted.

client/editor/tool_tile.cpp
```diff
@@ -0,0 +1,76 @@
...
+ *  \class editor_tool_tile
+ *  \brief A widget that allows a user to edit tiles on maps and scenarios.
+ *
+ *  As of March 2025, this is a brand new widget.
+ */
+
+/**
+ *  \brief Constructor for editor_tool_tile
+ */
+editor_tool_tile::editor_tool_tile(QWidget *parent)
```
lmoureaux, line 24: Call the parent class' constructor with `parent`:
```suggestion
editor_tool_tile::editor_tool_tile(QWidget *parent) : QWidget(parent)
```

client/editor/tool_tile.cpp
```diff
@@ -0,0 +1,76 @@
...
+  // Set the tool button or Tile Tool
+  ui.tbut_select_tile->setText("");
+  ui.tbut_select_tile->setToolTip(_("Select a Tile to Edit"));
+  ui.tbut_select_tile->setMinimumSize(32, 32);
+  ui.tbut_select_tile->setIcon(
+      QIcon::fromTheme(QStringLiteral("pencil-ruler")));
+  connect(ui.tbut_select_tile, &QAbstractButton::clicked, this,
+          &editor_tool_tile::select_tile);
+
+  setVisible(false);
```
lmoureaux, line 49: Widgets are invisible by default, this is not needed.
```suggestion
```

client/editor/tool_tile.cpp
lmoureaux: Figured it out, you need to inherit from `QFrame` instead of `QWidget`...

client/editor/tool_tile.cpp
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+/**
+ *  \brief Constructor for editor_tool_tile
+ */
+editor_tool_tile::editor_tool_tile(QWidget *parent)
+{
+  auto value_font = fcFont::instance()->getFont(fonts::chatline);
+  ui.setupUi(this);
+  parent = 0;
```
lmoureaux, line 28: This ain't doing anything (sets the `parent` variable to a null pointer, but you never use it again)
```suggestion
```

lmoureaux (changes requested):
Some more comments I discovered while playing with the border...

lmoureaux (approved):

PR #2554, opened 2025-03-10
Map Editor: Finish Players Combo

### LINKED ISSUE #2546: Map Editor: Players List
We have a list of players that needs to do something:

- [ ] When initialized, should select the current player
- [ ] The list of players should be alphabetized
- [ ] The list of players should give the Nation name in addition to the flag - FLAG | Player Name | Nation Name
- [ ] When a new/different player index is selected, the client should take that player and move the map to center on the capital city

jwrober:
Closes #2546

### DIFF
--- a/client/editor/map_editor.cpp
+++ b/client/editor/map_editor.cpp
@@ -4,13 +4,16 @@
 // Editor
 #include "editor/map_editor.h"
 
-// Client                   ::Where Needed::
+// Client                     ::Where Needed::
 #include "citydlg.h"        //map_editor::showEvent
 #include "client_main.h"    //map_editor::check_open
 #include "fc_client.h"      //map_editor::check_open
 #include "minimap_panel.h"  //map_editor::showEvent
 #include "page_game.h"      //map_editor::showEvent
 #include "views/view_map.h" //map_editor::showEvent
+
+// common
+#include "chatline_common.h" //map_editor::player_changed
 
 /**
  *  \class map_editor
@@ -28,6 +31,9 @@ map_editor::map_editor(QWidget *parent)
   setParent(parent);
   setSizePolicy(QSizePolicy::Maximum, QSizePolicy::Expanding);
   setAutoFillBackground(true);
+
+  // initialize in constructor
+  ett_wdg = new editor_tool_tile;
 
   ui.label_title->setText(_("MAP EDITOR"));
   ui.label_title->setAlignment(Qt::AlignCenter);
@@ -51,6 +57,11 @@ map_editor::map_editor(QWidget *parent)
       QIcon::fromTheme(QStringLiteral("editor-tile")));
   connect(ui.tbut_tool_tile, &QAbstractButton::clicked, this,
           &map_editor::select_tool_tile);
+
+  // Slot to handle selections from the players list
+  QObject::connect(ui.combo_players,
+                   QOverload<int>::of(&QComboBox::currentIndexChanged), this,
+                   &map_editor::player_changed);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


client/editor/map_editor.cpp
```diff
@@ -107,33 +118,78 @@ void map_editor::check_open()
...
- * \brief Populate a combo box with the current list of players
+ * \brief Populate a combo box with the current list of players by nation
+ * name
  */
 void map_editor::update_players()
 {
   if (!players_done) {
+
+    // Ensure we can take all possible nations
+    send_chat_printf("/set allowtake hHaAoObd");
```
lmoureaux, line 138: This will get sent quite often once we add editing players. Maybe move this next to enabling the editor.

client/editor/map_editor.cpp
```diff
@@ -107,33 +118,78 @@ void map_editor::check_open()
...
+
+    // Pirates and Barbarians are problematic, so we get rid of them
+    for (int i = 0; i <= ui.combo_players->count(); i++) {
+      if (ui.combo_players->itemText(i).toUtf8() == "Pirate") {
+        ui.combo_players->removeItem(i);
+      }
+      if (ui.combo_players->itemText(i).toUtf8() == "Barbarian") {
+        ui.combo_players->removeItem(i);
+      }
+    }
```
lmoureaux, line 157: This forgets the `Animal Kingdom` and will be problematic with translations or rulesets that redefine barbarian nations. Use `if (!is_barbarian(pplayer))` in the loop before.

client/editor/map_editor.cpp
```diff
@@ -107,33 +118,78 @@ void map_editor::check_open()
...
       auto sprite =
           *get_nation_flag_sprite(tileset, nation_of_player(pplayer));
-      ui.combo_players->addItem(sprite,
-                                _(qUtf8Printable(player_name(pplayer))));
+      ui.combo_players->addItem(
+          sprite, _(qUtf8Printable(nation_adjective_for_player(pplayer))));
     }
     players_iterate_end;
+
+    // Pirates and Barbarians are problematic, so we get rid of them
```
lmoureaux, line 149: Are they still problematic with `b` in `allowtake`? If yes that's issue-worthy.

client/editor/map_editor.cpp
```diff
@@ -107,33 +118,78 @@ void map_editor::check_open()
...
+ */
+void map_editor::player_changed(int index)
+{
+  struct city *capital;
+
+  ui.combo_players->setCurrentIndex(index);
+
+  players_iterate(pplayer)
+  {
+    if (ui.combo_players->hasFocus()) {
```
lmoureaux, line 177: Why do you require focus?

client/editor/map_editor.cpp
jwrober: Good call.

client/editor/map_editor.cpp
jwrober: Oh nice. I wasn't aware of the function.

client/editor/map_editor.cpp
jwrober: I guess not, removed

jwrober:
One section that isn't working all the time is the center on the capital city of the player.  I have this code in `map_editor::player_changed()`

```
  capital = player_primary_capital(client.conn.playing);
  if (capital->tile) {
    // Center on the tile
    queen()->mapview_wdg->center_on_tile(capital->tile);
  }
```

However, this does not always center the game on the city when you pick the nation from the combo box.  I tried adding this to the `players_iterate` block: 

```
  capital = player_primary_capital(pplayer);
```

but that doesn't really do anything at all. Any other ideas to get the player's capital city after the `/take` command is issued?

jwrober:
I think I am going to get rid of that capability then and make it a button or something when I get to cities.

PR #2557, opened 2025-03-12
Fix Coverity issues

### LINKED ISSUE #2375: New Coverity warnings
**Describe the bug**
Coverity is unhappy about the tileset and mapgen PRs (mostly)

image

**To Reproduce**
1. Open coverity report

**Expected behavior**
Fix discovered issues

 - OS: CI
 - Freeciv21 version: master

blabber:
Closes #2375

### DIFF
--- a/ai/default/aiunit.cpp
+++ b/ai/default/aiunit.cpp
@@ -13,6 +13,7 @@
 
 // utility
 #include "bitvector.h"
+#include "log.h"
 #include "rand.h"
 #include "registry.h"
 #include "registry_ini.h"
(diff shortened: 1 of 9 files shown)


ai/default/aiunit.cpp
```diff
@@ -796,7 +797,9 @@ int look_for_charge(struct ai_type *ait, struct player *pplayer,
         continue;
       }
       // Reduce want based on move cost.
-      def >>= move_cost / (2 * std::min(1, unit_move_rate(punit)));
+      int divisor = (2 * std::min(1, unit_move_rate(punit)));
```
lmoureaux, line 800: This looks like a logic error, the `min` will always return 1. Should likely be a `max`.

blabber: You are probably right, but I'd rather not change existing logic, when the goal is to fix a possible division by zero reported by a static analyzer.

XHawk87: The only reason that could cause a division by zero is this error in the first place :P

Issue #2547, opened 2025-03-08
Map Editor: Tile Tool Step 1


jwrober:
The Map Editor has a Tile Tool (`client/editor/tool_tile.cpp`), it needs many more capabilities:

- [x] `editor_tool_tile::select_tile()` doesn't do anything. Need to find the proper client or server function to change the mouse cursor to a universal selector so we can select any tile on the map.
- [x] When a tile is selected, populate a `struct tile` with all the properties of the tile
     - [x] Map X Coordinate
     - [x] Map Y Coordinate
     - [x] Native Map X Coordinate
     - [x] Native Map Y Coordinate
     - [x] Continent Number
     - [x] Terrain Type (e.g. Grassland, Jungle) with and icon/pixmap next to the label
     - [x] Resource (if any) with an icon/pixmap next to the label
     - [x] List of Specials - comma separated list for display is probably fine to start with (e.g. irrigation, mine, oil well, oil platform, pollution, hut, farmland, fallout, ruins)
     - [x] List of Roads - comma separated list for display is probably fine to start with
     - [x] List of Bases - comma separated list for display is probably fine to start with
     - [x] Tile Label
     - [x] Tile Visibility
- [x] Populate the labels on the form (not all labels are present at this time)
- [x] Disable universal selection
- [x] If a different tile is selected, then the labels should update

**NOTE**: Notice this task doesn't do anything with a city or units on the tile. That's later.

lmoureaux:
>  When a tile is selected, populate a struct tile with all the properties of the tile 

Sounds like you may want to store the tile ID and look this up from the map.

Another point:
* [ ] If tile information changes (i.e. from another client), the UI is updated accordingly.

PR #2558, opened 2025-03-12
Map Editor: Editor Tool Tile Phase 1

jwrober:
Part of #2547 

This PR gets us the population of the first 5 attributes of a selected tile. I have it in draft right now as I'm needing some help with Qt layouts. The StackedWidget wasn't working. I went back to a vertical layout, which "works", but doesn't look right.

### DIFF
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -274,7 +274,8 @@ void set_hover_state(const std::vector<unit *> &units,
                      enum unit_orders order)
 {
   fc_assert_ret(!units.empty()
-                || (state == HOVER_NONE || state == HOVER_DEBUG_TILE || state == HOVER_EDIT_TILE));
+                || (state == HOVER_NONE || state == HOVER_DEBUG_TILE
+                    || state == HOVER_EDIT_TILE));
   fc_assert_ret(state == HOVER_CONNECT || activity == ACTIVITY_LAST);
   fc_assert_ret((state == HOVER_GOTO || state == HOVER_GOTO_SEL_TGT)
                 || order == ORDER_LAST);
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -274,7 +274,7 @@ void set_hover_state(const std::vector<unit *> &units,
                      enum unit_orders order)
 {
   fc_assert_ret(!units.empty()
-                || (state == HOVER_NONE || state == HOVER_DEBUG_TILE));
+                || (state == HOVER_NONE || state == HOVER_DEBUG_TILE || state == HOVER_EDIT_TILE));
   fc_assert_ret(state == HOVER_CONNECT || activity == ACTIVITY_LAST);
   fc_assert_ret((state == HOVER_GOTO || state == HOVER_GOTO_SEL_TGT)
                 || order == ORDER_LAST);
@@ -1175,6 +1175,9 @@ void control_mouse_cursor(struct tile *ptile)
   case HOVER_GOTO_SEL_TGT:
   case HOVER_DEBUG_TILE:
     /* Select a tile to target / find targets on. */
+    mouse_cursor_type = CURSOR_SELECT;
+    break;
+  case HOVER_EDIT_TILE:
     mouse_cursor_type = CURSOR_SELECT;
     break;
   };
@@ -2466,6 +2469,8 @@ void do_map_click(struct tile *ptile, enum quickselect_type qtype)
         clear_hover_state();
       }
       return;
+    case HOVER_EDIT_TILE:
+      break;
     }
 
     clear_hover_state();
@@ -2749,6 +2754,9 @@ void key_cancel_action()
     keyboardless_goto_start_tile = nullptr;
     break;
   case HOVER_DEBUG_TILE:
+    clear_hover_state();
+    break;
+  case HOVER_EDIT_TILE:
     clear_hover_state();
     break;
   case HOVER_NONE:
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -1175,10 +1175,8 @@ void control_mouse_cursor(struct tile *ptile)
   case HOVER_ACT_SEL_TGT:
   case HOVER_GOTO_SEL_TGT:
   case HOVER_DEBUG_TILE:
-    /* Select a tile to target / find targets on. */
-    mouse_cursor_type = CURSOR_SELECT;
-    break;
   case HOVER_EDIT_TILE:
+    /* Select a tile to target / find targets on. */
     mouse_cursor_type = CURSOR_SELECT;
     break;
   };
@@ -2755,8 +2753,6 @@ void key_cancel_action()
     keyboardless_goto_start_tile = nullptr;
     break;
   case HOVER_DEBUG_TILE:
-    clear_hover_state();
-    break;
   case HOVER_EDIT_TILE:
     clear_hover_state();
     break;
--- a/client/control.cpp
+++ b/client/control.cpp
@@ -11,9 +11,9 @@
     \_____/ /                     If not, see https://www.gnu.org/licenses/.
       \____/        ********************************************************/
 
-#include <QSet>
-#include <QTimer>
-#include <algorithm>
+// self
+#include "control.h"
+
 // utility
 #include "bitvector.h"
 #include "fcintl.h"
@@ -27,8 +27,7 @@
 #include "path_finder.h"
 #include "unitlist.h"
 
-/* common/aicore */
-/* client/include */
+// client/include
 #include "citydlg_g.h"
 #include "dialogs_g.h"
 #include "mapctrl_g.h"
@@ -40,7 +39,6 @@
 #include "client_main.h"
 #include "climap.h"
 #include "climisc.h"
-#include "control.h"
 #include "goto.h"
 #include "governor.h"
 #include "options.h"
@@ -50,6 +48,13 @@
 #include "tileset/tilespec.h"
 #include "update_queue.h"
 #include "views/view_map.h"
+
+// Qt
+#include <QSet>
+#include <QTimer>
+
+// Std
+#include <algorithm>
 
 struct client_disband_unit_data {
   int unit_id;
--- a/client/editor.cpp
+++ b/client/editor.cpp
@@ -334,11 +334,7 @@ enum editor_tool_mode editor_tool_get_mode(enum editor_tool_type ett)
  */
 bool editor_is_active()
 {
-  if (editor != nullptr && can_conn_edit(&client.conn)) {
-    return true;
-  } else {
-    return false;
-  }
+  return editor != nullptr && can_conn_edit(&client.conn);
 }
 
 /**
--- a/client/editor.cpp
+++ b/client/editor.cpp
@@ -183,6 +183,8 @@ void editor_ruleset_changed()
  */
 void editor_init()
 {
+  fc_assert(editor != nullptr);
+
   editor = new editor_state[1]();
 
   tool_init(ETT_TERRAIN, _("Terrain"), ETF_HAS_VALUE | ETF_HAS_SIZE,
@@ -330,7 +332,7 @@ enum editor_tool_mode editor_tool_get_mode(enum editor_tool_type ett)
 /**
    Returns TRUE if the *client* is in edit mode.
  */
-bool editor_is_active() { return can_conn_edit(&client.conn); }
+bool editor_is_active() { editor != nullptr &&can_conn_edit(&client.conn); }
 
 /**
    Returns TRUE if the given tool should be made available to the user via
@@ -614,7 +616,7 @@ static void editor_grab_tool(const struct tile *ptile)
 /**
    Returns TRUE if the given tile has some objects with editable properties.
  */
-static inline bool can_edit_tile_properties(struct tile *ptile)
+bool can_edit_tile_properties(struct tile *ptile)
 {
   return ptile != nullptr;
 }
--- a/client/editor.cpp
+++ b/client/editor.cpp
@@ -183,8 +183,6 @@ void editor_ruleset_changed()
  */
 void editor_init()
 {
-  //fc_assert(editor == nullptr);
-
   editor = new editor_state[1]();
 
   tool_init(ETT_TERRAIN, _("Terrain"), ETF_HAS_VALUE | ETF_HAS_SIZE,
--- a/client/editor.cpp
+++ b/client/editor.cpp
@@ -332,7 +332,14 @@ enum editor_tool_mode editor_tool_get_mode(enum editor_tool_type ett)
 /**
    Returns TRUE if the *client* is in edit mode.
  */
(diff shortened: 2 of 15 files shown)


jwrober:
If clang check passes, I'll be super surprised

jwrober:
I'd recommend someone run an ASAN check on this. I'm going to do the same on my end.

jwrober:
Basic test plan:
- Open a local save game where you will have 'hack' level privileges
- Ctrl-Shift-E to open the main map editor widget (notice the layout is funky)
- You should be able to change the nation to edit by changing the nation in the players combo box
- Clicking the tile tool will open the editor for that object type. There is a pencil tool to select a tile and when clicked you should be able to click any tile (cities, ones with units, etc) and the base attributes should show on the left.

client/control.cpp
```diff
@@ -1177,6 +1178,9 @@ void control_mouse_cursor(struct tile *ptile)
     /* Select a tile to target / find targets on. */
     mouse_cursor_type = CURSOR_SELECT;
     break;
+  case HOVER_EDIT_TILE:
+    mouse_cursor_type = CURSOR_SELECT;
```
lmoureaux, line 1182: Why a separate `case` statement?

client/control.cpp
```diff
@@ -2751,6 +2757,9 @@ void key_cancel_action()
   case HOVER_DEBUG_TILE:
     clear_hover_state();
     break;
+  case HOVER_EDIT_TILE:
+    clear_hover_state();
```
lmoureaux, line 2761: Why a separate `case` statement?

client/editor.cpp
```diff
@@ -183,8 +183,6 @@ void editor_ruleset_changed()
  */
 void editor_init()
 {
-  fc_assert(editor == nullptr);
```
lmoureaux, line 186: If the editor is not null, it's already enabled, so you don't need to create a new one, no?

client/editor/map_editor.cpp
```diff
@@ -7,12 +7,13 @@
...
 #include "citydlg.h"        //map_editor::showEvent
 #include "client_main.h"    //map_editor::check_open
+#include "editor.h"         //map_editor::showEvent
 #include "fc_client.h"      //map_editor::check_open
 #include "minimap_panel.h"  //map_editor::showEvent
 #include "page_game.h"      //map_editor::showEvent
 #include "views/view_map.h" //map_editor::showEvent
 
-// common
+// Common
```
lmoureaux, line 16: This is the folder name
```suggestion
// common
```

client/editor/map_editor.cpp
```diff
@@ -33,7 +34,7 @@ map_editor::map_editor(QWidget *parent)
   setAutoFillBackground(true);
 
   // initialize in constructor
-  ett_wdg = new editor_tool_tile;
+  ett_wdg = new editor_tool_tile(0);
```
lmoureaux, line 37:
```suggestion
  ett_wdg = new editor_tool_tile(nullptr);
```

client/editor/map_editor.cpp
```diff
@@ -102,9 +102,6 @@ void map_editor::check_open()
     // Notify the server we are going into edit mode
     dsend_packet_edit_mode(my_conn, true);
 
-    // While editing, ensure we can take all possible nations
-    send_chat_printf("/set allowtake hHaAoObd");
```
lmoureaux, line 106: Hmm why removing it now?

client/mapctrl_common.cpp
```diff
@@ -370,5 +370,7 @@ void update_line(int canvas_x, int canvas_y)
   case HOVER_ACT_SEL_TGT:
   case HOVER_DEBUG_TILE:
     break;
+  case HOVER_EDIT_TILE:
```
lmoureaux, line 373: Why a separated `case` statement?

client/views/view_map_common.cpp
```diff
@@ -534,6 +534,8 @@ static void base_set_mapview_origin(float gui_x0, float gui_y0)
   case HOVER_ACT_SEL_TGT:
   case HOVER_DEBUG_TILE:
     break;
+  case HOVER_EDIT_TILE:
```
lmoureaux, line 537: Why a separate `case` statement?

common/networking/connection.cpp
```diff
@@ -471,8 +471,7 @@ const char *conn_description(const struct connection *pconn, bool is_private)
  */
 bool can_conn_edit(const struct connection *pconn)
 {
-  return (can_conn_enable_editing(pconn) && game.info.is_edit_mode
-          && (nullptr != pconn->playing || pconn->observer));
+  return (can_conn_enable_editing(pconn) && game.info.is_edit_mode);
```
lmoureaux, line 474: Hmm this means that a client connected to no player and also not observing can start editing (this is the state you are in when you connect after a delegation). Why is this required?

client/editor/map_editor.cpp
```diff
@@ -184,9 +187,21 @@ void map_editor::select_tool_tile()
...
+
+/**
+ * \brief Capture tile selected and pass tile structure to the editor tool
+ * tile widget
+ */
+void map_editor::tile_selected(struct tile *ptile)
+{
+  // Take the tile handed to us from mapctrl.cpp and pass it to the tile tool
+  // widget, which will do the real work.
+  ett_wdg->update_ett(ptile);
```
lmoureaux, line 206: How can we be sure that `ett_wdg` is active?

client/editor/tool_tile.cpp
```diff
@@ -8,8 +8,11 @@
 #include "editor/tool_tile.h"
 
 // Client
-#include "fcintl.h" //editor_tool_tile
-#include "fonts.h"  //editor_tool_tile
+#include "fcintl.h"         //editor_tool_tile::editor_tool_tile
+#include "fonts.h"          //editor_tool_tile::editor_tool_tile
+#include "mapctrl_common.h" //editor_tool_tile::select_tile
+
+bool is_tile_tool = false;
```
lmoureaux, line 15: Not a fan at all, this will not scale to multiple tools. Imagine you can end up with `is_tile_tool = true` and `is_terrain_tool = true`: what do you do?
Better have a variable in `map_editor` that knows which tool is currently active and calls it as needed.

client/editor/tool_tile.cpp
```diff
@@ -28,6 +31,8 @@ editor_tool_tile::editor_tool_tile(QWidget *parent)
   ui.setupUi(this);
 
   // Set default values and the font we want.
+  ui.value_continent->setText(_("N/A"));
```
lmoureaux, line 34: Side remark: not all instances of `N/A` translate the same, for instance in French I'd do:
* `tile: N/A` -> `tuile: aucune`
* `continent: N/A` -> `continent: aucun`

I translate to `aucun` ("none") because the French translation of N/A (`sans objet`) feels very formal and its abbreviation `s/o` is not very common.

client/editor/map_editor.cpp
```diff
@@ -184,9 +187,21 @@ void map_editor::select_tool_tile()
     ett_wdg_active = false;
   } else {
     ui.tbut_tool_tile->setToolTip(_("Close Tile Tool"));
-    ui.stackedWidget_tools->addWidget(ett_wdg);
+    ett_wdg = new editor_tool_tile(0);
```
lmoureaux, line 190:
```suggestion
    ett_wdg = new editor_tool_tile(nullptr);
```

lmoureaux, line 190: How comes we create the `editor_tool_tile` from two different places? Can we avoid it? Where do we delete it?

client/editor.cpp
jwrober: I removed because I always got an assert failure even when the was nullptr.

lmoureaux: That's not possible :thinking: At least not if the `editor` variable is null.

client/editor/map_editor.cpp
jwrober: because of this in mapctrl - `if (editor_is_active() && check_tile_tool())`

client/editor/tool_tile.cpp
jwrober: ok that sounds like a good idea. Would also help with the check in mapctrl.

client/editor/tool_tile.cpp
jwrober: Maybe use "Not Set" instead?

client/editor/map_editor.cpp
jwrober: I had the creation here at first and then got a coverity error so I moved it to the constructor. I added while troubleshooting the stacked widget. Let me see if I can remove it and also ensure I clear/free it when we close the dialog.

client/editor/tool_tile.cpp
lmoureaux: A dash would work. Maybe that's the solution for French :D

We may need a "click on a tile to view its properties" state for the whole widget?

client/editor/map_editor.cpp
lmoureaux: Setting it to nullptr in the constructor should make coverity happy I guess.

client/editor/tool_tile.cpp
jwrober: I'm going to use the editor tool type enum to cover this

client/editor/tool_tile.cpp
jwrober: I could add a status label, good idea

client/editor.cpp
jwrober: If I am reading this right and what the `fc_assert()` does, this needs to to be `!=`. If the editor hasn't been set (def at init) then it would be not equal. Let me play with that some.

client/editor/map_editor.cpp
jwrober: added another if()

client/editor/tool_tile.cpp
jwrober: got rid of it

client/editor/map_editor.cpp
jwrober: got rid of the extra "new" call

jwrober:
Code-wise I think I fixed all the review asks. I still want to fix the layout problem, but in a new PR after this one is merged.

image

qStackedWidget does show the tile editor sub-widget, but full functionality is lacking with what works with the vertical layout in use right now.

client/mapctrl.cpp
```diff
@@ -316,7 +325,10 @@ void map_view::shortcut_pressed(shortcut_id id)
     break;
 
   case SC_MAP_EDITOR:
-    queen()->map_editor_wdg->check_open();
+    // Editor is not meant to be used with observers
```
lmoureaux, line 328: Why?

lmoureaux (commented):
I can click on tiles through the editor panel. Set `Qt::WA_NoMousePropagation` to avoid this (see the city dialog).

I don't seem to need to click the "pick tile" button to pick a tile. I think it makes sense that whenever you click on a tile from within the tile tool, it switches to it. I don't see a need for the separate pick tile button.

If you close the editor while the tool tile is active, then reopen it, you need to enable the tool tile twice for it to show again.

client/editor.cpp
```diff
@@ -332,7 +332,14 @@ enum editor_tool_mode editor_tool_get_mode(enum editor_tool_type ett)
...
    Returns TRUE if the *client* is in edit mode.
  */
-bool editor_is_active() { return can_conn_edit(&client.conn); }
+bool editor_is_active()
+{
+  if (editor != nullptr && can_conn_edit(&client.conn)) {
+    return true;
+  } else {
+    return false;
+  }
```
lmoureaux, line 341:
```suggestion
  return editor != nullptr && can_conn_edit(&client.conn);
```

client/editor/tool_tile.cpp
```diff
@@ -52,11 +45,58 @@ editor_tool_tile::editor_tool_tile(QWidget *parent)
...
 void editor_tool_tile::select_tile()
 {
-  // TODO: have me do something real
-  return;
+  set_hover_state({}, HOVER_EDIT_TILE, ACTIVITY_LAST, nullptr, NO_TARGET,
+                  NO_TARGET, ACTION_NONE, ORDER_LAST);
+}
+
+/**
+ * \bried Close the Tile Tool
```
lmoureaux, line 79:
```suggestion
 * \brief Close the Tile Tool
```

client/editor/tool_tile.cpp
```diff
@@ -52,11 +45,58 @@ editor_tool_tile::editor_tool_tile(QWidget *parent)
...
+ */
+void editor_tool_tile::close_tool()
+{
+  clear_hover_state();
+  set_default_values();
+  editor_clear();
+}
+
+/**
+ * \bried Update the editor tool widget UI elements
```
lmoureaux, line 89:
```suggestion
 * \brief Update the editor tool widget UI elements
```

lmoureaux (commented):
And a few nitpicking code comments.

client/mapctrl.cpp
jwrober: Why is this a problem? The legacy fc editor operates the same. FWIW

https://github.com/longturn/freeciv21/blob/master/client/editor/map_editor.cpp#L169

barfs bad if you are an observer (of any kind) when invoked. So that is the main reason.

client/editor.cpp
jwrober: nice! this is why you are the code oracle :)

jwrober:
> I don't seem to need to click the "pick tile" button to pick a tile. I think it makes sense that whenever you click on a tile from within the tile tool, it switches to it. I don't see a need for the separate pick tile button.
> 
> If you close the editor while the tool tile is active, then reopen it, you need to enable the tool tile twice for it to show again.

This is not the use case I had in mind. There are going to to be many "tool" objects/classes after tile. I want to operate kind of like gimp where you have to be pretty specific on what you click on to get the result you want.

https://github.com/longturn/freeciv21/pull/2558/files#diff-a0b0e33304b930dbd78bd44c0b8a166865c855d3b872fa26c6397ccf41239543R200

is obviously not working as expected. How can I set another check so you don't allow the click unless `HOVER_EDIT_TILE` is also set?

client/mapctrl.cpp
lmoureaux: Ok, let's not try to outguess the old editor ^^

lmoureaux:
> This is not the use case I had in mind. There are going to to be many "tool" objects/classes after tile. I want to operate kind of like gimp where you have to be pretty specific on what you click on to get the result you want.

Getting the impression that you are doing two levels of Gimp: first one selects the tile inspector tool, then clicks on a button, then only on the map. In Gimp, once you activate a tool, you can immediately use it on the canvas.

I'll check the hover thing tomorrow. Might keep it as it is for now and do things more properly as we refactor mapctrl.

Issue #2464, opened 2024-12-23
Remove unused includes


lmoureaux:
**Describe the bug**
clang-tidy warns about many unused includes thorough the code. This slows down compilation.

**To Reproduce**
1. Run clang-tidy or use clangd
2. See unused include warnings in many files

**Expected behavior**
We minimize includes.

**Screenshots**
`city.cpp` seen by Kate with clangd enabled
![image](https://github.com/user-attachments/assets/f984673b-ffde-413b-a124-af8d8984faf8)

**Platform and version (please complete the following information):**
 - OS:  Linux
 - Freeciv21 version: 3.1-rc1

jwrober:
#2561 sets a check for includes. Does give a lot of warnings:

```
[467/468] Building CXX object tools/ruledit/CMakeFiles/freeciv21-ruledit.dir/validity.cpp.o
/home/jwrober/GitHub/freeciv21/tools/ruledit/validity.cpp:31:10: warning: no header providing "universal" is directly included [misc-include-cleaner]
   15 |   struct universal *uni;
      |          ^
/home/jwrober/GitHub/freeciv21/tools/ruledit/validity.cpp:39:52: warning: no header providing "effect" is directly included [misc-include-cleaner]
   15 | static bool effect_list_universal_needed_cb(struct effect *peffect,
      |                                                    ^
/home/jwrober/GitHub/freeciv21/tools/ruledit/validity.cpp:45:16: warning: no header providing "R__" is directly included [misc-include-cleaner]
   15 |     cbdata->cb(R__("Effect"), cbdata->requirers_data);
      |                ^
```

lmoureaux:
We'll need a CI check to prevent reintroducing mistakes after the cleanup.

PR #2572, opened 2025-03-22
Includes refactor: utility 1

jwrober:

Part of #2464 

Other than Qt still issuing warnings, all other missing include warnings are gone.


lmoureaux (approved):
Since you are touching file headers: should we also start a project to standardize them to SPDX format? (Not asking for it in this PR.)

lmoureaux:
Should we add the `// self` comment to the docs? (For me it's obvious that the one singled out include is self, but I don't mind the comment.)

jwrober:
> Should we add the `// self` comment to the docs? (For me it's obvious that the one singled out include is self, but I don't mind the comment.)

I was planning an update to the docs once I get `utility` done before I get into `common` and the others.

PR #2577, opened 2025-03-24
Remove deprecated qAsConst

lmoureaux:
It has been replaced by std::as_const in Qt 6, but the change can already be done in Qt 5. This was done using a simple search-and-replace.


jwrober:
Looks like Codacy isn't happy. I think we need to fix the issue it raised.

lmoureaux:
Fixed, apparently.

lmoureaux:
Codacy is even less happy now :sob:

PR #2581, opened 2025-03-24
Add terrain help links to tile info popup

blabber:
To avoid polluting `common/tile.{cpp,h}` with client headers (`client/helpdlg.h`), I have introduced an intermediate step with `tile_get_info`, which allows us to manipulate the string representation of the terrain information.


common/tile.cpp
```diff
@@ -761,28 +763,56 @@ bool tile_apply_activity(struct tile *ptile, Activity_type_id act,
    Add one entry about pollution situation to buffer.
    Return if there has been any pollution (even prior calling this)
  */
-static bool tile_info_pollution(char *buf, int bufsz,
-                                const struct tile *ptile,
-                                struct extra_type *pextra, bool prevp,
+static bool tile_info_pollution(QString &buf, QString pollution, bool prevp,
```
lmoureaux, line 766: Pass by const reference to avoid a copy
```suggestion
static bool tile_info_pollution(QString &buf, const QString &pollution, bool prevp,
```

client/text.cpp
```diff
@@ -164,14 +165,14 @@ const QString popup_info_text(struct tile *ptile, bool with_links)
   int tile_x, tile_y, nat_x, nat_y;
   bool first;
 
-  std::function<QString(const char *name, help_page_type hpt)> maybe_link;
+  std::function<QString(QString name, help_page_type hpt)> maybe_link;
   if (with_links) {
-    maybe_link = [](const char *name, help_page_type hpt) -> QString {
-      return create_help_link(name, hpt);
+    maybe_link = [](QString name, help_page_type hpt) -> QString {
+      return create_help_link(name.toStdString().c_str(), hpt);
```
lmoureaux, line 171: This could matter in case the system encoding isn't UTF-8 and `std::string` uses a weird encoding:
```suggestion
      return create_help_link(qUtf8Printable(name), hpt);
```

client/text.cpp
```diff
@@ -164,14 +165,14 @@ const QString popup_info_text(struct tile *ptile, bool with_links)
...
     };
   } else {
-    maybe_link = [](const char *name, help_page_type _) -> QString {
-      return QString(name);
+    maybe_link = [](QString name, help_page_type _) -> QString {
```
lmoureaux, line 174: Avoid a copy:
```suggestion
    maybe_link = [](const QString &name, help_page_type _) -> QString {
```

client/text.cpp
```diff
@@ -187,8 +188,23 @@ const QString popup_info_text(struct tile *ptile, bool with_links)
...
-  str += QString(_("Terrain: %1")).arg(tile_get_info_text(ptile, true, 0))
-         + qbr();
+
+  {
+    struct tile_info info = tile_get_info(ptile);
+    info.terrain = maybe_link(info.terrain, HELP_TERRAIN);
+    if (!info.resource.isEmpty()) {
+      info.resource = maybe_link(info.resource, HELP_EXTRA);
+    }
+    for (auto &extra : info.extras) {
```
lmoureaux, line 198: Avoid an expensive detach (delayed copy) of the container:
```suggestion
    for (auto &extra : std::as_const(info.extras)) {
```
(or set the whole `tile_info` to `const` when declaring it above)

client/text.cpp
```diff
@@ -187,8 +188,23 @@ const QString popup_info_text(struct tile *ptile, bool with_links)
...
+      extra = maybe_link(extra, HELP_EXTRA);
+    }
+    for (auto &nuisance : info.nuisances) {
```
lmoureaux, line 201: Avoid an expensive detach (delayed copy) of the container:
```suggestion
    for (auto &nuisance : std::as_const(info.nuisances)) {
```

common/tile.cpp
```diff
@@ -796,48 +826,40 @@ static bool tile_info_pollution(char *buf, int bufsz,
      "Hills (Coals)"
      "Hills (Coals) [Pollution]"
  */
-const char *tile_get_info_text(const struct tile *ptile,
+const char *tile_get_info_text(const struct tile_info info,
```
lmoureaux, line 829: Pass by reference to avoid a copy, and no need to repeat `struct` when not ambiguous:
```suggestion
const char *tile_get_info_text(const tile_info &info,
```

common/tile.h
```diff
@@ -170,3 +170,16 @@ bool tile_virtual_check(const tile *vtile);
 
 bool tile_set_label(struct tile *ptile, const char *label);
 bool tile_is_placing(const struct tile *ptile);
+
+/**
+  Contains some information about a tile as human readable translated
+  string.
+ */
+struct tile_info {
```
lmoureaux, line 178: I'm wondering if we shouldn't use a constructor instead of `tile_get_info`. That is,
```suggestion
struct tile_info {
    explicit tile_info(const tile *ptile);
```
This would be used as `tile_get_info_text(tile_info(ptile), ...)`.

lmoureaux (commented):
The split makes sense, if a bit overkill for what's essentially debug logging on the server side. I have a few Qt/C++ comments.

common/tile.cpp
blabber: Done

common/tile.h
blabber: I wasn't aware of struct constructors. But I like it, done.

common/tile.cpp
blabber: Done

client/text.cpp
blabber: Done

client/text.cpp
blabber: Done.

client/text.cpp
blabber: This loop has the sole purpose of manipulating the contents of the `QStringList`. This is no longer possible when working with `const`.

client/text.cpp
blabber: This loop has the sole purpose of manipulating the contents of the `QStringList`. This is no longer possible when working with `const`.

blabber:
@lmoureaux, I followed your advice in most cases. I also sneaked a fix for a bug I noticed while testing in (327cf71c0d).

lmoureaux (approved):

Issue #1834, opened 2023-02-23
Observing a player completely botches the governers


psampathkumar:
**Describe the bug**
When you observe a player and go back to your player, the info window of a city and governors are completely screwed. 

**To Reproduce**
1. Play a game
2. Set some governors for a city
3. Observe another player
4. Go back to your city
5. See the info panel of the city and notice now the governer also stopped working

**Expected behavior**
My nation works as is, without being affected by who I observed before. 

**Screenshots**
You see situations like this which dont match with the tiles, the city is working
![image](https://user-images.githubusercontent.com/30164619/220992764-9d3538dc-3832-4837-a147-db9fc8747f9e.png)

**Platform and version (please complete the following information):**
 - OS: linux
 - Freeciv21 version:master
 - Ruleset/Longturn game (if applicable): Predominantly seen in LT76

jwrober:
Closed via #2301

lmoureaux:
TriClad reports it's not fixed in https://github.com/longturn/freeciv21/commit/336c424276f469bbf53a241985447f5746e72096

jwrober:
This issue is also caused if are controlling more than one players in a game. So `/take` actions have the same issue as `/observe`.

PR #2582, opened 2025-03-25
Fix disabling the last governor

### LINKED ISSUE #2472: City Governors are "Sticky"
**Describe the bug**
CMA's assigned to a city don't stay "unset" after TC. You can set a CMA to optimize tiles and then turn it off and sometimes the CMA will be enabled to the same setting after TC. There is also a situation (hard to pin down) when you set a CMA, TC happens and you unset it, then it comes back. It feels "random", but there is probably some logic to it.

**To Reproduce**
I've had difficulty is being able to reproduce this issue with clarity. Keeping boiler plate steps for future update if we figure it out.

**Expected behavior**
Obviously, If I set an CMA, keep it. If it unset, then by golly keep it unset!

**Screenshots**
N/A

 - OS: Linux - Could be everyone and not Linux specific
 - Freeciv21 version: 3.1-rc.1
 - Ruleset/Longturn game (if applicable): LTT (League B3) and LTX (LT85) - I don't think this is a rules thing.

**Additional context**
Maybe #1863 is a solution?

lmoureaux:
Disabling the governor in the last city that had one was broken because the attribute block was not sent to the server when empty. Now it is and removing the last governor is working again.

Closes #2472.

Backport requested -- this is a 3.1 blocker.

### DIFF
--- a/client/attribute.cpp
+++ b/client/attribute.cpp
@@ -290,10 +290,6 @@ void attribute_flush()
     return;
   }
 
-  if (0 == attribute_hash->size()) {
-    return;
-  }
-
   pplayer->attribute_block.clear();
   serialize_hash(attribute_hash, pplayer->attribute_block);
   send_attribute_block(pplayer, &client.conn);


jwrober:
I'm going to test this some, however do you think the patch also resolves #1834?

lmoureaux:
No it doesn't fix #1834.

jwrober (approved):

PR #2587, opened 2025-03-25
Includes refactor: utility 3

jwrober:

Part of #2464

Other than Qt still issuing warnings, all other missing include warnings are gone.

This completes `/utility`


jwrober:
Ok apparently I ticked off MSYS2. I need to do more testing there.

PR #2594, opened 2025-03-30
Store demographics values at TC and use them for the demographics report

### LINKED ISSUE #2114: Some demographics aren't evaluated at turn change
**Describe the bug**
Some demographics aren't (fully) evaluated at turn change. A very famous example of this is Literacy, which increases dramatically when a nation loses cities.

**To Reproduce**
1. Don't be first in Literacy
2. Have an embassy with the leading nation in Literacy
3. Attack that nation, taking away many cities
4. See their Literacy demographic going above 100%, which is normally not possible

**Expected behavior**
Demographics are always evaluated at TC and don't change during the turn.

 - OS: all
 - Freeciv21 version: all so far
 - Ruleset/Longturn game (if applicable): all

lmoureaux:
This fixes #2114 by calculating (all) demographics at turn change and saving the values in `player->score.demographics`. They are also stored in saves so a save/load cycle doesn't lead to a recalculation.

### DIFF
--- a/common/player.h
+++ b/common/player.h
@@ -107,6 +107,7 @@ struct player_score {
   int game; // Total score you get in player dialog.
 
   /// Cached demographic values, indexed by their untranslated name.
+  /// \note Can be incomplete when loading from old saves.
   std::map<std::string, int> demographics;
 };
 
--- a/common/player.h
+++ b/common/player.h
@@ -19,6 +19,10 @@
 #include "tech.h"
 #include "traits.h"
 
+// std
+#include <map>
+#include <string>
+
 #define PLAYER_DEFAULT_TAX_RATE 0
 #define PLAYER_DEFAULT_SCIENCE_RATE 100
 #define PLAYER_DEFAULT_LUXURY_RATE 0
@@ -101,6 +105,9 @@ struct player_score {
                      * by combat or otherwise. */
   int culture;
   int game; // Total score you get in player dialog.
+
+  /// Cached demographic values, indexed by their untranslated name.
+  std::map<std::string, int> demographics;
 };
 
 struct player_ai {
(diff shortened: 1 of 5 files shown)


common/player.h
```diff
@@ -101,6 +105,10 @@ struct player_score {
                      * by combat or otherwise. */
   int culture;
   int game; // Total score you get in player dialog.
+
+  /// Cached demographic values, indexed by their untranslated name.
+  /// \note Can be incomplete when loading from old saves.
+  std::map<std::string, int> demographics;
```
jwrober, line 111: A nit, but 3 slashes is out of form for comment

jwrober (approved):
Looks good to me. Seems to work with the testing I did. One small change on comments. 

Question - Do we need to worry about `savegame2.cpp`? or is this only or 3+?

jwrober:
Codacy also seems unhappy

lmoureaux:
> Question - Do we need to worry about savegame2.cpp? or is this only or 3+?

savegame2.cpp is for loading saves created with an antiquated version of Freeciv. We can no longer save in this format, so these files will never contain the new fields.

server/report.cpp
```diff
@@ -1062,8 +1062,25 @@ static void dem_line_item(char *outptr, size_t out_size,
                           struct player *pplayer, const demographic &demo,
                           bv_cols selcols)
 {
+  int value = -1;
+  if (!pplayer || BV_ISSET(selcols, DEM_COL_QUANTITY)
+      || BV_ISSET(selcols, DEM_COL_RANK)
+      || BV_ISSET(selcols, DEM_COL_BEST)) {
+    if (pplayer->score.demographics.count(demo.name()) > 0) {
```
lmoureaux, line 1069: Codacy wants to say that global observers requesting demographics crash the server.

lmoureaux: For reference, one gets more context with `clang-tidy -p compile_commands.json server/report.cpp --checks=clang-analyzer-core.NullDereference`.

PR #2595, opened 2025-03-31
Map Editor: Tool Tile Phase 2

jwrober:
This phase adds most of the tile's features. We have a base terrain sprite with text for the rest. Part of #2547

### DIFF
--- a/client/editor/tool_tile.cpp
+++ b/client/editor/tool_tile.cpp
@@ -4,14 +4,20 @@
 // self
 #include "editor/tool_tile.h"
 
+// common
+#include "nation.h"
+
 // client
-#include "editor.h"         //editor_tool_tile::update_ett
-#include "fcintl.h"         //editor_tool_tile::editor_tool_tile
-#include "fonts.h"          //editor_tool_tile::editor_tool_tile
-#include "mapctrl_common.h" //editor_tool_tile::select_tile
+#include "editor.h"
+#include "fcintl.h"
+#include "fonts.h"
+#include "mapctrl_common.h"
+#include "tileset/tilespec.h"
+#include "views/view_map_common.h"
 
 // Qt
 #include <QPainter>
+#include <QPoint>
 
 /**
  *  \class editor_tool_tile
@@ -56,14 +62,59 @@ void editor_tool_tile::set_default_values()
   // Set default values and the font we want.
   ui.value_continent->setText(_("-"));
   ui.value_continent->setFont(value_font);
+  ui.value_continent->setAlignment(Qt::AlignLeft);
+
+  ui.value_owner->setText(_("-"));
+  ui.value_owner->setFont(value_font);
+  ui.value_owner->setAlignment(Qt::AlignLeft);
+
   ui.value_x->setText(_("-"));
   ui.value_x->setFont(value_font);
+  ui.value_x->setAlignment(Qt::AlignLeft);
+
   ui.value_y->setText(_("-"));
   ui.value_y->setFont(value_font);
+  ui.value_y->setAlignment(Qt::AlignLeft);
+
   ui.value_nat_x->setText(_("-"));
   ui.value_nat_x->setFont(value_font);
+  ui.value_nat_x->setAlignment(Qt::AlignLeft);
+
   ui.value_nat_y->setText(_("-"));
   ui.value_nat_y->setFont(value_font);
+  ui.value_nat_y->setAlignment(Qt::AlignLeft);
+
+  ui.value_terrain->setText(_("-"));
+  ui.value_terrain->setFont(value_font);
+  ui.value_terrain->setAlignment(Qt::AlignLeft);
+
+  ui.pixmap_terrain->setText(_("-"));
+  ui.pixmap_terrain->setFont(value_font);
+  ui.pixmap_terrain->setAlignment(Qt::AlignLeft);
+
+  ui.value_resource->setText(_("-"));
+  ui.value_resource->setFont(value_font);
+  ui.value_resource->setAlignment(Qt::AlignLeft);
+
+  ui.value_road->setText(_("-"));
+  ui.value_road->setFont(value_font);
+  ui.value_road->setAlignment(Qt::AlignLeft);
+
+  ui.value_infra->setText(_("-"));
+  ui.value_infra->setFont(value_font);
+  ui.value_infra->setAlignment(Qt::AlignLeft);
+
+  ui.value_base->setText(_("-"));
+  ui.value_base->setFont(value_font);
+  ui.value_base->setAlignment(Qt::AlignLeft);
+
+  ui.value_hut->setText(_("-"));
+  ui.value_hut->setFont(value_font);
+  ui.value_hut->setAlignment(Qt::AlignLeft);
+
+  ui.value_nuisance->setText(_("-"));
+  ui.value_nuisance->setFont(value_font);
+  ui.value_nuisance->setAlignment(Qt::AlignLeft);
 }
 
 /**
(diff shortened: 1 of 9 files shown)


client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+    // tile nuisance (pollution, fallout)
+    ui.value_nuisance->setText(get_tile_nuisance_name(ptile));
   }
-  clear_hover_state();
+}
+
+/**
+ * \brief Return a string of the final road on a tile
+ */
+const QString editor_tool_tile::get_tile_road_name(struct tile *ptile)
```
lmoureaux, line 198: I think all the `get_tile_xxx_name` functions could be combined into something like this:
```cpp
QString editor_tool_tile::get_tile_extra_text(
    const tile *ptile, const std::vector<extra_cause> &causes) const
{
  QVector<QString> names;
  for (auto cause : causes) {
    extra_type_by_cause_iterate(EC_ROAD, pextra)
    {
      if (tile_has_extra(ptile, pextra)) {
        names.push_back(extra_name_translation(pextra));
      }
    }
    extra_type_by_cause_iterate_end;
  }
  if (names.isEmpty()) {
    return QString();
  }
  return strvec_to_and_list(names);
}
```
To be used as:
```cpp
auto text = get_tile_extra_text(tile, {EC_ROAD});
ui.value_road->setText(text.isEmpty() ? Q_("?road:None") : text);
```

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+{
+  QString sroad = nullptr;
+  extra_type_by_cause_iterate(EC_ROAD, pextra)
+  {
+    if (tile_has_extra(ptile, pextra)) {
+      sroad = extra_name_translation(pextra);
```
lmoureaux, line 204: If you want rails to hide roads, you should check `hidden_by(pextra->hidden_by, sroad->id)`. As it is now, rivers hide roads. I'd suggest to list them all.

lmoureaux, line 204: Turn into hyperlinks with `create_help_link()`. You'll also need something like this on the label:
```cpp
  label->setTextFormat(Qt::RichText);
  label->setTextInteractionFlags(Qt::TextBrowserInteraction);
  label->setWordWrap(true);

  connect(label, &QLabel::linkActivated, follow_help_link);
```
The `label->set*` can likely be edited in Designer.

client/editor/tool_tile.cpp
```diff
@@ -56,14 +62,59 @@ void editor_tool_tile::set_default_values()
   // Set default values and the font we want.
   ui.value_continent->setText(_("-"));
   ui.value_continent->setFont(value_font);
```
lmoureaux, line 64: I didn't notice before that you were using the chat font. I don't see a need for changing it at all, could we use the standard one?

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
     ui.value_y->setNum(index_to_map_pos_y(ptile->index));
     ui.value_nat_x->setNum(index_to_native_pos_x(ptile->index));
     ui.value_nat_y->setNum(index_to_native_pos_y(ptile->index));
+
+    // tile resource
+    if (ptile->resource) {
+      struct extra_type *res = ptile->resource;
+      ui.value_resource->setText(extra_name_translation(res));
+    } else {
+      ui.value_resource->setText(_("None"));
```
lmoureaux, line 179:
```suggestion
      ui.value_resource->setText(Q_("?resource:None")));
```

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+
+  if (sinfra != nullptr) {
+    return sinfra;
+  } else {
+    return "None";
+  }
+}
+
+/**
+ * \brief Return a strong of the any nuisances on a tile
```
lmoureaux, line 246:
```suggestion
 * \brief Return a string of the any nuisances on a tile
```

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+
+  if (sroad != nullptr) {
+    return sroad;
+  } else {
+    return "None";
+  }
+}
+
+/**
+ * \brief Return a strong of the added infrastructure on a tile
```
lmoureaux, line 217:
```suggestion
 * \brief Return a string of the added infrastructure on a tile
```

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
   }
-  clear_hover_state();
+}
+
+/**
+ * \brief Return a string of the final road on a tile
+ */
+const QString editor_tool_tile::get_tile_road_name(struct tile *ptile)
+{
+  QString sroad = nullptr;
```
lmoureaux, line 200: Avoid assigning null to a string, it's only supported for legacy reasons. This does the same:
```suggestion
  QString sroad;
```

client/editor/tool_tile.cpp
lmoureaux, line 198: * `const` does nothing on a type return by value (no `*` or `&`)
* You don't want to accidentally modify the tile, so declare it `const`
* You don't want to accidentally modify the `editor_tool_tile`, so also make the function `const`
```suggestion
QString editor_tool_tile::get_tile_road_name(const tile *ptile) const
```
(with corresponding changes in the header)

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+  extra_type_by_cause_iterate(EC_ROAD, pextra)
+  {
+    if (tile_has_extra(ptile, pextra)) {
+      sroad = extra_name_translation(pextra);
+    }
+  }
+  extra_type_by_cause_iterate_end;
+
+  if (sroad != nullptr) {
```
lmoureaux, line 209: Null strings are only there for compatibility. Equivalent to:
```suggestion
  if (!sroad.isEmpty()) {
```

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+  }
+}
+
+/**
+ * \brief Return a pixmap of the selected tile.
+ * TODO: Add more layers over time, right now we show the base tile and any
+ * resource We don't show any other infrastructure extras, such as
+ * roads/rivers, mines, irrigation, bases, or pollution.
+ */
+QPixmap editor_tool_tile::get_tile_sprites(struct tile *ptile,
```
lmoureaux, line 322:
```suggestion
QPixmap editor_tool_tile::get_tile_sprite(const tile *ptile) const
```
I guess `terrain` is always `tile_terrain(ptile)` so we don't need it. It also returns a single pixmap.

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+                                           struct terrain *terrain)
+{
+  int width = tileset_full_tile_width(tileset);
+  int height = tileset_full_tile_height(tileset);
+  int canvas_y = height - tileset_tile_height(tileset);
+
+  QPixmap *tile_pixmap;
+  tile_pixmap = new QPixmap(width, height);
+  tile_pixmap->fill(Qt::transparent);
```
lmoureaux, line 331: Let's avoid playing with pointers:
```suggestion
  QPixmap tile_pixmap(width, height);
  tile_pixmap.fill(Qt::transparent);
```

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+  // roads, etc.
+  for (int i = 0; i < 3; ++i) {
+    auto sprites =
+        fill_basic_terrain_layer_sprite_array(tileset, i, terrain);
+    put_drawn_sprites(tile_pixmap, QPoint(0, canvas_y), sprites, false);
+  }
+
+  // Add any resource on the tile
+  // FIXME: fill_basic_extra_sprite_array() does not work with all EC_* cause
+  // codes.
```
lmoureaux, line 343: Does this warrant an issue?

client/tileset/tilespec.h
```diff
@@ -137,6 +137,9 @@ tileset_get_layers(const struct tileset *t);
 // Gfx support
 QPixmap *load_sprite(struct tileset *t, const QStringList &possible_names,
                      bool required, bool verbose = true);
+QPixmap *terrain_canvas(struct terrain *terrain,
+                        const struct extra_type *resource = nullptr,
+                        enum extra_cause cause = EC_COUNT);
```
lmoureaux, line 142: Nothing against moving it here, but it looks like you're not using this function in the end?

client/editor/tool_tile.cpp
```diff
@@ -86,17 +137,224 @@ void editor_tool_tile::close_tool()
...
+    return shut;
+  } else {
+    return "None";
+  }
+}
+
+/**
+ * \brief Return a string of any bases on a tile
+ */
+const QString editor_tool_tile::get_tile_base_name(struct tile *ptile)
```
lmoureaux, line 298: This doesn't seem to work on a fogged tile with a fortress.

lmoureaux (changes requested):
It works quite well. Spotted some edge cases with extras on a mid-game map.

I think a great deal of code duplication could be avoided by merging the `get_tile_xxx_name` functions. Translation support also needs improvements (the docs have an explicit example for [`None`](https://longturn.readthedocs.io/en/latest/Coding/internationalization.html#translating-text)).

client/editor/tool_tile.cpp
jwrober: #2622

client/tileset/tilespec.h
jwrober: I did end up not using it due to the bug

client/editor/tool_tile.cpp
jwrober: I like this idea, but I think I want to come back to it in phase 3 along with fixing the fogged tile issue you found. I also don't get a tile label yet that was also planned for phase 3.

client/editor/tool_tile.cpp
jwrober: Ideally I'd like to show a string in the label like `river->road->railroad->maglev`. Ideas on how to do that? (re include in phase 3)

client/editor/tool_tile.cpp
lmoureaux: Use `&pixmap` to get the address of the variable on the stack.

client/editor/tool_tile.cpp
jwrober: I opened an issue for this

lmoureaux (approved):

PR #2597, opened 2025-04-02
Includes Refactor: Common 1 (aicore)

jwrober:
See #2464. This set includes common/aicore. There are some NOLINT comments that I am hoping to be able to come back and correct after all the other common includes are refactored.

PR #2599, opened 2025-04-02
Utility Include: Re-look at utility header files

jwrober:
As part of learning something new in #2597 taking a re-look at the `.h` files in the `utility` directory to ensure we are doing full IWYU there.

See #2464


lmoureaux:
Should update the docs accordingly

PR #2600, opened 2025-04-03
Docs: Add more tips for IWYU and clang-tidy

jwrober:
See #2599. I also removed an item in `.clang-tidy` that isn't supported by 17


lmoureaux (approved):
Rebase and let's merge!

jwrober:
looks like I need a rebase

PR #2617, opened 2025-04-08
Anishinaabe

hugomflavio:
The red background seems to be more common.

Stylized the proportions a bit to make the symbol more visible.


lmoureaux:
The new files are significantly larger. Can you compress them with `pngquant`?

jwrober:
> Can do. Which parameters should I set on pngquant?

When I use the tool I just give the name of the image to squash and leave rest as default

hugomflavio:
hm I tried that but the space savings were minimal (9kb to 8.8kb on one, 8.7kb to 8.2kb on a second one)

hugomflavio:
that did it, thanks!

lmoureaux (approved):
Let's squash the commits to avoid introducing the larger files in the history

PR #2628, opened 2025-04-11
Update some functions deprecated in Qt6

lmoureaux:
This updates some functions that are deprecated in Qt 6 to their new name. I only did a few easy ones that required no thinking.


lmoureaux:
Looks like the QCheckBox deprecation is too recent.

PR #2645, opened 2025-04-18
Fix one more -Wmaybe-uninitialized

lmoureaux:
Compiling stable in debug mode was failing with another -Wmaybe-uninitialized false positive. Ensure the compiler knows the value is initialized.

Backport requested.


jwrober (approved):
Backport approved

PR #2646, opened 2025-04-19
[stable] Update vcpkg to 2025.04.09 to fix macOS builds

lmoureaux:
This fixes the build but now we get an ARM-only dmg. I haven't been able to change it to x86 or to create a universal binary. It seems vcpkg doesn't support the latter.

Binaries for x86 Mac is a release blocker. Older macs cannot execute ARM code, while newer macs can execute x86.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -197,7 +197,7 @@ jobs:
       - uses: lukka/run-vcpkg@v11
         name: Install dependencies
         with:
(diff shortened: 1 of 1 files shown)


jwrober (changes requested):
I think we need to set the triplet by hand - https://learn.microsoft.com/en-us/vcpkg/users/triplets. The page has a section for macOS specifics

lmoureaux:
I spent a few hours trying to make that work last night before giving up. Let's do it later.

jwrober (approved):
I'm good with that since we have an open issue to track against.

Issue #2104, opened 2023-12-31
Move away from SDL2 and use Qt Sound/Multimedia instead


jwrober:
**Is your feature request related to a problem? Please describe.**
We have had some recent issues with SDL2 and SDL2_mixer autodetection and MacOS. 

**Describe the solution you'd like**
@lmoureaux suggested that we move away from this and use native Qt for this. Good to include as part of the migration to Qt6 #1198

**Describe alternatives you've considered**
None. Removing a dependency on an external library is always a good idea.

**Additional context**
None

lmoureaux:
> None. Removing a dependency on an external library is always a good idea.

Note that we'd technically be replacing the dependency. The main advantage is that everything would be more Qt-centric.

lmoureaux:
Why migrate now: The Qt Multimedia API was overhauled in Qt6 and the SDL_mixer API was overhauled in SDL3.

PR #2648, opened 2025-04-19
Create the Qt audio plugin

lmoureaux:
We are moving away from SDL for audio to reduce the number of dependencies. This PR introduces a new audio plugin based on Qt Multimedia.

It has some differences compared to SDL:
* It doesn't fade audio out when stopping it.
* WAVE is the only supported format for sound effects. Music supports everything supported by the platform.

See https://github.com/longturn/freeciv21/issues/2104.

Testing: switch to the Qt plugin for your audio and play. You shouldn't notice any difference.

The plan is to phase out SDL before 3.2, but keep both plugins active until we get more reports from non-Linux platforms.


client/audio/audio.cpp
```diff
@@ -25,6 +25,7 @@ received a copy of the GNU General Public License along with Freeciv21.
 
 // client
 #include "audio_none.h"
+#include "audio_qt.h"
```
jwrober, line 28: Do we need an `#ifdef` for Qt like we have for SDL?

client/audio/audio_qt.cpp
```diff
@@ -0,0 +1,189 @@
+// SPDX-License-Identifier: GPLv3-or-later
+// SPDX-FileCopyrightText: Freeciv authors
+// SPDX-FileCopyrightText: Freeciv21 authors
```
jwrober, line 3: In my other IWYU work, I've been combining these two into a single line
`// SPDX-FileCopyrightText: Freeciv21 and Freeciv Contributors`

client/audio/audio_qt.cpp
```diff
@@ -0,0 +1,189 @@
...
+    current_callback = cb;
+
+    player->play();
+
+    qDebug() << "Playing music file" << fullpath;
+
+  } else {
+    // Non-repeating media is used for sound effects. Latency is more
+    // important for this, so we use QSoundEffect.
+    if (cache.count(fullpath) == 0) {
```
jwrober, line 74: How much is the latency different between the `QSoundEffect` an the `QMediaPlayer`? The move to WAV files seems problematic.

client/audio/audio_qt.h
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: GPLv3-or-later
+// SPDX-FileCopyrightText: Freeciv authors
+// SPDX-FileCopyrightText: Freeciv21 authors
```
jwrober, line 3: In my other IWYU work, I've been combining these two into a single line
`// SPDX-FileCopyrightText: Freeciv21 and Freeciv Contributors`

cmake/FreecivDependencies.cmake
```diff
@@ -116,7 +116,13 @@ add_subdirectory(dependencies/sol2)
...
 
 # Compression
-find_package(KF6Archive REQUIRED)
+include(FetchContent)
+FetchContent_Declare(KF6Archive
+  URL https://download.kde.org/stable/frameworks/6.13/karchive-6.13.0.tar.xz
+  URL_HASH SHA256=3c9b5dcf3abdfe2761e2153d70d9d667f1ff0fd2f6c80addba7549da954fcc90
+  FIND_PACKAGE_ARGS NAMES KF6Archive
+)
+FetchContent_MakeAvailable(KF6Archive)
```
jwrober, line 125: This seems out of place. Are you thinking of compiling kf6archive as an external dep like we do the font for the time being? If so, this should be a separate PR

jwrober (commented):
Some comments to start. I need to test this out on my local.

jwrober:
Works for me on my Fedora 41 local. I want to test on Msys2 next

client/audio/audio.cpp
lmoureaux: Only if we make the plugin optional.

client/audio/audio_qt.cpp
lmoureaux: WAV files add 1MB to the repo and installer, which is significant but not too much.

cmake/FreecivDependencies.cmake
lmoureaux: Sorry, it leaked from work I've been doing in parallel. Will remove.

lmoureaux:
As discussed on Discord, I'll add checks for codec support and playback (if possible at zero volume, I don't want tests to start playing random beeps and honks). Qt uses its ffmpeg backend by default, which "should" provide ogg/vorbis support in a platform-independent way.

On FreeBSD one needs to install ogg and vorbis plugins for gstreamer separately. This might be because the gstreamer backend is used (which is not recommended by Qt for desktop but might be FreeBSD's choice).

blabber:
Regarding FreeBSD: I checked the `qt6-multimedia` port and it is build with both backends, `gstreamer` and `ffmpeg`, by default. I didn't find any setting that defines the default backend, but `gstreamer` seems to be preferred on FreeBSD. You can force the use of `ffmpeg` by setting the environment variable `QT_MEDIA_BACKEND` to `ffmpeg`. In this case the installation of additional packages is not required to play ogg files.

client/audio/audio.cpp
jwrober: Ok cool, that makes sense. This leads to me to another question. At the time we fully pull SDL out would we also take out the "none" plugin and just keep Qt and quit asking for an option in the interface options?

jwrober:
Given the comments about using `ffmpeg` as the backend, I'm not going to worry about testing on Windows/MSYS2.

client/audio/audio.cpp
lmoureaux: I'd keep the option to disable music/sounds, but I guess "none" could be removed because these are separate.

lmoureaux:
I added a configure warning if OGG/Vorbis is not supported. Let's say if it complains on MSYS2.

lmoureaux:
Seems like music will work on Windows out of the box:
image

lmoureaux:
Testing Nix, Fedora, ... in my fork, don't merge just yet.

lmoureaux:
The configure check correctly warns about broken music on FreeBSD:
```
2025-04-20T01:11:55.8962063Z -- Performing Test HAVE_OGG_VORBIS
2025-04-20T01:11:57.9424591Z -- Performing Test HAVE_OGG_VORBIS - Failed
2025-04-20T01:11:57.9426923Z CMake Warning at cmake/FreecivDependencies.cmake:180 (message):
2025-04-20T01:11:57.9428015Z   Qt Multimedia does not support OGG Vorbis on your system.  Additional
2025-04-20T01:11:57.9430363Z   runtime libraries or environment variables (QT_MEDIA_BACKEND) may be needed
2025-04-20T01:11:57.9431006Z   to play in-game music.
2025-04-20T01:11:57.9431546Z Call Stack (most recent call first):
2025-04-20T01:11:57.9431935Z   CMakeLists.txt:76 (include)
```

jwrober:
What's up with clang-tidy-review?

lmoureaux:
It doesn't know about the new dependency because it runs with the config from `master`.

jwrober (approved):

lmoureaux:
All workflows passed in my fork except snap, but snap is also broken on master.

PR #2650, opened 2025-04-19
Fix including a deleted file when updating autorev

### LINKED ISSUE #2649: release.yaml deletes a file we want to keep
The `release.yaml` file deletes a problematic `.cmd` file used for windows server. However the `create-pull-request` action we uses picks up the deleted file and adds it to the commit. We don't want that.

jwrober:
Closes #2649

Backport request

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -40,7 +40,7 @@ jobs:
           branch: release/autorevision/${{ steps.branch.outputs.branch }}
           # Ensure we only add the changed cmake/AutoRevision.txt file in the commit
           add-paths: |
-            cmake/*.txt
+            cmake/AutoRevision.txt
           title: "Release update of AutoRevision.txt"
           body: >
             Automatic changes triggered by a new release.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -38,6 +38,9 @@ jobs:
           base: ${{ steps.branch.outputs.branch }}
           commit-message: "Release update of AutoRevision.txt"
           branch: release/autorevision/${{ steps.branch.outputs.branch }}
+          # Ensure we only add the changed cmake/AutoRevision.txt file in the commit
+          add-paths: |
+            cmake/*.txt
```
lmoureaux, line 43: Why not be even more selective and add `cmake/AutoRevision.txt`?

jwrober: Good question. The docs for the action - https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#add-specific-paths - says we need to "should follow git's pathspec syntax". Both examples have a wildcard, so I followed. In reading the link to the pathspec docs, I don't get a sense you can use the full filename. If you think I can use the full specific file we want to include and ignore everything else I'm good.

lmoureaux: Pathspecs are basically like shell expansion. SO sent me here: https://css-tricks.com/git-pathspecs-and-how-to-use-them/

The same syntax is e.g. in `.gitignore` and `.gitatribbutes`

lmoureaux (approved):

Issue #2647, opened 2025-04-19
x86 macOS binaries


lmoureaux:
**Describe the bug**
Following #2646 our macOS CI will build ARM binaries that are not compatible with old Macs.

**Expected behavior**
We need x86 to support older Apple hardware.

**Platform and version (please complete the following information):**
 - OS: macOS
 - Freeciv21 version: 3.1

jwrober:
I wonder if we have two actions running on Mac 26 (Tahoe) and let it ride for a couple years? Looks like Apple is going to phase out Intel support all over the map pretty soon. https://arstechnica.com/gadgets/2025/06/apple-details-the-end-of-intel-mac-support-and-a-phaseout-for-rosetta-2/

lmoureaux:
Apple will support Intel systems into 2028, that's in a while.

vcpkg doesn't support universal binaries, one has to build for Intel and ARM separately and then merge (there's a tool for that).

lmoureaux:
GitHub's x86 images are only for paying customers. Currently we cannot provide an x86 package at all.

lmoureaux:
Removing from v3.1: builds on the `stable` branch are x86.

jwrober:
<eyeroll> Apple has to make stuff hard don't they. So maybe the answer is we offer x86 and ARM binaries on Stable. ARM is the only way on Master branch. Qt6 requires mac 14+ (current/latest is 15)

lmoureaux:
Qt requires [Xcode 15+](https://doc.qt.io/qt-6/supported-platforms.html) and can target macOS 12+. Xcode 15 requires [a Mac running macOS Ventura 13.5 or later](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes). So it's possible ?

PR #2654, opened 2025-04-21
[stable] macOS CI: Use macos-13 for intel (x64-osx)

jwrober:
The github actions runner images for `macos-latest` (`macos-14+`) are based on the `arm64` architecture. Those images have an `intel` architecture option only for paying customers.  Solution is to use the `macos-13` image as it is still on `intel`.

See https://github.com/actions/runner-images/discussions/10810

Attempts to use varying `CMAKE_TRIPLET...` environment variables effectively creates a cross-compile environment and our build system isn't setup to do that it seems.

Part of #2647

Not closing the issue as this is effectively a work around and not a perm fix. It allows us to release 3.1, but we need to figure it out for further 3.2 series packages too.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -207,7 +207,7 @@ jobs:
 
   os_x:
     name: "macOS"
-    runs-on: macos-latest
+    runs-on: macos-13
     needs: clang-format
     env:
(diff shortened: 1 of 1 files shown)


lmoureaux:
Enshittification has begun.

lmoureaux (approved):

PR #2657, opened 2025-04-22
Emit warning if the Qt audio plugin does not support Ogg Vorbis

blabber:
Commit #2648 added a check that emits a configure warning if the backend of the Qt audio plugin does not support Ogg Vorbis. This commit adds an identical check to the Qt audio plugin itself and emits a warning when the plugin is loaded at runtime.

### DIFF
--- a/client/audio/audio_qt.cpp
+++ b/client/audio/audio_qt.cpp
@@ -13,6 +13,7 @@
 // Qt
 #include <QAudio>
 #include <QAudioOutput>
+#include <QMediaFormat>
 #include <QMediaPlayer>
 #include <QSoundEffect>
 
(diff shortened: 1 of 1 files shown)


lmoureaux (approved):

jwrober:
@lmoureaux Do we want to i18n the warning message?

lmoureaux:
Every user-visible string should be. Depends how we consider warnings that end up in the logs.

blabber:
I have oriented myself to the `qCritical`, which is used a few lines above the warning and is also not localized. It seemed reasonable to me that issues that are not displayed in the Gui are not localized. But I am willing to follow your guidance :)

jwrober:
I think its ok. I re-read https://longturn.readthedocs.io/en/latest/Coding/internationalization.html#not-translated.

PR #2679, opened 2025-06-06
Refactor: Common includes - major pass 2

jwrober:

Part of #2464

Sorry for this being so large, but with the major overhaul of `fc_types.h` I wanted to ensure I touched every file.  Still need to do `common/scriptcore`, which will also change the generated headers for that section.


lmoureaux (approved):
I hope we can simplify all these enum/struct definitions over time. It's really too much stuff hard-coded into the engine.

PR #2688, opened 2025-07-20
Support compiling without sphinx_lua_ls

lmoureaux:
sphinx_lua_ls isn't packaged for my system and I want to do the main build outside a virtualenv if possible.

Make sphinx_lua_ls optional. We get warnings when it's not there, which is fine.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,6 +27,12 @@ try:
 except ImportError:
     has_updated_by_git = False
 
+try:
+    import sphinx_lua_ls
+    has_sphinx_lua_ls = True
+except ImportError:
+    has_sphinx_lua_ls = False
+
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


XHawk87 (approved):

jwrober (approved):

XHawk87:
> sphinx_lua_ls isn't packaged for my system and I want to do the main build outside a virtualenv if possible.
> 
> Make sphinx_lua_ls optional. We get warnings when it's not there, which is fine.

Which system is this? Should we create an issue in https://github.com/taminomara/sphinx-lua-ls/issues to see if it can get packaged there?

Btw, I installed this for my system using pyenv, as recommended in the install instructions: https://taminomara.github.io/sphinx-lua-ls/#installation . Does this get around the issue for you?

lmoureaux:
There seems to be no package in the Debian/Ubuntu repos. I avoid installing things outside of apt because I forget that they are around and they accumulate bitrot. We should create a venv from within cmake or something like that...

PR #2693, opened 2025-07-21
Use markdown to render help texts in the help browser.

### LINKED ISSUE #4: Help dialog should render tables in html
Currently using ascii art

blabber:
This is a work in progress to render help texts in the help browser as markdown. Code generating help texts has been changed accordingly (this mainly concerns the generation of tables).

Although the help renders correctly for the most part, the contents of `data/helpdata.txt` needs to be reviewed and adapted to markdown. This is not in scope for this PR though and will be handled in seperate issues/PRs.

Closes #4

### DIFF
--- a/common/helpdata.cpp
+++ b/common/helpdata.cpp
@@ -136,22 +136,15 @@ static bool insert_veteran_help(char *outbuf, size_t outlen,
       CATLSTR(outbuf, outlen, "\n\n");
     }
     // TODO: Report raise_chance and work_raise_chance
-    CATLSTR(
-        outbuf, outlen,
-        /* TRANS: Header for fixed-width veteran level table.
-         * TRANS: Translators cannot change column widths :(
-         * TRANS: "Level name" left-justified, other two right-justified */
-        _("Veteran level      Power factor   Move bonus\n"));
+    cat_snprintf(outbuf, outlen, "| %s | %s | %s |\n", _("Veteran level"),
+                 _("Power factor"), _("Move bonus"));
     CATLSTR(outbuf, outlen,
             // TRANS: Part of header for veteran level table.
-            _("--------------------------------------------"));
+            _("| :-- | --: | --: |"));
     for (i = 0; i < veteran->levels; i++) {
       const struct veteran_level *level = &veteran->definitions[i];
       const char *name = name_translation_get(&level->name);
-      /* Use get_internal_string_length() for correct alignment with
-       * multibyte character encodings */
-      cat_snprintf(outbuf, outlen, "\n%s%*s %4d%% %12s", name,
-                   MAX(0, 25 - (int) get_internal_string_length(name)), "",
+      cat_snprintf(outbuf, outlen, "\n| %s | %4d%% | %s |", name,
                    level->power_fact,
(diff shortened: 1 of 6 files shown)


jwrober:
I'd venture that less than 1% of fc/fc21 users even know about the manual. I'd vote to delete it, especially with our pyciv work that allows for much more rich manuals in the rst space.

lmoureaux:
I don't think freeciv21-manual has many users, and we have the pyciv-based docs if someone wants to work on it... Still, we could convert markdown to HTML with QTextDocument.
So I'm not opposed to changing the output format or removing the tool completely, but maybe explore `QTextDocument` first in case it's an easy change.

blabber:
I taught `freeciv21-manual` to handle the markdown fragments generated as help texts. I removed the wiki syntax option, as the markdown handling only plays well with HTML (and we don't use mediawiki anyways).

One remaining issue I don't have a solution for are multiple warnings being emitted because `QFont::setPixelSize: Pixel size <= 0 (-1)`. The warning is harmless but irritating.

lmoureaux:
> One remaining issue, I don't have a solution for, are multiple warnings being emitted because QFont::setPixelSize: Pixel size <= 0 (-1). The warning is harmless but irritating.

Might need a `QGuiApplication` instead of a `QCoreApplication`? (And linking to QtGui.) It's the one managing fonts I think.

blabber:
> Might need a `QGuiApplication` instead of a `QCoreApplication`? (And linking to QtGui.) It's the one managing fonts I think.

Worked like a charm. Thanks for the hint.

blabber:
I rejected the idea of correcting the indentation of the bullet points.

1. The indentation does not affect how the lists are displayed in the Help Browser or the generated HTML manual.
2. The indentation and bullet points are part of the translated text. Leaving them as they are ensures that no translations will be broken.

common/helpdata.cpp
```diff
@@ -3804,28 +3780,20 @@ void helptext_extra(char *buf, size_t bufsz, struct player *pplayer,
...
             buf, bufsz,
             _("\nTime to build and output bonus depends on terrain:\n\n"));
-        CATLSTR(buf, bufsz,
-                /* TRANS: Header for fixed-width road properties table.
-                 * TRANS: Translators cannot change column widths :( */
-                _("Terrain       Time     Bonus F/P/T\n"
-                  "----------------------------------\n"));
+        cat_snprintf(buf, bufsz, "| %s | %s | %s |\n", _("Terrain"),
+                     _("Time"), _("Bonus F/P/T"));
+        CATLSTR(buf, bufsz, "| :-- | --: | :-- |\n");
```
lmoureaux, line 3785: Suggest aligning the last column to the right. Appears in Terrain > Terrain Alterations > Road

lmoureaux (commented):
We may want to reformat the following pages:
* License
* civ2civ3 README
* Controls

Not necessarily in this PR.

lmoureaux:
We should also change the default font size to Libertinus Sans, maybe size 14.

blabber:
It's the difference between using `-` and `*` as bullet point.

jwrober:
Ah ok. So this goes back to standardizing the formatting of some files.

lmoureaux (approved):
Let's merge and create issues for remaining bugs

Issue #2612, opened 2025-04-06
vcpkg does not have kf6archive


jwrober:
Opened a new port request

https://github.com/microsoft/vcpkg/issues/44852

jwrober:
Looks like there is another issue for all of the KDE Frameworks

https://github.com/microsoft/vcpkg/issues/44808

jwrober:
We are in serious trouble now LoL - https://github.com/microsoft/vcpkg/pull/46399

PR #2696, opened 2025-07-21
vcpkg updates for kf6archive

jwrober:
Part of #2612

We need to have MSFT approve my [PR](https://github.com/microsoft/vcpkg/pull/46399) for this to fully work. Once that happens, then we can enable the CI for Windows Visual Studio and macOS.

The main piece of this PR is to remove the baseline commit id. We can update to a new baseline when kf6archive is stable in vcpkg.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,11 @@
 #   server and other components
 cmake_minimum_required(VERSION 3.21...3.22 FATAL_ERROR)
 
+# Suppress policy warnings due to normalized install directories
+if(POLICY CMP0177)
+  cmake_policy(SET CMP0177 OLD)
+endif()
+
 # Set vcpkg if exists. Used by MacOS and Visual Studio
 if(DEFINED ENV{VCPKG_ROOT} AND FREECIV_USE_VCPKG)
   message(STATUS "Microsoft VCPKG enabled, setting toolset specific settings.")
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,7 +1,6 @@
 {
   "name": "freeciv21",
-  "version-string": "3.0.0",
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -4,6 +4,11 @@
 #   server and other components
 cmake_minimum_required(VERSION 3.21...3.22 FATAL_ERROR)
 
+# Suppress policy warnings due to normalized install directories
+if(POLICY CMP0177)
+  cmake_policy(SET CMP0177 OLD)
```
lmoureaux, line 9: Note that this is deprecated, so we should try and update our usage instead.

> The OLD behavior of a policy is [deprecated by definition](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#manual:cmake-policies(7)) and may be removed in a future version of CMake.

jwrober: We will need to figure out a different way of how we normalize directories for the windows build. That is where the message comes from on the policy. I set the value to reduce a ton of messages at configure time.

lmoureaux: I didn't fully understand the docs, do you have an example?

jwrober: These lines cause the policy to complain. They only come up when compiling on MSYS2 or in Visual Studio. I honestly don't fully understand what the problem is, but cmake decided to change "normalized directories" behavior.

https://github.com/longturn/freeciv21/blob/master/CMakeLists.txt#L64-L68

here is the docs on the policy: https://cmake.org/cmake/help/latest/policy/CMP0177.html

I'll open an issue to see if we can correct how we are doing things.

lmoureaux (approved):
Bump the version string and we can merge!

PR #2697, opened 2025-07-22
fix locales on windows

### LINKED ISSUE #2687: Locales on Windows do not work
Freeciv21 on Windows only ever uses English, despite translations being installed.

Original issue text:

----
Hello,
short question: How can i set the language of the game? I need german and the game starts every time in english and i found no settings.
thx

jwrober:
Closes #2687


jwrober:
Suggest a backport too

.github/workflows/build.yaml
```diff
@@ -169,6 +169,8 @@ jobs:
           cmake . -B build -G Ninja \
            -DCMAKE_INSTALL_PREFIX=$PWD/build/install \
            -DCMAKE_BUILD_TYPE=Release
+          sed -i 's/LOCALEDIR/LOCALEDIRBAD/' build/utility/fc_config.h
+          echo '#define LOCALEDIR "./locale"' >> build/utility/fc_config.h
```
lmoureaux, line 173: This means that locally-built installers will still be broken. I'd prefer to do this in CMake, or directly in the code:
```cpp
#ifdef Q_OS_WIN
  return "./locale";
#else
  return LOCALEDIR;
#endif
```

lmoureaux: If I run cpack locally to create the msi nothing changes, right? Or am I missing something?

lmoureaux:
Superseded by #2698 / #2699

PR #2698, opened 2025-07-22
Use QStandardPaths to find the locale folder on Win and Mac

lmoureaux:
Installed Windows packages could not find locale files because a build-time path was hardcoded (#2687). Mac builds never included translations but would suffer from a similar issue.

Use QStandardPaths::AppDataLocation to find the folder on these OSes. Keep the same logic on Linux.


lmoureaux:
Bingo!

image

image

Issue #2720, opened 2025-07-28
Game freezes on Windows when leaving and re-entering from the main menu


lmoureaux:
**Describe the bug**
A Windows crash was reported that appears when leaving a game and choosing another option from the main menu.

**To Reproduce**
1. Click New Game
2. Cancel
3. Click New Game again
4. Crash

**Expected behavior**
No crash

**Screenshots**

<img width="812" height="443" alt="Image" src="https://github.com/user-attachments/assets/39f07ea1-83ca-4a6e-9fba-4bf2b8524b51" />

**Platform and version (please complete the following information):**
 - OS: Unspecified Windows version
 - Freeciv21 version: 3.0.5

**Additional context**

Report on Discord: https://discord.com/channels/378908274113904641/379424175623766027/1399466212273946776

lmoureaux:
3.1-rc3 isn't crashing but freezes for quite some time.

PR #2723, opened 2025-07-29
Fix deleting the server when disconnecting and reconnecting quickly

lmoureaux:
We were seeing a bug where the client would fail to create a new server quickly after disconnecting from another (https://github.com/longturn/freeciv21/issues/2720). This lead to it trying to connect for almost 50 seconds, freezing everything since delays are implemented with sleep().

This commit ensures that we always end up in a state where the server is gone after client_kill_server(). This may block the UI for a short time when disconnecting while the server saves the game. We limit it to 10s (hard coded, which is bad, but only very large games will take this long to save).

We will need better APIs in the future.
Suggest to backport to stable.

Closes https://github.com/longturn/freeciv21/issues/2720.

### DIFF
--- a/client/connectdlg_common.cpp
+++ b/client/connectdlg_common.cpp
@@ -164,6 +164,15 @@ void client_kill_server(bool force)
        * server.  In this case we don't want to kill it. */
       send_chat("/quit");
       server_quitting = true;
+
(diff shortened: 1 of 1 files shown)


jwrober (approved):
Have a question, but not blocking merge. I also think a backport is warranted for this bug.

PR #2743, opened 2025-08-03
[stable] Fix macro to select Windows version of anim_delay

lmoureaux:
Fixes: 2408738725242bd2be7b37761dc5f7e9eafa1cbc.


lmoureaux:
I've played a few hours with this patch and had many battles without a single crash.

PR #2763, opened 2025-08-11
update macos build and docs

jwrober:
No related issue. Small update to the macOS install steps and docs. Main thing it turns out is we can go all in on ARM support and allow Intel users to use Rosetta2 from Apple ("the Intel-to-Arm app translation technology created to ease the transition between the Intel and Apple Silicon eras")


lmoureaux (commented):
I think the macOS build should run on PRs? At least on this one.

.github/workflows/build.yaml
```diff
@@ -286,14 +287,14 @@ jobs:
...
             --window-pos 200 120 \
-            --window-size 800 400 \
+            --window-size 600 400 \
             --icon-size 128 \
             --icon "Freeciv21.app" 200 190 \
             --hide-extension "Freeciv21.app" \
-            --app-drop-link 600 185 \
-            "Freeciv21-${{steps.split.outputs.fragment}}.dmg" \
+            --app-drop-link 400 185 \
+            "Freeciv21-${{steps.split.outputs.fragment}}x64-osx.dmg" \
```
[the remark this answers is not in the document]
jwrober: and that's the wrong triplet forgot we are running mac-latest in master

lmoureaux (changes requested):
There are two downsides with ARM code:
* It code needs to be signed to run. Let's keep distributing x86 until we figure this out.
* It will not work on x86 hardware (while x86 binaries do work on ARM thanks to Rosetta)

I think we should solve the first issue before distributing ARM binaries, and keep distributing x86 (or possibly universal) binaries as long as possible.

jwrober:
Then I guess we have to figure out how to compile on macos-15, but "target" macos-13

jwrober:
WRT this PR... I'd prefer we commit with some caveats. 

Rationale:

1) Figuring out how to compile x86 on Arm is not trivial. Cross-compiling is always a challenge. Universal libraries is also not trivial.
2) I want to get cmake to do the app bundle and dmg generation natively. I have a PR ready to start this process based on this one. This can be done in "Arm" mode in CI. My x64_x86 VM can be used for testing in the short term.
3) This PR has some good doco updates I'd like to keep.
4) I can comment out the lines to upload binaries to releases or regular builds until we get the signing or cross-compile thing figured out. macOS isn't ... well ... supported in 3.2-dev* right now. You want to run it, you do roll your own. Nothing new here.

Long story short - Apple is moving forward and we need to figure out how to follow whether we like it or not. This is not something that will happen in this PR.

.github/workflows/build.yaml
```diff
@@ -286,14 +289,14 @@ jobs:
...
-            --app-drop-link 600 185 \
-            "Freeciv21-${{steps.split.outputs.fragment}}.dmg" \
+            --app-drop-link 400 185 \
+            "Freeciv21-${{steps.split.outputs.fragment}}-arm64-osx.dmg" \
             "Freeciv21.app"
-          shasum -a 256 Freeciv21-${{steps.split.outputs.fragment}}.dmg > Freeciv21-${{steps.split.outputs.fragment}}.dmg.sha256
+          shasum -a 256 Freeciv21-${{steps.split.outputs.fragment}}-arm64-osx.dmg > Freeciv21-${{steps.split.outputs.fragment}}-arm64-osx.dmg.sha256
       - name: Upload package
         uses: softprops/action-gh-release@v2
         if: startsWith(github.ref, 'refs/tags/')
```
lmoureaux, line 302: Will want to restrict this to one of the jobs.

jwrober: Can correct at a later date

lmoureaux (approved):

PR #2792, opened 2025-08-26
Fixed techs with Never as req causing segfault

### LINKED ISSUE #2790: Lua give unresearchable tech segfault
**Describe the bug**
When setting up a tech that cannot be researched e.g. the root of a separate tech tree for a scenario, the server segfaults when using `Player:give_tech` to grant that tech to a player in Lua.

**To Reproduce**
1. Add a tech that cannot be researched. E.g. 
```
[advance_red_flag]
name        = _("Red Flag-making")
req1        = "Never"
req2        = "Never"
flags       = ""
graphic     = "None"
graphic_alt = "-"
```
2. Start the game.
3. Grant the tech to a player. E.g. `find.player(0):give_tech(find.tech_type("Red Flag-making", 0, false, "team_start_tech"))`

**Expected behavior**
There should be no error, and the player should get the tech. This is necessary to grant players access to new research lines via script.

 - OS: Linux
 - Freeciv21 version: 3.2-dev.2 (5a32bad)
 - Ruleset/Longturn game (if applicable): Capture the Flag ruleset

**Additional context**
[Discord thread](https://discord.com/channels/378908274113904641/1409023313514205236)

XHawk87:
Closes #2790

### DIFF
--- a/server/ruleset.cpp
+++ b/server/ruleset.cpp
@@ -2286,17 +2286,6 @@ static bool load_ruleset_units(struct section_file *file,
     // Some more consistency checking:
     unit_type_iterate(u)
     {
-      if (!valid_advance(u->require_advance)) {
-        qCCritical(
-            ruleset_category,
-            "\"%s\" unit_type \"%s\": depends on removed tech \"%s\".",
-            filename, utype_rule_name(u),
-            advance_rule_name(u->require_advance));
-        u->require_advance = A_NEVER;
-        ok = false;
-        break;
-      }
-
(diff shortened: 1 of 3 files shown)


server/ruleset.cpp
```diff
@@ -2286,17 +2286,6 @@ static bool load_ruleset_units(struct section_file *file,
     // Some more consistency checking:
     unit_type_iterate(u)
     {
-      if (!valid_advance(u->require_advance)) {
-        qCCritical(
```
jwrober, line 2290: Why get rid of this `if()` block?

jwrober (approved):
Code looks good. I just have one question about some removed code.

Issue #2793, opened 2025-08-26
Unhappiness from foreign citizens even while peace with the original nation


wieder-fi:
It's currently a problem if people start doing city gifting and if they want to revenge to someone or just to boost some player. At the moment it's only possible to make the foreign citizens unhappy if the current owner is at war with the original nation of those citizens. However a friendly nation allowing some nation to grab cities may agree to instant peace once the gifting is done.

Adding a setting that would make it possible to define how much unhappiness is created if there is peace between the nations. 

Alternatives include more strict empire sizes but those have been very hard to make work well enough and they also address many other issues.

XHawk87:
Agreed. This should be controlled by `DiplRel` rather than being hardcoded to enemies only.

XHawk87:
Should this be closed?

PR #2794, opened 2025-08-26
Allow using DiplRel in Enemy_Citizen_Unhappy_Pct

lmoureaux:
Enemy_Citizen_Unhappy_Pct controls the fraction of "enemy" citizens that cause unhappiness. Only players you are at war with are enemies. There are exploits with city gifting: one can be at war for a short time and give cities away, then come back to peace to avoid unhappiness. This motivates finer-grained control.

This patch allows the effect to use a DiplRel requirement to specify when citizens should be unhappy.

A side effect is that the number of unhappy citizens is now calculated globally based on all foreign citizens and not per nationality. This may lead to more unhappiness in cities with many different nationalities.

See #2793.

**To do:**
- [x] Migrate old rulesets by adding a DiplRel, war requirement to all Enemy_Citizen_Unhappy_Pct effects
- [x] Add a capability to disable this behaviour
- [x] Update shipped rulesets
- [x] Test

### Test plan

Change the following effect in the `experimental` ruleset:
```ini
[effect_base_enemy_unhappiness]
type    = "Per_Citizen_Unhappy_Pct"
value   = 100
reqs =
    { "type", "name", "range"
      "DiplRel", "War", "Local"
    }
```
Remove the `DiplRel` requirement to make it easier to see the effect and play with `value`.

Start a game with the modified ruleset, found a city, check how many citizens become unhappy.

### DIFF
--- a/docs/Modding/Rulesets/effects.rst
+++ b/docs/Modding/Rulesets/effects.rst
@@ -410,7 +410,7 @@ Make_Content_Mil_Per
 Make_Happy
     Make AMOUNT citizens happy.
 
-.. _effect-enemy-citizen-unhappy-pct:
+.. _effect-per-citizen-unhappy-pct:
 
 Per_Citizen_Unhappy_Pct
     There will be one extra unhappy citizen for each value/100 matching citizens in the city. This


XHawk87 (commented):
Looks good so far. Would changing the name be difficult? Possibly keeping the old one for legacy reason and adding a new one? Perhaps `Per_Citizen_Unhappy_Pct` and then you can have unhappiness caused by own citizens as well in addition to `City_Unhappy_Size`. Since these are applied in a different order and work a slightly different way, it could be useful for things that should or should not be affected by `Make_Content` for example.

jwrober (approved):
One small suggested change. Rest looks good to me.

PR #2800, opened 2025-08-27
Units are killed immediately from bombardment if they drop to 0HP

### LINKED ISSUE #2382: Units bombarded down to 0hp should die
**Describe the bug**
Extensively bombarded units inside a city. To the point they should have *0hp* left - especially since they have 0 defence. But the ground troops who then ended them reported winning without taking out any hp. see screenshot below.

**To Reproduce**
1. Let an idle city produce 10+ raiders so their defence becomes 0 (idk the actual endpoint that makes i t0)
2. bombard that city with 10 archers, unload them all.
3. attack the city with your own raiders after.

**Expected behavior**
defenders should have 1hp left?
edit: *they should die when bombarded down to 0*

**Screenshots**
image

 - OS: Linux
 - Freeciv21 version: 3.1-beta.2
 - Ruleset/Longturn game (if applicable): Sim06

**Additional context**
noticed during the test game

*Edit:*

* `Bombard_Limit_Pct` has a value of 0 in the ruleset (it is not set)
* See comment in #867

XHawk87:
Closes #2382

Image

Image

### DIFF
--- a/server/unithand.cpp
+++ b/server/unithand.cpp
@@ -130,6 +130,10 @@ static bool unit_do_help_build(struct player *pplayer, struct unit *punit,
                                const struct action *paction);
 static bool unit_bombard(struct unit *punit, struct tile *ptile,
                          const struct action *paction);
+static void notify_defender_lost_in_bombardment(struct unit *pdefender,
+                                                struct unit *pbombarder);
+static void notify_unit_killed_in_bombardment(struct unit *pbombarder,
+                                              struct unit *pdefender);
 static bool unit_nuke(struct player *pplayer, struct unit *punit,
                       struct tile *def_tile, const struct action *paction);
 static bool unit_do_destroy_city(struct player *act_player,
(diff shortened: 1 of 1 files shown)


server/unithand.cpp
```diff
@@ -3794,6 +3811,67 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
...
+  int defender_fp;
+  get_modified_firepower(pbombarder, pdefender, &attacker_fp, &defender_fp);
+
+  notify_player(
+      unit_owner(pbombarder), unit_tile(pbombarder), E_UNIT_WIN_ATT,
+      ftc_server,
+      /* TRANS: "Your green Catapult [id:100 A:5.0 FP:1 BR:10]
+       * utterly destroyed a Greek green Legion [id:200 ...D:4.0 HP:10]
+       * in a bombardment." */
+      _("Your %s %s [id:%d A:%.1f FP:%d BR:%d] utterly destroyed a"
```
lmoureaux, line 3865: Would drop "utterly"

server/unithand.cpp
```diff
@@ -3794,6 +3811,67 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
...
+        " %s %s %s [id:%d D:%.1f HP:%d] in a bombardment."),
```
lmoureaux, line 3866: Is "a" appropriate? We know which

server/unithand.cpp
```diff
@@ -3794,6 +3811,67 @@ static bool unit_bombard(struct unit *punit, struct tile *ptile,
...
+  int defender_fp;
+  get_modified_firepower(pbombarder, pdefender, &attacker_fp, &defender_fp);
+
+  notify_player(
+      unit_owner(pdefender), unit_tile(pdefender), E_UNIT_LOST_DEF,
+      ftc_server,
+      /* TRANS: "Your green Warriors [id:100 D:1.0 HP:10]
+       * helplessly perished in a bombardment by the Greek green Catapult
+       * [id:200 ...A:4.0 FP:1 BR:10]." */
+      _("Your %s %s [id:%d D:%.1f HP:%d] helplessly perished in a"
```
lmoureaux, line 3834: Not convinced by "helplessly" in this case...

lmoureaux (commented):
Code looks good. Not sure about reusing the messages for a combat where the attacker doesn't lose any hp, I don't really find the situation comparable

XHawk87:
> Not sure about reusing the messages for a combat where the attacker doesn't lose any hp, I don't really find the situation comparable

It seems comparable to me. This is a situation where the attacker destroyed the defender without taking any damage. They died without being able to fight back. If we add artillery duels, then a more neutral message where both sides take damage seems appropriate.

lmoureaux:
I do prefer helpless over utterly.

The difference I see is actually on another aspect. "Helpless" and "utterly" are there purely to add flavour when you win a fight in an exceptional way (since even a tank usually loses one round vs a phalanx). Using them in a case where this is the normal way of killing opponents weakens the "exceptional" aspect of these messages.

PR #2815, opened 2025-08-31
OBS Debian Integration

### LINKED ISSUE #2813: OBS: Enable Debian build integration
Open Build Service integration is brand new. Right now its only supporting RPM based distros. We need to support DEB based distros such as Debian, Ubuntu, etc. Eventually get rid of our own DEB package from CPack.

Add Debian build files to `.obs`

- `debian.changes` -- see if we can use the `%autochangelog` macro
- `debian.control`
- `debian.rules`
- `debian.series`
- `freciv21.dsc`

jwrober:
Closes #2813

### DIFF
--- a/.obs/debian.changelog
+++ b/.obs/debian.changelog
@@ -1,0 +1,5 @@
+freeciv21 (3.2.0-1) unstable; urgency=low
+
+  * Unstable 3.2-dev.2
+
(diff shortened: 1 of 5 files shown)


lmoureaux (approved):
Seems to be running on the OBS side

jwrober:
I'm not totally sure if the webhook is really doing anything at this point, but it is running

PR #2820, opened 2025-09-02
Added homeless_gold_upkeep to server

### LINKED ISSUE #509: Unhomed (or all) units are paid (gold upkeep) from national budget, not individual city
Requested by Corbeau for the Sim ruleset.

XHawk87:
Closes #509

### DIFF
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -664,12 +664,9 @@ std::vector<unit_view_entry> get_units_view_data()
               client_has_player()
               && nullptr != can_upgrade_unittype(client_player(), unittype);
 
-          // Only units with a home city have upkeep
-          if (punit->homecity != 0) {
-            gold_cost += punit->upkeep[O_GOLD];
-            food_cost += punit->upkeep[O_FOOD];
-            shield_cost += punit->upkeep[O_SHIELD];
-          }
+          gold_cost += punit->upkeep[O_GOLD];
+          food_cost += punit->upkeep[O_FOOD];
+          shield_cost += punit->upkeep[O_SHIELD];
         }
       }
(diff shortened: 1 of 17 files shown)


lmoureaux (commented):
For client support you would minimally need to modify `player_get_expected_income` in `common/player.cpp`. Showing the units in the economy view would require changing `get_economy_report_units_data` in `client/repodlgs_common.cpp`. This is looking reasonably easy so I think we want it in this PR.

jwrober (approved):
Code looks good to me. I'm able to see gold upkeep for unhomed units in both units and econ views.

client/views/view_units.cpp
```diff
@@ -664,12 +664,9 @@ std::vector<unit_view_entry> get_units_view_data()
...
 
-          // Only units with a home city have upkeep
-          if (punit->homecity != 0) {
-            gold_cost += punit->upkeep[O_GOLD];
-            food_cost += punit->upkeep[O_FOOD];
-            shield_cost += punit->upkeep[O_SHIELD];
-          }
+          gold_cost += punit->upkeep[O_GOLD];
+          food_cost += punit->upkeep[O_FOOD];
+          shield_cost += punit->upkeep[O_SHIELD];
```
lmoureaux, line 669: Shouldn't this depend on `homeless_gold_upkeep`?

XHawk87: No need, since the `upkeep` doesn't get populated unless `homeless_gold_upkeep` is enabled. I'm guessing this if-block was either a mistake or premature optimisation. It's better if any upkeep automatically gets added to the view regardless of how the server arrived at the decision to apply it so that it's future-proof against further changes.

PR #2828, opened 2025-09-12
Allow Action reqs for combat bonus effects and introduced req_context

### LINKED ISSUE #2776: Combat bonus effects should allow Action requirements for different types of attacks
Currently, it's not possible to differentiate between the various Bombard, Attack, or Suicide Attack actions when applying `Attack_Bonus`, `Defend_Bonus`, and `Fortify_Defense_Bonus` effects. This would be useful to reduce odds to hit for when a unit is choosing to sit back and Bombard over moving in close for a direct Attack. It could also be used to adjust odds for the 3 different ranges of `Bombard 2/3`.

Supply the Action type when evaluating these effects. 

Expand the unit type bonuses system to allow full requirements rather than just the opposing unit flag.

**Additional context**
None yet.

XHawk87:
Closes #2776

### DIFF
--- a/common/effects.cpp
+++ b/common/effects.cpp
@@ -604,16 +604,8 @@ bool is_building_replaced(const struct city *pcity,
 }
 
 /**
-   Returns the effect bonus of a given type for any target.
-
-   target gives the type of the target
-   (player,city,building,tile) give the exact target
-   effect_type gives the effect type to be considered
-
-   Returns the effect sources of this type _currently active_.
-
-   The returned vector must be freed (building_vector_free) when the caller
-   is done with it.
+ * This function is deprecated. If you want to add new parameters, switch to
+ * using get_target_bonus or get_target_bonus_effects by req_context.
  */
 int get_target_bonus_effects(
     struct effect_list *plist, const struct player *target_player,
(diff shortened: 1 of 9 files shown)


common/effects.cpp
```diff
@@ -604,16 +604,8 @@ bool is_building_replaced(const struct city *pcity,
...
-   effect_type gives the effect type to be considered
-
-   Returns the effect sources of this type _currently active_.
-
-   The returned vector must be freed (building_vector_free) when the caller
-   is done with it.
+ * This function is deprecated. If you want to add new parameters, switch to
+ * using get_target_bonus or get_target_bonus_effects by req_context.
  */
 int get_target_bonus_effects(
```
lmoureaux, line 610: Let's remind ourselves to remove this and other deprecated functions. In the headers:
```cpp
[[deprecated]]
int get_target_bonus_effects(...)
```

common/requirements.h
```diff
@@ -32,6 +32,43 @@ class QString;
...
+  const struct player *player;
+  const struct specialist *specialist;
+  const struct tile *tile;
+  const struct unit *unit;
+  const struct unit_type *utype;
+  const enum vision_layer vision_layer = V_COUNT;
+};
+
+// Context functions.
+#define req_action(_rc_) (_rc_ ? _rc_->action : nullptr)
```
lmoureaux, line 60: Macros, not functions. A function would be:
```suggestion
auto req_action(const req_context *rc) { return rc ? rc->action : nullptr; }
```
…and I'd like it better.

lmoureaux (approved):

PR #2830, opened 2025-09-26
Bump CI to Ubuntu 25.10 (questing)

lmoureaux:
We were using 24.10, which has reached EOL and has been removed from the Azure mirror.


lmoureaux:
Will need to bump the required CI checks to 25.10.

PR #2848, opened 2026-02-28
Move more packet formatting code to dataio

lmoureaux:
This builds on #2845 to move basic packet serialization logic out of `generate_packets.py`. Enums and arrays no longer need explicitly generated code; they rely on template functions in `dataio` instead. For instance, writing a packet now looks like this regardless of the data types involved:
```cpp
  ...
  /* field 5 is folded into the header */
  /* field 6 is folded into the header */
  if (BV_ISSET(fields, 7)) {
    log_packet_detailed("  field 'name' has changed");
    dio_put(dout, packet->name);
  }
  if (BV_ISSET(fields, 8)) {
    log_packet_detailed("  field 'hp' has changed");
    dio_put(dout, packet->hp, std::uint8_t{});
  }
  if (BV_ISSET(fields, 9)) {
    log_packet_detailed("  field 'activity' has changed");
    dio_put(dout, packet->activity, std::uint8_t{});
  }
  ...
```
The generated code used to contain a lot of logic such as loops. This is now entirely gone (as far as the `dio_*` part is concerned).

### DIFF
--- a/client/attribute.cpp
+++ b/client/attribute.cpp
@@ -220,11 +220,8 @@ static enum attribute_serial unserialize_hash(attributeHash *hash,
       return A_SERIAL_FAIL;
     }
 
-    QByteArray pvalue(value_length, Qt::Initialization::Uninitialized);
-    dio_put<std::uint32_t>(pvalue, value_length);
-    if (!dio_get(din, pvalue.data() + 4, value_length)) {
-      qDebug("attribute.cpp unserialize_hash() "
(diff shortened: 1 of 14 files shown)


lmoureaux:
Agreed on the rebase. I fixed the unreachable code from codacy, the other checks trigger whenever one uses `memcpy()` (also when it is used correctly).

lmoureaux:
Rebased

jwrober (approved):

PR #2879, opened 2026-04-14
macOS Qt6 linked Dynamically

### LINKED ISSUE #2766: build macOS package with dynamic libraries vs static
The installed size of Freeciv21 on macOS is 500MB! The main reason is the executables are all statically linked to QT6 - 7 times (for each program).

We dynamically link on all other OS's including Windows. We need to do the same w/ macOS

Use ``--dynamic` vcpkg triplets.

https://github.com/microsoft/vcpkg/discussions/31635
https://stackoverflow.com/questions/71588349/qt5-build-as-shared-library-on-macos-using-cmake-and-vcpkg/71860329#71860329

**Additional context**
None

jwrober:
With a custom triplet, we can compile Qt in dynamic mode and then link the executable to the proper dylib's with a similar approach that we use on Windows.

Closes #2766


cmake/FreecivInstall.cmake
```diff
@@ -356,6 +356,55 @@ if(APPLE)
...
+      COMMAND ${CMAKE_BINARY_DIR}/vcpkg_installed/x64-osx-custom/tools/Qt6/bin/macdeployqt6  
+    ${CMAKE_BINARY_DIR}/Freeciv21.app -verbose=1 
+    -executable=${CMAKE_BINARY_DIR}/Freeciv21.app/Contents/MacOS/freeciv21-ruleup
+    )
+    # freeciv21-server
+    execute_process(
+      COMMAND ${CMAKE_BINARY_DIR}/vcpkg_installed/x64-osx-custom/tools/Qt6/bin/macdeployqt6  
+    ${CMAKE_BINARY_DIR}/Freeciv21.app -verbose=1 
+    -executable=${CMAKE_BINARY_DIR}/Freeciv21.app/Contents/MacOS/freeciv21-server
+    )
```
lmoureaux: This could be a for loop:
```cmake
foreach(exec freeciv21-client freeciv21-manual freeciv21-modpack...
  execute_process(
      COMMAND "${CMAKE_BINARY_DIR}/vcpkg_installed/x64-osx-custom/tools/Qt6/bin/macdeployqt6"  
    "${CMAKE_BINARY_DIR}/Freeciv21.app" -verbose=1 
    -executable="${CMAKE_BINARY_DIR}/Freeciv21.app/Contents/MacOS/${exec}"
  )
endforeach()
```

CMakeLists.txt
```diff
@@ -8,6 +8,12 @@ if(DEFINED ENV{VCPKG_ROOT} AND FREECIV_USE_VCPKG)
   message(STATUS "Microsoft VCPKG enabled, setting toolset specific settings.")
   set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
       CACHE STRING "")
+  if(APPLE)
```
lmoureaux: Not a big fan of forcing it on all MacOS users. Could we use a preset instead?

cmake/FreecivBackward.cmake
```diff
@@ -9,8 +9,8 @@ include(CheckSymbolExists)
 #   - native API on MSVC (MSYS and MINGW aren't ABI-compatible with it)
 #   - libunwind if available
 #   - backtrace() if everything else fails.
-if (MSVC OR MSYS)
-  message(STATUS "Using the Windows native API for stack unwinding. "
+if (MSVC OR MSYS OR APPLE)
+  message(STATUS "Using the native API for stack unwinding. "
```
lmoureaux: There is no native API on MacOS, `libunwind` gets use if nothing else is specified.

cmake/FreecivBackward.cmake
```diff
@@ -23,7 +23,7 @@ if (NOT CAN_UNWIND_STACK)
     pkg_search_module(unwind IMPORTED_TARGET libunwind)
     if (unwind_FOUND)
       message(STATUS "Using libunwind for stack unwinding. "
-                     "This is the preferred option.")
+                     "This is the secondary option.")
```
lmoureaux: It is definitely preferred on Linux and MacOS.

cmake/FreecivInstall.cmake
jwrober: good idea

CMakeLists.txt
jwrober: ok sounds good

cmake/FreecivBackward.cmake
jwrober: There is built in support in the XCode command line tools - hence "native" so I think this is fine.

cmake/FreecivBackward.cmake
jwrober: I'll update the messages

lmoureaux:
Are the indentation changes done on purpose? They'll make it harder to explore the commit history

jwrober:
I installed a YAML linter extension to VS Code (that I'm using on my hackintosh) that seems to have an auto-save/fix setting enabled to properly format on save. That is a lot of lines changed at once. Ideas on how to help commit history? I'll bet many of our other YAML files would get a similar change if I edit them.

jwrober:
After thinking on this some more, I'm going to redo this PR so the build.yaml file doesn't get clobbered.

PR #2890, opened 2026-05-11
Support Lua 5.5

### LINKED ISSUE #2889: Msys2 build failing with incorrect Lua
**Describe the bug**
`build.yaml` is failing during the Msys2 step. Msys2 is now offering Lua 5.5.x. Sol is curently only working for Lua 5.4.x and below.

**To Reproduce**
See any recent build run log in Actions.

**Expected behavior**
Msys2 builds work since this is how we create our windows package.

**Screenshots**
N/A

 - OS: Windows Msys2
 - Freeciv21 version: N/A

**Additional context**
None

jwrober:
Closes #2889

### DIFF
--- a/common/scriptcore/luascript.cpp
+++ b/common/scriptcore/luascript.cpp
@@ -86,7 +86,7 @@ static const char *luascript_unsafe_symbols_permissive[] = {
   Lua libraries to load (all default libraries, excluding operating system
   and library loading modules). See linit.c in Lua 5.1 for the default list.
  */
-#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM >= 504
+#if LUA_VERSION_NUM >= 503 && LUA_VERSION_NUM <= 505
 static luaL_Reg luascript_lualibs_secure[] = {
     // Using default libraries excluding: package, io, os, and bit32
     {"_G", luaopen_base},
--- a/common/scriptcore/luascript.cpp
+++ b/common/scriptcore/luascript.cpp
@@ -86,7 +86,7 @@ static const char *luascript_unsafe_symbols_permissive[] = {
   Lua libraries to load (all default libraries, excluding operating system
   and library loading modules). See linit.c in Lua 5.1 for the default list.
  */
-#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM == 504
+#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM >= 504
 static luaL_Reg luascript_lualibs_secure[] = {
     // Using default libraries excluding: package, io, os, and bit32
     {"_G", luaopen_base},
--- a/dependencies/sol2/sol/sol.hpp
+++ b/dependencies/sol2/sol/sol.hpp
@@ -3537,9 +3537,9 @@ COMPAT53_API void luaL_requiref(lua_State *L, const char *modname,
 /* other Lua versions */
 #if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 505
 
-#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"
+#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, 5.4 or 5.5)"
 
-#endif /* other Lua versions except 5.1, 5.2, 5.3, and 5.4 */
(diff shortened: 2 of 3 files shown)


common/scriptcore/luascript.cpp
```diff
@@ -86,7 +86,7 @@ static const char *luascript_unsafe_symbols_permissive[] = {
   Lua libraries to load (all default libraries, excluding operating system
   and library loading modules). See linit.c in Lua 5.1 for the default list.
  */
-#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM == 504
+#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM >= 504
```
XHawk87: This might have been done this way on purpose to be careful about not automatically including libraries that get deprecated or removed. I.e. forcing the dev to check these each time you update, rather than risk introducing bugs that only show up when compiled with specific versions of Lua while running scripts.

Perhaps `LUA_VERSION_NUM >= 503 && LUA_VERSION_NUM <= 505`?

dependencies/sol2/sol/sol.hpp
```diff
@@ -3535,7 +3535,7 @@ COMPAT53_API void luaL_requiref(lua_State *L, const char *modname,
 #endif /* Lua 5.2 only */
 
 /* other Lua versions */
-#if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 504
+#if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 505
 
 #  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"
```
XHawk87: Don't forget to update the error message and comment below too.

dependencies/sol2/sol/sol.hpp
XHawk87, line 4431: This label no longer matches the check above. Are you sure we're meant to be doing this? What is this Kepler Project? It might be worth looking it up to see how this is supposed to be done. There might be a 5.5-specific compatibility shim to install.

XHawk87:
There is a fair amount of Lua in the game by default, so running a test game should be good. Though it might also be worth testing with an [authentication database](https://longturn.readthedocs.io/en/latest/Manuals/Server/fcdb.html) since there are also `/fcdb` commands that might work differently to `/lua cmd`

XHawk87 (changes requested):

dependencies/sol2/sol/sol.hpp
jwrober: I don't anything about Kepler. The patch came from something Louis94 found. I think leaving these alone is fine.

common/scriptcore/luascript.cpp
```diff
@@ -86,7 +86,7 @@ static const char *luascript_unsafe_symbols_permissive[] = {
   Lua libraries to load (all default libraries, excluding operating system
   and library loading modules). See linit.c in Lua 5.1 for the default list.
  */
-#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM >= 504
+#if LUA_VERSION_NUM == 503 || LUA_VERSION_NUM <= 505
```
XHawk87: Lack of coffee error. Please reboot programmer.

lmoureaux (approved):

lmoureaux:
We need @XHawk87 to confirm that the changes he requested have been implemented

common/scriptcore/luascript.cpp
XHawk87: I still don't think this is a good idea. It's something you should check each time a new Lua version comes out to see if any libraries have been deprecated or removed rather than assuming everything is fine until there's a crash.

XHawk87: Even if you are going to go this route, the logic should be simplified `LUA_VERSION_NUM >= 503`

lmoureaux: This doesn't seem resolved

lmoureaux: And `LUA_VERSION_NUM == 503 || LUA_VERSION_NUM >= 504` is equivalent to `LUA_VERSION_NUM >= 503` 😉

XHawk87 (approved):

lmoureaux:
I added a CMake check to get a nicer error message once 5.6 comes out. I think it's ready to merge now.

PR #2892, opened 2026-05-20
macOS Qt6 linked Dynamically

### LINKED ISSUE #2766: build macOS package with dynamic libraries vs static

jwrober:
With a custom triplet, we can compile Qt in dynamic mode and then link the executable to the proper dylib's with a similar approach that we use on Windows.

Closes https://github.com/longturn/freeciv21/issues/2766


cmake/FreecivDependencies.cmake
```diff
@@ -154,6 +154,10 @@ if (FREECIV_ENABLE_CLIENT
     OR FREECIV_ENABLE_RULEDIT)
   # May want to relax the version later
   find_package(Qt6 6.6 COMPONENTS Widgets REQUIRED)
+  # Link to openSSL for Apple and Win32
+  if(WIN32 OR APPLE)
+    find_package(OpenSSL REQUIRED)
```
lmoureaux: Why? We shouldn't need to link to openssl explicitly, Qt loads it dynamically. Or am I missing something?

jwrober: This fixes a long standing issue where we have to manually copy ssl libs for visual studio. Impacts mac too. Fixing here and will swing back to clean up visual studio after merge.

lmoureaux: Having to copy SSL libs manually is expected behaviour. Qt loads them dynamically at runtime for licensing reasons. This way will not work with linker flags such as `-Wl,--as-needed`.

Maybe we should pass [`--include-plugins`](https://doc.qt.io/qt-6/windows-deployment.html#plugins) or [`--force-openssl`](https://doc.qt.io/qt-6/windows-deployment.html#system-and-runtime-components) to `windeployqt`.

jwrober: Let me look at this

jwrober:
This is in my fork, runs all the tests.
https://github.com/jwrober/freeciv21/actions/runs/27909727725

lmoureaux:
The `tls` plugin is there twice:
image

jwrober:
https://github.com/jwrober/freeciv21/actions/runs/27920617836

lmoureaux (approved):

Issue #2832, opened 2025-09-27
unreachable_protects that only protects unreachable units


lmoureaux:
**Is your feature request related to a problem? Please describe.**
Currently, `unreachable_protects` makes a stack with a mech and a fighter is very strong: the mech defends against air attacks, and land units cannot attack.

**Describe the solution you'd like**
An intermediate value `unreachable_protect=UNREACHABLE` where only unreachable units are protected. That is, a tank could kill the mech, and fighters clear the sky afterwards.

**Describe alternatives you've considered**
Effects, but I don't think they would work in this context. Lua, but too complicated.

**Additional context**
@ec429 would like more control for Aviation.

XHawk87:
How does it work right now with `unreachableprotects=disabled`? Does it mainly affect how `killstack` operates? I.e. If you have a mech and an engineer on a tile with a fighter above, and someone attacks it with a tank and beats the mech, if `killstack` is enabled all three die, and if it's off only the mech dies? If it's just a diplomat and a fighter and a tank attacks, would the fighter defend because it has better defence than a diplomat even though it's supposed to be unreachable?

ec429:
Imho this should be a `killstack` setting `reachable` , rather than an `unreachable_protects` setting.  So that to get the desired behaviour, you set `unreachable_protects=disabled`, `killstack=reachable`; then the tank can attack the stack but if it wins only the mech and eng die, not the fighter.
Bc `unreachable_protects` is about "is attacking this stack allowed", and `killstack` is about "what happens if you win".

> @ec429 would like more control for Aviation.

Well tbqh Avi games normally run with both `unreachable_protects` and `killstack` disabled and I'm happy with that, so this feature isn't a priority for me.

> E.g. Player A could control the ground while Player B could control the air.

That would be fantabulous but does seem like a lot more work and should probably be a separate Issue.

PR #2895, opened 2026-06-02
Unit_Reachable (stable-branch backport)

### LINKED ISSUE #2833: "Reachable" effect
In general, only units that are "reachable" can be attacked (with exceptions controlled by targets=). Which units are reachable is controlled by the `Unreachable` unit class flag and a few hard-coded rules: units in cities and bases are always reachable. This isn't very flexible: for instance, one may want submarines to be unreachable when in deep water.

Add a `Reachable` effect that controls reachability of a unit based on the tile it's on (and maybe the action performed by the opponent). Unhardcode existing rules.

Can't think of any.

**Additional context**
@ec429 would like this for Avi.

ec429:
Implements an effect `Unit_Reachable`, and uses it to unhardcode the existing rule that units in cities and native bases are always reachable regardless of `UCF_UNREACHABLE`.

This PR is against `stable` because my system is not new enough to build (and hence test) `master`.  There are conflicts with `master` because many specenums have moved to `fc_types.h`.  So someone else will have to do a nonzero amount of work to port and re-test on `master`.  (It's your own fault for always needlessly updating your deps to the bleeding edge… 😛️)

### DIFF
--- a/ai/default/aiair.cpp
+++ b/ai/default/aiair.cpp
@@ -124,8 +124,8 @@ static int dai_evaluate_tile_for_air_attack(struct unit *punit,
   int sortie_time;
 #define PROB_MULTIPLIER 100 // should unify with those in combat.c
 
-  if (!can_unit_attack_tile(punit, dst_tile)
-      || !(pdefender = get_defender(punit, dst_tile))) {
+  if (!can_unit_attack_tile(punit, dst_tile, nullptr)
+      || !(pdefender = get_defender(punit, dst_tile, nullptr))) {
     return 0;
   }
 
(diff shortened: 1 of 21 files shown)


XHawk87:
The documentation will also need to be updated to include the new effect. 

FYI, a few of us run into the same issue building the more recent versions, however, if you're on linux, a workaround is to build it with [Nix](https://longturn.readthedocs.io/en/latest/Getting/compile.html#building-with-nix).

XHawk87:
Does this link to an open issue?

lmoureaux:
It implements #2833.
I'm wondering how it interacts with #2832 (but I filed them at the same time so probably not).

ec429:
Documented, given the effect access to "Action" req (mostly; it's janky for the AI), and made the effect more consistent with existing effects by making the unit reachable if the effect value is `> 0`, rather than merely `!= 0`.

lmoureaux (approved):
Looks good, thanks!

clang-format is not very happy, if you have trouble running it I can do

lmoureaux:
Need #2890

jwrober:
I'm not sure we should be pushing this into `stable`. We don't add features there only fix bugs.

lmoureaux:
> I'm not sure we should be pushing this into stable. We don't add features there only fix bugs.

We want to use it for a game, and the server is currently on `stable`. Haven't investigated if we can even build `master`...

jwrober:
I guess this is ready to merge to `stable`. The check failures have other backport commits that will need to be merged at somepoint.

lmoureaux:
We'll keep it master-only and come back to it if needed

PR #2898, opened 2026-06-20
Do not use Qt::Desktop

lmoureaux:
It is a no-op since Qt 6 and is now deprecated.


jwrober:
Needs #2890

lmoureaux:
Rebased

lmoureaux:
(This code gets triggered when changing the tileset. It propagates the tileset change event to all widgets. The city image at the top left corner of the city screen is one example.)

jwrober (approved):

PR #2900, opened 2026-06-20
Do not fail if find_package(Lua) does not define LUA_VERSION_STRING

lmoureaux:
I suspect that vcpkg only defines the standard-named Lua_VERSION.

Should wait for MacOS builds.


jwrober (approved):
Looks good. Let's see what macOS thinks

PR #2913, opened 2026-07-07
More refactoring of packet code

lmoureaux:
The big change in this PR is the way capabilities are handled. Previously, new receive/send functions would be created for every possible combination of capabilities that could affect a packet. I changed this to use `if` statements testing the capabilities at runtime. This simplifies the generation code and reduces the amount of generated code.

A second important change is the switch from stateless function pointers to stateful objects for packet formatting. This helps keep the code more isolated and allowed me to remove `genhash` completely :tada:

An important thing to test with this PR (beyond some play-testing that I already did) is compatibility with older clients/servers (3.0), since compatibility code paths are driven by capabilities. Connecting to a game with a few cities should already test most paths.


lmoureaux:
Interesting include :sweat_smile: 
```
Error: /Users/runner/work/freeciv21/freeciv21/common/networking/dataio_raw.h:14:10: fatal error: 'qt5/QtCore/qbitarray.h' file not found
   14 | #include <qt5/QtCore/qbitarray.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
```

jwrober:
Wow, this is a beast. Great work. I want to give this one a good workout this weekend.

jwrober (approved):
I connected to a server built with this PR with a client from same PR, 3.1.1 and 3.0-patch.5 clients with no issues that I could find.

PR #2918, opened 2026-07-12
Added lua function to make units perform actions

### LINKED ISSUE #2896: Add Unit:perform_action() Lua binding
To implement 'active defence against actions' (à la #2744, which is bogged down in sprawling discussion because the generic case is too flexible to easily express in declarative rulesets).
The specific case I'm interested in is making fighters in Avi scramble against bombers doing things like firebombing (triggering the fighter's Attack action), and also QRA (automatic retaliation) for nuclear MAD (#972).

Lua `perform_action()` method on the `Unit` type.  When called, the specified unit does the identified `struct action`, as though it had been instructed to by the player.  (Unclear whether it should be validated by the actionenabler; if so, the return value should probably indicate whether the action was possible as well as whether it was successful.)

Solve #2744.

**Additional context**
Split out from #972.
Upstream Freeciv 3.1 has `edit.perform_action()` ([commit](https://github.com/freeciv/freeciv/commit/4a6ecbf5d4db2528e5d1212b0a4b79312c966103), [commit](https://github.com/freeciv/freeciv/commit/d143fab5f8ac417035f5791ada234611c7801440)); possibly we should match/copy how they did it.

XHawk87:
Closes #2896

Performing actions can be tested using chat commands:
```
/lua cmd log.debug("%s", find.unit(nil, 101):perform_action(find.action("Found City"), find.unit(nil, 101)))
/lua cmd log.debug("%s", find.unit(nil, 108):perform_action(find.action("Fortify")))
/lua cmd log.debug("%s", find.unit(nil, 101):perform_action(find.action("Build Road"), find.unit(nil, 101), "Road"))
/lua cmd log.debug("%s", find.unit(nil, 130):perform_action(find.action("Attack"), find.unit(nil, 118).tile))
```

### DIFF
--- a/common/scriptcore/stubs/game.lua
+++ b/common/scriptcore/stubs/game.lua
@@ -579,7 +579,7 @@ function Achievement:rule_name() end
 --- @return string name The localised translation of the name.
 function Achievement:name_translation() end
 
---- Represents a type of action that a unit can perform in this ruleset.
+--- Represents a type of action that a unit can perform in this ruleset. See :ref:`modding-ruleset-actions`
 ---
 --- @class Action
 --- @field id int The unique ID of the action type.
--- a/docs/Modding/Rulesets/actions.rst
+++ b/docs/Modding/Rulesets/actions.rst
@@ -7,6 +7,8 @@
 
 .. include:: /global-include.rst
 
+.. _modding-ruleset-actions:
+
 Actions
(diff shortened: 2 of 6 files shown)


jwrober (approved):
Looks good to me. Sample commands worked for me just fine.

PR #2919, opened 2026-07-14
Allow querying and modifying buildings from Lua

### LINKED ISSUE #2640: Lua: Allow adding and removing buildings from cities

Currently, it's impossible to add or remove buildings from cities with Lua. This would be useful for:
- Customising the selling of buildings on negative treasury.
- Customising building destruction on city capture or nuclear strike.
- Customising selling of buildings due to obsolescence.
- Allow bombardment to destroy buildings.
- Custom disasters that destroy buildings.
- Give free buildings due to scripted events, adopting a government, developing a tech, disbanding a unit etc.

Add the following Lua methods:
- `edit.create_building(city, building)` - Add a new building to a city.
- `edit.remove_building(city, building)` - Remove a building from a city.
- `City:create_building(building)` - Alias for `edit.create_building(city, building)`.
- `City:remove_building(building)` - Alias for `edit.remove_building(city, building)`.
- `City:buildings_iterate()` - Iterate over buildings in the city.

**Additional context**
Most of this has been done in freeciv: https://github.com/freeciv/freeciv/commit/153d03100610a6a29e830e2ca25b1bac39fc7814

XHawk87:
Closes #2640

Tested with Lua commands on the console and chat window:
```
/lua cmd for obj in find.government_types_iterate() do log.debug("%s", obj) end
-- Repeated for each iterator in find
/lua cmd for obj in find.city(nil, 192):buildings_iterate() do log.debug("%s", obj) end
/lua cmd log.debug("%s", find.city(nil, 192):building_construct(find.building_type(16)))
/lua cmd log.debug("%s", find.city(nil, 192):building_built_turn_set(find.building_type(16), 1000))
/lua cmd log.debug("%s", find.city(nil, 192):building_destroy(find.building_type(16), "scripted", nil))
/lua cmd log.debug("%s", find.city(nil, 192):building_unmake(find.building_type(16)))
```

### DIFF
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -3281,7 +3281,23 @@ void city_add_improvement(struct city *pcity,
 }
 
 /**
-  Removes an improvement (and its effects) from a city.
+ * Changes the turn on which an improvement has been built in a city. This
+ * can affect its 'Age' in effects. Note: this is a local change only.
+ */
+void city_improvement_built_turn_set(struct city *pcity,
+                                     const struct impr_type *pimprove,
+                                     int turn)
+{
+  log_debug("History rewrite: Improvement %s that was built on turn %d in "
+            "%s was now built on turn %d",
+            improvement_rule_name(pimprove),
+            pcity->built[improvement_index(pimprove)].turn, pcity->name,
+            turn);
+  pcity->built[improvement_index(pimprove)].turn = turn;
+}
+
+/**
+ * Removes an improvement (and its effects) from a city.
  */
 void city_remove_improvement(struct city *pcity,
                              const struct impr_type *pimprove)
@@ -3295,6 +3311,32 @@ void city_remove_improvement(struct city *pcity,
     // Client just read the info from the packets.
     wonder_destroyed(pcity, pimprove);
   }
+}
+
+/**
+ * Removes an improvement (and its effects) from a city, as if it had never
+ * been built.
+ */
+void city_improvement_unmake(struct city *pcity,
+                             const struct impr_type *pimprove)
+{
+  log_debug("History rewrite: Improvement %s was never built in city %s",
+            improvement_rule_name(pimprove), pcity->name);
+  pcity->built[improvement_index(pimprove)].turn = I_NEVER;
+  if (is_server() && is_wonder(pimprove)) {
+    // Client just read the info from the packets.
+    wonder_unmade(pcity, pimprove);
+  }
+}
+
+/**
+ * Gets the turn on which an improvement was built. If it was never built, it
+ * returns I_NEVER. If it was destroyed, it returns I_DESTROYED.
+ */
+int city_improvement_built_turn(struct city *pcity,
+                                const struct impr_type *pimprove)
+{
+  return pcity->built[improvement_index(pimprove)].turn;
 }
 
 /**
--- a/common/city.cpp
+++ b/common/city.cpp
@@ -3284,9 +3284,9 @@ void city_add_improvement(struct city *pcity,
  * Changes the turn on which an improvement has been built in a city. This
  * can affect its 'Age' in effects. Note: this is a local change only.
  */
-void city_improvement_built_turn_set(struct city *pcity,
-                                     const struct impr_type *pimprove,
-                                     int turn)
+void city_improvement_built_turn_override(struct city *pcity,
+                                          const struct impr_type *pimprove,
+                                          int turn)
 {
   log_debug("History rewrite: Improvement %s that was built on turn %d in "
             "%s was now built on turn %d",
@@ -3333,7 +3333,7 @@ void city_improvement_unmake(struct city *pcity,
  * Gets the turn on which an improvement was built. If it was never built, it
  * returns I_NEVER. If it was destroyed, it returns I_DESTROYED.
  */
-int city_improvement_built_turn(struct city *pcity,
+int city_improvement_built_turn(const struct city *pcity,
                                 const struct impr_type *pimprove)
 {
   return pcity->built[improvement_index(pimprove)].turn;
--- a/common/improvement.cpp
+++ b/common/improvement.cpp
@@ -787,8 +787,8 @@ void wonder_built(const struct city *pcity, const struct impr_type *pimprove)
 }
 
 /**
- * Remove a wonder from a city as if it had never been built, allowing it to
- * be constructed again.
+ * Called when a wonder is deleted from a city to ensure it is also deleted
+ * from the player and game info so that it can be constructed again.
  */
 void wonder_unmade(const struct city *pcity,
                    const struct impr_type *pimprove)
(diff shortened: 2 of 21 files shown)


common/scriptcore/stubs/signal_events.lua
```diff
@@ -27,8 +27,8 @@
 ---    signal.connect(signal_name, callback_name)
 ---
 --- Currently, signals are only sent to ruleset scripts running on the server.
---- The set of signals is defined in :file:`server/scripting/script_server.cpp`
---- (search for "``script_server_signals_create``").
+--- The set of signals is defined in
+--- `server/scripting/script_server.cpp <https://github.com/longturn/freeciv21/blob/a6ce76be735fea0a61b39495631f86e12fdafd96/server/scripting/script_server.cpp#L391>`_.
```
jwrober: Is it wise to tag a file based on a commit hash vs just going to the file? In the lines that were removed you state where to go and look. Kind of thinking that is a better approach.

common/scriptcore/tolua_game.pkg
```diff
@@ -730,6 +742,30 @@ do
   function Unit:cargo_iterate()
     return safe_iterate_list(private.Unit.cargo_list_head(self))
   end
+
+  -- Safe iteration over each Building_Type present in the City
+  function City:buildings_iterate()
```
jwrober, line 747: I'm assuming this would also show small wonders since we treat them like city improvements (e.g. buildings)?

jwrober (approved):
This is great. Just a couple questions.

common/scriptcore/tolua_game.pkg
XHawk87: It will show city improvements, small, and great wonders present. You can then check on each one what type it is if desired.

common/scriptcore/stubs/signal_events.lua
XHawk87: I wonder if it would be better still to actually do the search on that file for the right text.

jwrober: That seems like a good idea to me.

XHawk87: Since this is such a small change, I'll wait for @lmoureaux to review the code first and see if anything else needs to be done with it.

common/scriptcore/tolua_game.pkg
```diff
@@ -706,6 +706,18 @@ do
     return value_iterator(objs)
   end
 
+  -- Iterate over the values in the iterator, apply a filter, and create a safe
+  -- copy
+  local function safe_iterate_filtered(iterator, filter)
+    local objs = {}
+    for obj in iterator() do
+      if filter(obj) then
+        objs[#objs + 1] = obj
```
lmoureaux, line 715: Lua objects are really just references to the corresponding C++ objects. What's the behaviour in case an object is deleted mid-iteration?

common/city.cpp
```diff
@@ -3281,7 +3281,23 @@ void city_add_improvement(struct city *pcity,
 }
 
 /**
-  Removes an improvement (and its effects) from a city.
+ * Changes the turn on which an improvement has been built in a city. This
+ * can affect its 'Age' in effects. Note: this is a local change only.
+ */
+void city_improvement_built_turn_set(struct city *pcity,
```
lmoureaux: Use in `city_add_improvement`, `city_remove_improvement`? Or do I fail to understand a semantic difference?

common/city.cpp
```diff
@@ -3297,6 +3313,32 @@ void city_remove_improvement(struct city *pcity,
...
+    // Client just read the info from the packets.
+    wonder_unmade(pcity, pimprove);
+  }
+}
+
+/**
+ * Gets the turn on which an improvement was built. If it was never built, it
+ * returns I_NEVER. If it was destroyed, it returns I_DESTROYED.
+ */
+int city_improvement_built_turn(struct city *pcity,
```
lmoureaux: Use in `packhand.cpp`, maybe also `city_built_iterate`

common/improvement.cpp
```diff
@@ -786,6 +786,25 @@ void wonder_built(const struct city *pcity, const struct impr_type *pimprove)
   }
 }
 
+/**
+ * Remove a wonder from a city as if it had never been built, allowing it to
+ * be constructed again.
+ */
+void wonder_unmade(const struct city *pcity,
```
lmoureaux, line 793:
```suggestion
void wonder_unmake(const struct city *pcity,
```

server/scripting/stubs/server.lua
```diff
@@ -260,6 +260,57 @@ function edit.resize_city(city, size, reason) end
...
+--- Instantly construct a building in the given city. A notification is
+--- generated for the city owner, and data is sent to clients about the change
+--- immediately.
+---
+--- Version added: 3.2
+---
+--- @param city City The city to construct the building in
+--- @param building_type Building_Type The building to construct
+--- @return bool success False, if the building was already present
+function edit.city_building_construct(city, building_type) end
```
lmoureaux, line 276: Worth noting that this allows building a harbour in the middle of the Sahara?

server/scripting/api_server_edit.cpp
```diff
@@ -662,6 +662,102 @@ void api_edit_remove_city(lua_State *L, City *pcity)
...
+  }
+  return true;
+}
+
+/**
+ * Instantly construct a building in the given city as if it had been built
+ * on a given turn. Or modifies the turn if it has already been built. No
+ * notifications are generated, and data about the change is sent to clients
+ * immediately. Returns false if the building was already built on the given
+ * turn.
```
lmoureaux: This doesn't modify culture though, so "as if" is not complete (nor can it be, this is just a remark that could be added to the docs).

lmoureaux (commented):
Looks good overall, some questions about the API

common/scriptcore/tolua_game.pkg
XHawk87: That's where [Nonexistent:exists()](https://longturn.readthedocs.io/en/latest/Modding/Rulesets/script.html#lua-Nonexistent.exists) comes in. I could add a note somewhere about this case.

XHawk87: Although in this specific case, I don't think we ever actually do delete `Building_Type` or `Tech_Type` during a game, right?

common/city.cpp
XHawk87: This was intended for situations where we're editing the history, hence the debug log. It could also be used in the scenario editor or a debug/cheat menu. If we want one central place to modify the turn property for an improvement, we could do that, but it will literally just be a one-liner.

common/scriptcore/tolua_game.pkg
lmoureaux: Indeed (though there are certainly requests to do that)

common/city.cpp
lmoureaux: Maybe `_override` instead of `_set` then?

common/improvement.cpp
XHawk87: In this case, it's the function description that is wrong. It's actually already been unmade/destroyed and now we're updating player and game info records accordingly. I'll update both of those to be more accurate.

XHawk87: There is also no such thing as `great_wonder_transfer` wtf

server/scripting/stubs/server.lua
XHawk87: Correct! In true modding style, you can do the impossible if you wish. I'll add a note to this effect.

PR #2936, opened 2026-07-23
Fix name parameter type in Tile has_extra/has_base/has_road stubs

### LINKED ISSUE #2935: Lua Tile has_extra, has_base, and has_road show incorrect type for name parameter
See [Tile:has_extra()](https://longturn.readthedocs.io/en/latest/Modding/Rulesets/script.html#lua-Tile.has_extra). The `name` parameter is displayed as `boolean` when it's actually a `string`. Same with `has_base` and `has_road`.

**Do you have suggestions?**
Switch them from `boolean` to `string`

LuShadowX:
The Lua API stubs for `Tile:has_extra`, `has_base` and `has_road` annotate the `name` parameter as `boolean`, but it's a string (the ruleset name of the extra/base/road). That's why the generated docs show the wrong type. Changed the three `@param name` annotations from `boolean` to `string`. The `@return boolean` lines are correct and left as-is.

Fixes #2935

### DIFF
--- a/common/scriptcore/stubs/game.lua
+++ b/common/scriptcore/stubs/game.lua
@@ -327,17 +327,17 @@ function Tile:city() end
 function Tile:city_exists_within_max_city_map(center) end
 
 --- 
---- @param name boolean The ruleset name of any type of terrain extra.
+--- @param name string The ruleset name of any type of terrain extra.
 --- @return boolean present True, if the extra is present on this tile.
 function Tile:has_extra(name) end
 
 --- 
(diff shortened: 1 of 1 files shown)


XHawk87 (approved):

lmoureaux:
Thank you for the fix!

LuShadowX:
Happy to help, thanks for the quick review!

LuShadowX:
If you'd ever like the rest of the Lua/Python stub signatures checked against the C API, I'm happy to take that on — just ping me here or on an issue.

PR #2944, opened 2026-07-26
Fix client bugs with the Capture The Flag ruleset

lmoureaux:
This fixes two bugs that lead to a lot of logging output when connecting:
* A bug with `extra_flag_captured`, which has `style=none`. This likely leads to workers building roads shown as building rails, and rails as maglev (haven't checked though).
* A bug that prevented receiving effects with `"UnitClassFlag", "LimitedCR",` correctly. This seems to be used only for `Combat_Rounds`, which isn't displayed directly in the UI but affects battle win chances.

I also slightly improved the logging.

Back port recommended, at least for the two bugfix commits.

### DIFF
--- a/client/tileset/layer_abstract_activities.cpp
+++ b/client/tileset/layer_abstract_activities.cpp
@@ -44,6 +44,10 @@ void layer_abstract_activities::initialize_extra(const extra_type *extra,
                                                  const QString &tag,
                                                  extrastyle_id style)
 {
+  // This function is not called for extras with style=none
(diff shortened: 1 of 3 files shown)


jwrober:
I can confirm that workers do look like they are building rails when they are building road. I think the better logging is worth backport too.

jwrober:
image

jwrober (approved):

XHawk87 (approved):

PR #2945, opened 2026-07-29
IWYU Utility: Remove pragma keeps

jwrober:
Before I get into `server`, `client` and others, I want to ensure that `utility` and `common` are as clean as possible. This draft is to get `utility` done.

### DIFF
--- a/utility/astring.cpp
+++ b/utility/astring.cpp
@@ -8,8 +8,9 @@
 #include "fcintl.h"
 
 // Qt
+#include <QString>
 #include <QStringLiteral>
-#include <QVector>        // IWYU pragma: keep
+#include <QVector>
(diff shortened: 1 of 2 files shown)


jwrober:
To spot I'm stuck on:

fciconv.cpp:58 has `#include <libintl.h> // IWYU pragma: keep`
fcintl.h:20 has the same `#include <libintl.h> // IWYU pragma: keep`

My general thought is we only need one - fcintl.cpp/h since everything in fciconv.cpp/h is deprecated. I can't seem to figure out how to remove the pragma keep here.

lmoureaux (approved):

lmoureaux:
`libintl.h` may export different things depending on the configuration (it's different on Windows for instance, IIRC). What's unused on one system may not be on another.

PR #2952, opened 2026-08-04
IWYU: server a to c

jwrober:
IWYU Header update for `server` a* to c*.


server/aiiface.cpp
```diff
@@ -1,36 +1,33 @@
...
 
-#include "aiiface.h"
+// Qt
+#include <QtLogging> // qDebug, qWarning, qCritical
+
+// std
+#include <cstdlib> // EXIT_FAILURE
+
+#ifdef AI_MODULES
+#include <ltdl.h>
```
lmoureaux, line 29: Side note: we should review whether we want to support dynamic AI modules.

lmoureaux (approved):

PR #2953, opened 2026-08-04
Removed rogue name property from E module stubs

### LINKED ISSUE #2942: Lua docs 'name' property incorrectly placed in E module

This doesn't belong in the E module, it's not a notification event type.

Image

**Do you have suggestions?**

Remove it, and find out if it belongs somewhere else.

XHawk87:
Closes #2942


jwrober (approved):
LGTM

PR #2954, opened 2026-08-04
Removed find.cities_iterate and find.units_iterate

### LINKED ISSUE #2929: Lua: Iterators for destructible types may be incomplete
**Describe the bug**
Currently, the iterators in the `find` module work by checking each index number in turn and stopping as soon as a nil value is encountered. This works for fixed lists like `Unit_Type` and `Tech_Type` but doesn't work correctly for objects that can be deleted, such as `Unit` and `City`. 

**To Reproduce**
1. Start a new game, any ruleset.
2. Use `/lua cmd for obj in find.units_iterate() do log.debug("%s", obj) end` to see initial unit list.
3. Type `/lua cmd find.unit(nil, 4):kill("deleted", nil)` to delete one of the units.
4. Use command from Step 1 to check again and see the rest of the list is missing.

**Expected behavior**
All currently existing objects of that type should be iterated over, skipping deleted objects.

**Screenshots**

 - OS: Any. Found using Linux.
 - Freeciv21 version: 3.2-dev.3 (https://github.com/longturn/freeciv21/commit/0682e3cd7306ebc378424310ef2e2693428a774e)
 - Ruleset/Longturn game (if applicable): Any. Found using LT-CTF.

**Additional context**

XHawk87:
Closes #2929

Script authors have an easy workaround:
```
for player in find.players_iterate() do
  for unit in player:units_iterate() do
    log.debug("%s", unit)
  end
end
```

And it would be a lot of work just to shorten it to:
```
for player in find.units_iterate() do
  log.debug("%s", unit)
end
```
With less control over how the units are ordered. 

So I just removed them. It's not like they ever worked in the first place to cause a problem in existing scripts.

### DIFF
--- a/common/scriptcore/stubs/game.lua
+++ b/common/scriptcore/stubs/game.lua
@@ -676,35 +676,11 @@ function find.teams_iterate() end
 --- @return City city The city if it exists and belongs to player, or nil
 function find.city(player, city_id) end
 
---- Safe iteration over each :lua:obj:`City` in the game. See 
---- :lua:obj:`Player.cities_iterate` for iterating over a specific player's
---- cities.
----
---- .. note::
----    If there is a chance that a :lua:obj:`City` is destroyed during
----    iteration, it is a good idea to use :lua:obj:`~Nonexistent.exists` to
----    check it is still present before attempting to use it.
----
---- Version added: 3.2
-function find.cities_iterate() end
-
 --- 
 --- @param player Player The player who owns the unit, or nil for any player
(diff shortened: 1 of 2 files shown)


lmoureaux:
When had these functions been added? The docs say since 3.2, so this should be harmless in any case?

XHawk87:
> When had these functions been added? The docs say since 3.2, so this should be harmless in any case?

Yeh, they'd literally just been added and didn't work.

lmoureaux (approved):

PR #2956, opened 2026-08-08
Fix upkeep unhomed units and server setting

### LINKED ISSUE #2948: Homeless units are appearing in costs on the client despite homeless_gold_upkeep being disabled
**Describe the bug**

Homeless units are showing a gold upkeep cost in the client despite the `homeless_gold_upkeep` option being disabled. This has no effect on actual gold income.

**To Reproduce**
1. Start any game with 'homeless_gold_upkeep' disabled.
2. Check income in client
3. See negative income despite all initial units being unhomed.

**Expected behavior**

Income should be zero, and unhomed units should cost 0 upkeep.

**Screenshots**

Image

 - OS: Linux
 - Freeciv21 version: 3.2.3-dev (https://github.com/longturn/freeciv21/commit/50c1c0508dd8055861ee0175d8a66603b8c345ef)
 - Ruleset/Longturn game (if applicable): Any

**Additional context**
See [discord discussion](https://discord.com/channels/378908274113904641/411517969059938304/1528103665745788978)

jwrober:
Closes #2948 

This might be a back port candidate

### DIFF
--- a/client/repodlgs_common.cpp
+++ b/client/repodlgs_common.cpp
@@ -120,8 +120,9 @@ void get_economy_report_units_data(struct unit_entry *entries,
   {
     cost = utype_upkeep_cost(unittype, client.conn.playing, O_GOLD);
 
+    // Econ view only shows units that have a cost. If the base upkeep is
+    // zero we skip the iterator completely.
     if (cost == 0) {
-      // Short-circuit all of the following checks.
       continue;
     }
 
@@ -130,13 +131,15 @@ void get_economy_report_units_data(struct unit_entry *entries,
 
     unit_list_iterate(client.conn.playing->units, punit)
     {
-      if (unit_type_get(punit) == unittype && punit->upkeep[O_GOLD] > 0) {
+      // As we move into the iterator, we need to ensure the unit matches the
+      // unittype for accurate counts and that the unit has some gold upkeep.
+      // We also need to check if the unit is not homeless or the game
+      // setting is set for homeless gold upkeep.
+      if ((!unit_is_homeless(punit) || game.server.homeless_gold_upkeep)
+          && (unit_type_get(punit) == unittype
+              && punit->upkeep[O_GOLD] > 0)) {
         count++;
-        // get upkeep if the unit isn't homeless or we have the server
-        // setting set
-        if (!unit_is_homeless(punit) || game.server.homeless_gold_upkeep) {
-          partial_cost += punit->upkeep[O_GOLD];
-        }
+        partial_cost += punit->upkeep[O_GOLD];
       }
     }
     unit_list_iterate_end;
--- a/client/repodlgs_common.cpp
+++ b/client/repodlgs_common.cpp
@@ -130,9 +130,13 @@ void get_economy_report_units_data(struct unit_entry *entries,
 
     unit_list_iterate(client.conn.playing->units, punit)
     {
-      if (unit_type_get(punit) == unittype && punit->upkeep[O_GOLD] > 0) {
+      if (unit_type_get(punit) == unittype) {
         count++;
-        partial_cost += punit->upkeep[O_GOLD];
+        // get upkeep if the unit isn't homeless or we have the server
+        // setting set
+        if (!unit_is_homeless(punit) || game.server.homeless_gold_upkeep) {
+          partial_cost += punit->upkeep[O_GOLD];
+        }
       }
     }
     unit_list_iterate_end;
--- a/client/repodlgs_common.cpp
+++ b/client/repodlgs_common.cpp
@@ -130,7 +130,7 @@ void get_economy_report_units_data(struct unit_entry *entries,
 
     unit_list_iterate(client.conn.playing->units, punit)
     {
-      if (unit_type_get(punit) == unittype) {
+      if (unit_type_get(punit) == unittype && punit->upkeep[O_GOLD] > 0) {
         count++;
         // get upkeep if the unit isn't homeless or we have the server
         // setting set
--- a/client/views/view_units.cpp
+++ b/client/views/view_units.cpp
@@ -663,8 +663,11 @@ std::vector<unit_view_entry> get_units_view_data()
           upgradable =
               client_has_player()
               && nullptr != can_upgrade_unittype(client_player(), unittype);
-
-          gold_cost += punit->upkeep[O_GOLD];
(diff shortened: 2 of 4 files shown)


client/views/view_units.cpp
```diff
@@ -664,7 +664,11 @@ std::vector<unit_view_entry> get_units_view_data()
               client_has_player()
               && nullptr != can_upgrade_unittype(client_player(), unittype);
 
-          gold_cost += punit->upkeep[O_GOLD];
+          if (unit_is_homeless(punit) && !game.server.homeless_gold_upkeep) {
+            gold_cost = 0;
```
XHawk87: If any unit in the list is homeless and homeless gold upkeep is off, it resets the accumulated gold count to zero. That doesn't look right.

common/unittype.cpp
```diff
@@ -142,6 +142,11 @@ int utype_upkeep_cost(const struct unit_type *ut, struct player *pplayer,
     return 0;
   }
 
+  if (otype == O_GOLD && !game.server.homeless_gold_upkeep) {
+    // Special case: homeless unit gold upkeep
+    return 0;
```
XHawk87: All units gold upkeep is zero if homeless gold upkeep is off regardless of if the unit is homeless or not? That doesn't look right.

server/plrhand.cpp
```diff
@@ -2004,7 +2004,7 @@ void player_update_homeless_unit_gold_upkeep(struct player *pplayer)
   // punit struct
   unit_list_iterate(pplayer->units, punit)
   {
-    if (unit_is_homeless(punit)) {
+    if (unit_is_homeless(punit) && game.server.homeless_gold_upkeep) {
```
XHawk87: Why even run this function at all if homeless gold upkeep is off? It's not going to change during iteration.

common/player.cpp
```diff
@@ -1263,7 +1263,7 @@ int player_total_homeless_unit_gold_upkeep(const struct player *pplayer)
             player_name(pplayer));
   unit_list_iterate(pplayer->units, punit)
   {
-    if (unit_is_homeless(punit)) {
+    if (unit_is_homeless(punit) && game.server.homeless_gold_upkeep) {
```
XHawk87: Same here. No point iterating over the whole unit list if none of them are going to add to the gold total. Just return 0 at the start.

XHawk87:
I am not sure back-porting is necessary. The last time I checked, this bug isn't present on stable.

XHawk87 (changes requested):

jwrober:
> I am not sure back-porting is necessary. The last time I checked, this bug isn't present on stable.

Got to thinking about this. If its not present in `stable` that means we created a regression in `master`.

XHawk87:
> > I am not sure back-porting is necessary. The last time I checked, this bug isn't present on stable.
> 
> Got to thinking about this. If its not present in `stable` that means we created a regression in `master`.

Yeh, I am still wondering about what happened. Though I do think that coding defensively and having extra checks like these is a good thing.

client/views/view_units.cpp
jwrober: I think I need to figure out where this is calculated: `punit->upkeep[O_GOLD];`

client/repodlgs_common.cpp
```diff
@@ -130,9 +130,13 @@ void get_economy_report_units_data(struct unit_entry *entries,
 
     unit_list_iterate(client.conn.playing->units, punit)
     {
-      if (unit_type_get(punit) == unittype && punit->upkeep[O_GOLD] > 0) {
```
XHawk87, line 133: This changes the logic slightly. Before units with no upkeep were not counted, and now they are counted but just don't add to the total cost. Is this an intended change? Should this report display only units with upkeep, or all units even if their upkeep is zero? If it's the latter, the short-circuit above should probably go.

common/player.cpp
```diff
@@ -1263,7 +1263,9 @@ int player_total_homeless_unit_gold_upkeep(const struct player *pplayer)
             player_name(pplayer));
   unit_list_iterate(pplayer->units, punit)
   {
-    if (unit_is_homeless(punit)) {
+    // get upkeep if the unit isn't homeless or we have the server setting
+    // set
+    if (!unit_is_homeless(punit) || game.server.homeless_gold_upkeep) {
```
XHawk87, line 1268: This now includes units that aren't homeless in the homeless unit gold upkeep, or all units if homeless gold upkeep is off. 

I would leave it as it was, but add a `if (!game.server.homeless_gold_upkeep) { return 0; }` further up instead.

server/plrhand.cpp
```diff
@@ -2004,7 +2004,9 @@ void player_update_homeless_unit_gold_upkeep(struct player *pplayer)
   // punit struct
   unit_list_iterate(pplayer->units, punit)
   {
-    if (unit_is_homeless(punit)) {
+    // get upkeep if the unit isn't homeless or we have the server setting
+    // set
+    if (!unit_is_homeless(punit) || game.server.homeless_gold_upkeep) {
```
XHawk87, line 2009: Same here. Just add a `if (!game.server.homeless_gold_upkeep) { return; }` further up instead and set this back as it was.

XHawk87 (changes requested):

client/repodlgs_common.cpp
jwrober: The econ view is only supposed to show you stuff that costs gold. Good catch on the logic change. Updated in latest commit.

client/repodlgs_common.cpp
```diff
@@ -132,7 +132,11 @@ void get_economy_report_units_data(struct unit_entry *entries,
     {
       if (unit_type_get(punit) == unittype && punit->upkeep[O_GOLD] > 0) {
         count++;
-        partial_cost += punit->upkeep[O_GOLD];
+        // get upkeep if the unit isn't homeless or we have the server
+        // setting set
+        if (!unit_is_homeless(punit) || game.server.homeless_gold_upkeep) {
```
XHawk87: Since this report only shows units that cost gold, shouldn't this check actually be on line 133 above to exclude this unit from the `count` as well as the `partial_cost`?

XHawk87 (changes requested):

XHawk87 (approved):

PR #2961, opened 2026-08-12
Make Unit_Class Lua methods accept a const punit_class

ec429:
They're all read-only, and the most likely way for a Lua script to get a
 uclass (utype.unit_class) gives a const, so this change is needed to be
 able to actually look up unit class flags.

### DIFF
--- a/common/scriptcore/api_game_methods.cpp
+++ b/common/scriptcore/api_game_methods.cpp
@@ -1324,7 +1324,8 @@ int api_methods_unit_type_veteran_level_max_get(lua_State *L,
 /**
  * Return true if the unit class has the given unit class flag.
  */
-bool api_methods_unit_class_has_flag(lua_State *L, Unit_Class *punit_class,
+bool api_methods_unit_class_has_flag(lua_State *L,
+                                     const Unit_Class *punit_class,
                                      const char *flag)
 {
   LUASCRIPT_CHECK_STATE(L, false);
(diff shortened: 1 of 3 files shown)


lmoureaux:
Can't push to your branch, so you'll have to implement it yourself :grin: 
```diff
diff --git a/common/scriptcore/api_game_methods.cpp b/common/scriptcore/api_game_methods.cpp
index 67778131d0..91fc8e41a2 100644
--- a/common/scriptcore/api_game_methods.cpp
+++ b/common/scriptcore/api_game_methods.cpp
@@ -1355,8 +1355,9 @@ const char *api_methods_unit_class_rule_name(lua_State *L,
 /**
  * Return the translated name of the unit class.
  */
-const char *api_methods_unit_class_name_translation(lua_State *L,
-                                                    const Unit_Class *punit_class)
+const char *
+api_methods_unit_class_name_translation(lua_State *L,
+                                        const Unit_Class *punit_class)
 {
   LUASCRIPT_CHECK_STATE(L, nullptr);
   LUASCRIPT_CHECK_SELF(L, punit_class, nullptr);
diff --git a/common/scriptcore/api_game_methods.h b/common/scriptcore/api_game_methods.h
index c42d765eba..058abafc5e 100644
--- a/common/scriptcore/api_game_methods.h
+++ b/common/scriptcore/api_game_methods.h
@@ -199,8 +199,9 @@ bool api_methods_unit_class_has_flag(lua_State *L,
                                      const char *flag);
 const char *api_methods_unit_class_rule_name(lua_State *L,
                                              const Unit_Class *punit_class);
-const char *api_methods_unit_class_name_translation(lua_State *L,
-                                                    const Unit_Class *punit_class);
+const char *
+api_methods_unit_class_name_translation(lua_State *L,
+                                        const Unit_Class *punit_class);
 
 // Unit_List_Link Type
 Unit *api_methods_unit_list_link_data(lua_State *L, Unit_List_Link *link);
```

XHawk87:
Does this work with the `find.unit_class` function still? Should that return a `const` too?

ec429:
It should work — I haven't tested it, but passing a `T *` to a function that takes a `const T *` is generally allowed (as long as `T` is not itself a pointer type), unlike the reverse.

lmoureaux (approved):

Issue #2778, opened 2025-08-20
Review Markdown formatting in help


lmoureaux:
Formatting of the help text needs to be reviewed following the transition to Markdown (#2693). See PR for comments.

lmoureaux:
Yes this could be useful. It also matters for ruleset `README`s so it should go in the Modding section.

jwrober:
@blabber I'm willing to help here, but could use some guidance on what you are planning/thinking.

jwrober:
> My only plan is to go through `data/helpdata.txt` and make sure it renders in a sensible way.
> 
> For some sections, like "Controls" and "Chatline", this will be more work than for most others.

I think that is fine. My ask is for you to add some notes/comments to `style-guide.rst` to explain what you've done to make the primary static help look better. Than can be used to help with updating the rulesets, which will take awhile.

PR #2984, opened 2026-09-03
Implement links to other help topics from texts in the help browser

blabber:
This PR disables the internal link handling of the `QTextBrowser` in the help browser and replaces it with a manual handler, that supports a new URL scheme to link other help topics. This scheme is documented in the markdown style guide, corresponding updates are part of this PR.

As an example and proof of concept I added a plethora of links in the "Strategy and Tactics" help.

The downside of this is, that these links will be part of the HTML generated by `freeciv21-manual` and no browser will be able to interpret these links.

Part of #2778


client/helpdlg.cpp
```diff
@@ -797,6 +816,19 @@ void help_widget::add_info_separator()
  */
 void help_widget::info_panel_done() { info_layout->addStretch(); }
 
+/**
+ Handles clicked urls.
+*/
+void help_widget::anchor_clicked(const QUrl &url)
+{
+  if (url.scheme() == QString("fch")) {
+    follow_help_link(url.path());
```
lmoureaux: `follow_help_link` needs a localized name. This is problematic in case a unit name has been translated, but the help text referring to it hasn't.

It may be something to fix in `follow_help_link`.

client/fc_client.cpp
```diff
@@ -295,6 +301,14 @@ void fc_client::closeEvent(QCloseEvent *event)
   }
 }
 
+/**
+ * Opens a help link (fch://).
+ */
+void fc_client::handle(const QUrl &url) const
+{
+  follow_help_link(url.path());
```
lmoureaux, line 309: follow_help_link takes a localized name. This is problematic in case a unit name has been translated, but the help text referring to it hasn't. I think URLs should be English-only.

lmoureaux (changes requested):
With the links becoming visible externally, I would prefer if they used names instead of numeric identifiers for categories. Enum `help_page_type` will change, e.g., if we add documentation about actions. I think it should be turned into a specenum and the names used in URLs of the form `fch:Unit/Settlers` (`/` seems more standard than `,`).

blabber:
> With the links becoming visible externally, I would prefer if they used names instead of numeric identifiers for categories. Enum `help_page_type` will change, e.g., if we add documentation about actions. I think it should be turned into a specenum and the names used in URLs of the form `fch:Unit/Settlers` (`/` seems more standard than `,`).

Done. Updated help texts and style guide accordingly.

client/fc_client.cpp
blabber: Can I convince you to take over this topic? I'm not really up to speed on localization...

lmoureaux: Sure. Let's turn it into an issue and add it to the 3.2 blockers.

lmoureaux (approved):

PR #2990, opened 2026-09-06
Remove explicit link_activated handlers

blabber:
Remove all instances of explicit `link_activated` handlers. This fixes the generated links in the help browser and the links in the editor.

Part of #2778


jwrober:
Do the links in the map editor still work or do they get disabled with this change? How about links on the map that come up with middle-click on a tile?

blabber:
All help links I was able to find in theUI work as expected. Including the links in the editor and the middle-click popup.

jwrober (approved):
This seems to work fine for me, so going to approve.

Issue #3000, opened 2026-09-12
Add a clang-tidy CI check


jwrober:
**Is your feature request related to a problem? Please describe.**
As we move further into the IWYU project (https://github.com/longturn/freeciv21/issues/2464), but not finished yet, we need a check against known cleaned directories to ensure that code updates to those directories stay clean

**Describe the solution you'd like**
Add a CI lint check at the same time as we run clang-format.  I see this as happening in two steps:

1. Run a full build and then lint the known good/clean directories. Fail on warning.
2. Once the effort is complete, change to use a GH action to just check the changed files

```
# Get only changed .cpp and .h files
- name: Get changed C++ files
  id: changed-cpp-files
  uses: tj-actions/changed-files@v45
  with:
    files: |
      **/*.cpp
      **/*.h

# Run clang-tidy on the changed files (if any exist)
- name: Run Clang-Tidy Include Cleaner
  if: steps.changed-cpp-files.outputs.any_changed == 'true'
  run: |
    clang-tidy \
      --checks="-*,misc-include-cleaner" \
      --warnings-as-errors="*" \
      ${{ steps.changed-cpp-files.outputs.all_changed_files }}
```

**Describe alternatives you've considered**
Added a build that a PR approver can review https://github.com/longturn/freeciv21/pull/2992, however would prefer something automatic.

**Additional context**

PR #3003, opened 2026-09-13
Add clang-tidy details to current check

jwrober:
In order for us to have a `clang-tidy` pre-check like we do for `clang-format` we need to fully know what the CI is going to see. Adding code to the existing CI to run against the whole tree. It will always fail, but for me that is ok. We use it for diagnostics.

part of #3000


.github/workflows/build.yaml
```diff
@@ -174,6 +174,25 @@ jobs:
...
+          
+          # Utility - because it's special
+          clang-tidy \
+            $(find utility -type d -name tests -prune -o -type f \( -name '*.cpp' -o -name '*.h' \) ! -name 'spec*' -print) \
+            -p build-clang \
+            --checks="-*,misc-include-cleaner"
+          
+          # The Rest
+          clang-tidy \
+            $(find ai client common data server tools -type d -name tests -prune -o -type f \( -name '*.cpp' -o -name '*.h' \) -print) \
```
lmoureaux: I don't think we have code in `data`
```suggestion
            $(find ai client common server tools -type d -name tests -prune -o -type f \( -name '*.cpp' -o -name '*.h' \) -print) \
```

lmoureaux (commented):
A failure in `utility` will prevent the workflow from continuing. Since we have [`continue-on-error`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?versionId=free-pro-team%40latest&productId=actions&restPage=reference%2Cworkflows-and-actions%2Cexpressions#jobsjob_idstepscontinue-on-error), what about a structure like this:
1. **utility** - if this fails, the job fails
2. **common** - if this fails, the job fails
3. **ai client server tools** - still allowed to fail for now

Then we move folders from 3 to 1 over time.

.github/workflows/build.yaml
jwrober: We have tests in data, but since QTest isn't fully IWYU compliant pulling this out makes sense.

jwrober:
> A failure in `utility` will prevent the workflow from continuing. Since we have [`continue-on-error`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?versionId=free-pro-team%40latest&productId=actions&restPage=reference%2Cworkflows-and-actions%2Cexpressions#jobsjob_idstepscontinue-on-error), what about a structure like this:
> 
> 1. **utility** - if this fails, the job fails
> 2. **common** - if this fails, the job fails
> 3. **ai client server tools** - still allowed to fail for now
> 
> Then we move folders from 3 to 1 over time.

This is effectively what I was thinking too. I have a branch in my local I started testing with yesterday when I realized the repo isn't quite ready. The test for `utility` and `common` I think can be combined. I kept `utility` separate because of the `spec*.h` files that don't pass any kind of IWYU check due to their nature.

image

jwrober:
I'll need to bring in #3001 and #3002 in here in a rebase for this to work as expected.

lmoureaux:
Both merged

